PR #7, opened 2019-02-21
API Doc for RadioButtons

kikisaints:
Request to review and merge RadioButtons API doc.

### DIFF
--- a/active/RadioButtonsAPI.md
+++ b/active/RadioButtonsAPI.md
@@ -1,0 +1,81 @@
+<!-- The purpose of this spec is to describe a new feature and
+its APIs that make up a new feature in WinUI. -->
+
+<!-- There are two audiences for the spec. The first are people
+that want to evaluate and give feedback on the API, as part of
+the submission process.  When it's complete
+it will be incorporated into the public documentation at
+docs.microsoft.com (http://docs.microsoft.com/uwp/toolkits/winui/).
+Hopefully we'll be able to copy it mostly verbatim.
+So the second audience is everyone that reads there to learn how
+and why to use this API. -->
+
+# Background
+<!-- Use this section to provide background context for the new API(s)
+in this spec. -->
+
+<!-- This section and the appendix are the only sections that likely
+do not get copied to docs.microsoft.com; they're just an aid to reading this spec. -->
+
+<!-- If you're modifying an existing API, included a link here to the
+existing page(s) -->
+
+<!-- For example, this section is a place to explain why you're adding this API rather than
+modifying an existing API. -->
+
+<!-- For example, this is a place to provide a brief explanation of some dependent
+area, just explanation enough to understand this new API, rather than telling
+the reader "go read 100 pages of background information posted at ...". -->
+
+RadioButtons is our answer to enabling better keyboarding and narrator support for RadioButtons that need to be presented in a related list.
+
+Currently, the only way to group RadioButtons is through using a StackPanel and individually labeling each RadioButton through its GroupName property in order for narrator to read out the options somewhat understandably.
+
+However, using that method still left gaps in the narrator experience and StackPanel does not behave the way a user would expect when navigating the list via keyboard - RadioButtons was introduced to help solve those issues.
+
+# Description
+<!-- Use this section to provide a brief description of the feature.
+For an example, see the introduction to the PasswordBox control
+(http://docs.microsoft.com/windows/uwp/design/controls-and-patterns/password-box). -->
+
+RadioButtons is a class extension of ListView that enables you to create related lists of RadioButtons easily, while also fully supporting keyboarding and narrator functionality.
+
+# Examples
+<!-- Use this section to explain the features of the API, showing
+example code with each description. The general format is:
+  feature explanation,
+  example code
+  feature explanation,
+  example code
+  etc.-->
+
+<!-- As an example of this section, see the Examples section for the PasswordBox control
+(https://docs.microsoft.com/windows/uwp/design/controls-and-patterns/password-box#examples). -->
+
+The following example demonstrates how to create a small collective list of RadioButtons using RadioButtons.
+
+```xaml
+<RadioButtons Header="App Mode" SelectedIndex="2">
+    <x:String>Dark</x:String>
+    <x:String>Light</x:String>
+    <x:String>Windows Default</x:String>         
+</RadioButtons>
+```
+
+# API Details
+<!-- The exact API, in MIDL3 format (https://docs.microsoft.com/en-us/uwp/midl-3/) -->
+
+# API Notes
+<!-- Give a one or two line description of each API (type
+and member), or at least the ones that aren't obvious
+from their name.  These descriptions are what show up
+in IntelliSense. -->
+
+### Notable Properties
+The following properties similarly to how items do in ListView, however there are some behavioral differences due to the fact that these items function as grouped RadioButtons.
+
+| Name | Description |
+|:-:|:--|
+| Header | Places a text label above the list of RadioButtons and is read out by the narrator on focus. |
+| SelectedIndex | Gets or sets the index of the selected item. Selection is denoted by that item's RadioButton icon being checked. |
+| Items | Gets or sets an object source used to generate the content of the ItemsControl. All items placed within the RadioButtons will get a RadioButton icon inline with the item. |


kikisaints:
@MikeHillberg or @jevansaks Not sure what Tammy's handle is on here, but I need dev assistance with generating the IDL for the **API Details** section.

jevansaks:
> @MikeHillberg Mike Hillberg FTE or @jevansaks Jevan Saks FTE Not sure what Tammy's handle is on here, but I need dev assistance with generating the IDL for the **API Details** section.

@teaP -- you can [look up individuals here](https://repos.opensource.microsoft.com/people)

But actually IDL is already available:
* https://github.com/Microsoft/microsoft-ui-xaml/blob/master/dev/RadioButtons/RadioButtons.idl
* https://github.com/Microsoft/microsoft-ui-xaml/blob/master/dev/RadioButtons/RadioButtonsAutomationPeer.idl
* https://github.com/Microsoft/microsoft-ui-xaml/blob/master/dev/RadioButtons/RadioButtonsPrimitives.idl

kikisaints:
Thanks @jevansaks! I added Tammy to this PR and the IDLs to the doc.

active/RadioButtonsAPI.md
```diff
@@ -0,0 +1,105 @@
...
+}
+```
+
+### RadioButtons Primitives
+
+```
+namespace Microsoft.UI.Xaml.Controls.Primitives
+{
+[webhosthidden]
+unsealed runtimeclass RadioButtonsListView : Windows.UI.Xaml.Controls.ListView
```
stmoy: Is *.Primitives the right place for this API? It contains things like ButtonBase and Popup.

jevansaks: It's an implementation detail of the control itself but needs to be public in case you retemplate. Primitives is where we've put things like this in the past but I'm fine if you think another namespace is appropriate?

active/RadioButtonsAPI.md
```diff
@@ -0,0 +1,105 @@
...
+
+RadioButtons is our answer to enabling better keyboarding and narrator support for RadioButtons that need to be presented in a related list.
+
+Currently, the only way to group RadioButtons is through using a StackPanel and individually labeling each RadioButton through its GroupName property in order for narrator to read out the options somewhat understandably.
+
+However, using that method still left gaps in the narrator experience and StackPanel does not behave the way a user would expect when navigating the list via keyboard - RadioButtons was introduced to help solve those issues.
+
+# Description
+
+RadioButtons is a class extension of ListView that enables you to create related lists of RadioButtons easily, while also fully supporting keyboarding and narrator functionality.
```
jevansaks: It contains a ListView but it intentionally does not derive from ListView to avoid confusing the API with all the ListView API. This may or may not be the right choice but we shouldn't say it's a "class extension".

active/RadioButtonsAPI.md
```diff
@@ -0,0 +1,105 @@
...
+
+# Examples
```
jevansaks: A big enhancement is this control's ability to work in an MVVM setup, could you show an example of databinding with this?

active/RadioButtonsAPI.md
```diff
@@ -0,0 +1,105 @@
...
+    Windows.UI.Xaml.DataTemplate ItemTemplate;
+
+    Windows.UI.Xaml.DependencyObject ContainerFromItem(Object item);
+    Windows.UI.Xaml.DependencyObject ContainerFromIndex(Int32 index);
+
+    Int32 SelectedIndex;
+    Object SelectedItem;
+    event Windows.UI.Xaml.Controls.SelectionChangedEventHandler SelectionChanged;
+
+    Int32 MaximumColumns;
```
jevansaks: Need samples for some of these other properties.

active/RadioButtonsAPI.md
```diff
@@ -0,0 +1,105 @@
...
+```
+
+# API Notes
+
+### Notable Properties
+The following properties similarly to how items do in ListView, however there are some behavioral differences due to the fact that these items function as grouped RadioButtons.
+
+| Name | Description |
+|:-:|:--|
+| Header | Places a text label above the list of RadioButtons and is read out by the narrator on focus. |
```
jevansaks: Other controls with "Header" also have a HeaderTemplate property, I think we might be missing that?

active/RadioButtonsAPI.md
```diff
@@ -0,0 +1,105 @@
...
+{
+[webhosthidden]
+[contentproperty("Items")]
+unsealed runtimeclass RadioButtons : Windows.UI.Xaml.Controls.Control
+{
+    RadioButtons();
+
+    Object ItemsSource;
+    Windows.Foundation.Collections.IVector<Object> Items{ get; };
+    Windows.UI.Xaml.DataTemplate ItemTemplate;
```
jevansaks: Needs ItemTemplateSelector to complete the trinity?

active/RadioButtonsAPI.md
kikisaints: Okay, I'll just say "control" then.

active/RadioButtonsAPI.md
kikisaints: You're referring to this ``` ItemsSource="{x:Bind radioButtonItems}" ``` kind of binding, correct?

active/RadioButtonsAPI.md
kikisaints: I'm not sure what you mean... ItemTemplateSelector should be in the IDL or in the API Notes?

active/RadioButtonsAPI.md
jevansaks: Right, and binding of the SelectedItem as well.

active/RadioButtonsAPI.md
```diff
@@ -0,0 +1,153 @@
...
+        radioButtonItems.Add(new RadioButtonItems() { label = "Item 1" });
+        radioButtonItems.Add(new RadioButtonItems() { label = "Item 2" });
+        radioButtonItems.Add(new RadioButtonItems() { label = "Item 3" });
+
+    }
+}
+```
+
+### Multiple Columns
+Some groups of RadioButtons may want a multi-column layout. This is an example on how to set that up.
```
jevansaks: Maybe a picture here?

active/RadioButtonsAPI.md
jevansaks: Right ... it's missing. It can wait until when @teaP is back but there's a couple APIs I'd expect to see in this control that I don't.

active/RadioButtonsAPI.md
```diff
@@ -0,0 +1,153 @@
...
+```xaml
+<RadioButtons Header="App Mode" SelectedIndex="2">
+    <x:String>Dark</x:String>
+    <x:String>Light</x:String>
+    <x:String>Windows Default</x:String>         
+</RadioButtons>
+```
+
+### DataBinding with RadioButtons
+An example of simple data binding of the ItemsSource with RadioButtons.
```
stmoy: The example below doesn't show data binding with _RadioButtons_ - it shows data binding to _RadioButtonItems_. Is this the expected usage? Or should it look like: 

```
        radioButtonItems = new List<RadioButton>();
        radioButtonItems.Add(new RadioButton() { Content = "Item 1" });
```

?

jevansaks: I was confused too, RadioButtonItem isn't a UI type it's a custom data type defined above in the sample. But I think the sample type should be changed to like "MyOption" or something like that to reduce confusion.

active/RadioButtonsAPI.md
stmoy: I don't have strong pushback (or a better idea 😅) it just seemed like a more complex control than some of the other types found in that namespace.

active/RadioButtonsAPI.md
```diff
@@ -0,0 +1,153 @@
...
+
+```xaml
+<preview:RadioButtons x:Name="RadioButtonGroup" Header="App Mode" ItemsSource="{x:Bind radioButtonItems}" />
+```
+
+```C#
+public sealed partial class MainPage : Page
+{
+    public class RadioButtonItems
```
jevansaks:
```suggestion
    public class OptionDataModel
```

jevansaks: Just a suggestion to reduce confusion (and if you accept then update the names below. At the very least this type shouldn't be plural. :)

active/RadioButtonsAPI.md
kikisaints: You can't put RadioButtons into the RadioButton**s** group because you'll get this:

image

kikisaints: I think I fixed this by taking Jevan's suggestion. Take a look!

active/RadioButtonsAPI.md
teaP: I was trying to keep it light and simple. Why would this control ever need multiple ItemTemplates?

kikisaints:
@MikeHillberg does this look ready for an API review to you?

active/radiobuttons/RadioButtonsAPI.md
```diff
@@ -0,0 +1,155 @@
...
+
+RadioButtons is our answer to enabling better keyboarding and narrator support for RadioButtons that need to be presented in a related list.
+
+Currently, the only way to group RadioButtons is through using a StackPanel and individually labeling each RadioButton through its GroupName property in order for narrator to read out the options somewhat understandably.
+
+However, using that method still left gaps in the narrator experience and StackPanel does not behave the way a user would expect when navigating the list via keyboard - RadioButtons was introduced to help solve those issues.
+
+# Description
+
+RadioButtons is a new control that enables you to create related lists of RadioButtons easily, while also fully supporting keyboarding and narrator functionality.
```
StephenLPeters: I think it is confusing that RadioButtons is control, but we are refering to collections of objects of the RadioButton class as RadioButtons in this spec as well. Perhaps we could be a little more verbose to avoid this confusion by replacing the RadioButtons which do not refer to the new control with "Groups of RadioButton objects",

StephenLPeters: I think it is confusing that RadioButtons is a control, but we are referring to collections of objects of the RadioButton class as RadioButtons in this spec as well. Perhaps we could be a little more verbose to avoid this confusion by replacing the RadioButtons which do not refer to the new control with something like "RadioButton Elements".

kikisaints: I see two comments of your that look very similar - are you suggesting either using "RadioButton Elements" or "Groups of RadioButton objects"? Either way I think I understand what you're getting at and will try and re-word it to be less confusing.

StephenLPeters: Sorry, I thought I deleted the other one. Personally I think RadioButton Elements is cleaner.  The main point is that because of the name of this control I don't think we should use RadioButtons as the plural of RadioButton, since we are trying to name this new control RadioButtons.

kikisaints: Makes sense! I just went through and updated the wording in those sections - take a look and let me know if that helps make it clearer.

active/radiobuttons/RadioButtonsAPI.md
```diff
@@ -0,0 +1,155 @@
+# Background
+
+The RadioButtons control is our answer to enabling better keyboarding and narrator support for RadioButtons that need to be presented in a related list.
```
StephenLPeters:
```suggestion
The RadioButtons control is our answer to enabling better keyboarding and narrator support for RadioButton elements that need to be presented in a related list.
```

active/radiobuttons/RadioButtonsAPI.md
StephenLPeters: Just one more, made a separate comment with the suggested change.

kikisaints: Noted! I committed them.

active/radiobuttons/RadioButtonsAPI.md
```diff
@@ -0,0 +1,155 @@
...
+        radioButtonItems.Add(new OptionDataModel() { label = "Item 3" });
+    }
+}
+```
+
+### Multiple Columns
+Some groups of RadioButton elements may want a multi-column layout. This is an example on how to set that up.
+
+```xaml
+<preview:RadioButtons x:Name="RadioButtonGroup" Header="App Mode" MaximumColumns="3">
```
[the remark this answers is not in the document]
kikisaints: @teaP Something you can look at as well?

active/radiobuttons/RadioButtonsAPI.md
```diff
@@ -0,0 +1,155 @@
...
+```
+
+# API Notes
+
+| Name | Description |
+|:-:|:--|
+| Header | Places a text label above the group of RadioButton elements and is read out by the narrator on focus. |
+| HeaderTemplate | Identifies the Header dependency property. |
+| SelectedItem | Gets or sets the index of the selected item. Selection is denoted by that item's RadioButton icon being checked. |
+| SelectedIndex | Gets or sets the selected item.|
```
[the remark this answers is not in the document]
kikisaints: Was calling out that the selection behavior is slightly different from the selected item behavior you get in a listview - are you suggestion this is too redundant and I should remove?

active/radiobuttons/RadioButtonsAPI.md
```diff
@@ -0,0 +1,155 @@
...
+
+Currently, the only way to group single RadioButton elements today is through using a StackPanel and individually labeling each RadioButton through its GroupName property in order for narrator to read out the options somewhat understandably.
+
+However, using that method still leaves gaps in the narrator experience and StackPanel does not behave the way users expect when navigating the list via keyboard - RadioButtons was introduced to help solve those issues.
+
+# Description
+
+RadioButtons is a new control that enables you to create groups of RadioButton elements easily, while also correctly supporting keyboarding and narrator functionality.
+
+# Examples
```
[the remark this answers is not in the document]
kikisaints: This unfortunately doesn't really work all that well. We get this:
image

@teaP - this seems like a bug, is it known?

active/radiobuttons/RadioButtonsAPI.md
```diff
@@ -0,0 +1,155 @@
...
+
+        radioButtonItems = new List<OptionDataModel>();
+        radioButtonItems.Add(new OptionDataModel() { label = "Item 1" });
+        radioButtonItems.Add(new OptionDataModel() { label = "Item 2" });
+        radioButtonItems.Add(new OptionDataModel() { label = "Item 3" });
+    }
+}
+```
+
+### Multiple Columns
```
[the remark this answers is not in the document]
kikisaints: The reason for this specific layout when setting multiple columns was to imitate the common and known behavior keyboard users recognized from Win32. @chigy did some extensive research here about why the layout should follow this.

We were also running into consistency issues with our app team partners about spacing and column formatting when they were using StackPanels + RadioButtons in their Settings pages. That's the main reason we decided to leave out too much customization, to keep a pattern with this control that is consistent for everyone.

active/radiobuttons/RadioButtonsAPI.md
```diff
@@ -0,0 +1,155 @@
...
+    RadioButtonsListViewItem();
+}
+}
+```
+
+# API Notes
+
+| Name | Description |
+|:-:|:--|
+| Header | Places a text label above the group of RadioButton elements and is read out by the narrator on focus. |
```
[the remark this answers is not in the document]
kikisaints: I made a change to this, please check for accuracy.

active/radiobuttons/RadioButtonsAPI.md
```diff
@@ -0,0 +1,155 @@
...
+}
+```
+
+### RadioButtons Primitives
+
+```
+namespace Microsoft.UI.Xaml.Controls.Primitives
+{
+[webhosthidden]
+unsealed runtimeclass RadioButtonsListView : Windows.UI.Xaml.Controls.ListView
```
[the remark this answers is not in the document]
kikisaints: Not sure exactly what you're referring to... the webhosthidden? @teaP to help.

active/radiobuttons/RadioButtonsAPI.md
```diff
@@ -0,0 +1,155 @@
...
+[contentproperty("Items")]
+unsealed runtimeclass RadioButtons : Windows.UI.Xaml.Controls.Control
+{
+    RadioButtons();
+
+    Object ItemsSource;
+    Windows.Foundation.Collections.IVector<Object> Items{ get; };
+    Windows.UI.Xaml.DataTemplate ItemTemplate;
+
+    Windows.UI.Xaml.DependencyObject ContainerFromItem(Object item);
```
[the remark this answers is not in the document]
kikisaints: @teaP - to confirm but I thought the container type was RadioButtonListViewItem?

active/radiobuttons/RadioButtonsAPI.md
```diff
@@ -0,0 +1,182 @@
...
+
+```xaml
+<muxc:RadioButtons Header="App Mode" SelectedIndex="2">
+    <x:String>Dark</x:String>
+    <x:String>Light</x:String>
+    <x:String>Windows Default</x:String>         
+</muxc:RadioButtons>
+```
+
+![alt text](DefaultRadioButtonGroup.png)
```
YuliKl: Can you update the image to show Dark/Light/Windows Default, instead of Item1/Item2/Items3? And actually show the selected index as selected.

active/radiobuttons/RadioButtonsAPI.md
```diff
@@ -0,0 +1,182 @@
...
+
+## DataBinding with RadioButtons
+An example of simple data binding of the ItemsSource with RadioButtons.
+
+```xaml
+<preview:RadioButtons Header="App Mode" ItemsSource="{x:Bind radioButtonItems}" />
```
YuliKl: I assume the preview namespace is temporary, so...
```suggestion
<muxc:RadioButtons Header="App Mode" ItemsSource="{x:Bind radioButtonItems}" />
```

active/radiobuttons/RadioButtonsAPI.md
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    List<OptionDataModel> radioButtonItems;
+
+    public MainPage()
+    {
+        this.InitializeComponent();
+
+        radioButtonItems = new List<OptionDataModel>();
+        radioButtonItems.Add(new OptionDataModel() { label = "Item 1" });
```
YuliKl: I'm guessing you're trying to show the same app mode example as above, just using data binding. So perhaps use the dark/light strings from the previous example?

active/radiobuttons/RadioButtonsAPI.md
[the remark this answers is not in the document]
YuliKl: Are the descriptions for SelectedItem and SelectedIndex backwards?

active/radiobuttons/RadioButtonsAPI.md
```diff
@@ -0,0 +1,182 @@
...
+| Name | Description |
+|:-:|:--|
+| Header | Places a text label above the group of RadioButton elements and is exposed to the UIA tree via the DataContext property on the RadioButtons control's ContentPresenter. |
+| HeaderTemplate | The template that specifies the visualization of the header object |
+| SelectedItem | Gets or sets the index of the selected item. Selection is denoted by that item's RadioButton icon being checked. |
+| SelectedIndex | Gets or sets the selected item.|
+| Items | Gets or sets an object source used to generate the content of the ItemsControl. All items placed within the RadioButtons control will get the content of a generated RadioButton. |
+| ItemsSource | Gets or sets an object source used to generate the content of the ItemsControl. |
+| ItemTemplate | The template set on any generated RadioButton elements within the RadioButtons control. If the root of the ItemTemplate is a RadioButton it will be used as the generated container. |
+| MaxColumns | Defines the number of columns to divide the RadioButton group items into. |
```
YuliKl:
```suggestion
| MaxColumns | Defines the number of columns used to determine the control's layout |
```

active/radiobuttons/RadioButtonsAPI.md
```diff
@@ -0,0 +1,182 @@
...
+
+| Name | Description |
+|:-:|:--|
+| Header | Places a text label above the group of RadioButton elements and is exposed to the UIA tree via the DataContext property on the RadioButtons control's ContentPresenter. |
+| HeaderTemplate | The template that specifies the visualization of the header object |
+| SelectedItem | Gets or sets the index of the selected item. Selection is denoted by that item's RadioButton icon being checked. |
+| SelectedIndex | Gets or sets the selected item.|
+| Items | Gets or sets an object source used to generate the content of the ItemsControl. All items placed within the RadioButtons control will get the content of a generated RadioButton. |
+| ItemsSource | Gets or sets an object source used to generate the content of the ItemsControl. |
+| ItemTemplate | The template set on any generated RadioButton elements within the RadioButtons control. If the root of the ItemTemplate is a RadioButton it will be used as the generated container. |
```
YuliKl: Perhaps an example of using ItemTemplate would be helpful?

active/radiobuttons/RadioButtonsAPI.md
```diff
@@ -0,0 +1,182 @@
...
+| MaxColumns | Defines the number of columns to divide the RadioButton group items into. |
+
+## RadioButton Elements in the RadioButtons controls
+Properties set on a RadioButton that is placed in a RadioButtons control are ignored by the narrator.
```
YuliKl:
```suggestion
Properties set on a RadioButton that is placed in a RadioButtons control are ignored by Narrator.
```

active/radiobuttons/RadioButtonsAPI.md
```diff
@@ -0,0 +1,182 @@
...
+
+For example, in this scenario:
+
+```xaml
+<muxc:RadioButtons Header="App Mode">
+    <RadioButton GroupName="group test">test 1</RadioButton>
```
YuliKl: It definitely wasn't clear to me from earlier in the spec that this scenario (adding RadioButton controls into RadioButtons control) was supported. Not sure how that can be clarified earlier.

active/radiobuttons/RadioButtonsAPI.md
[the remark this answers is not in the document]
kikisaints: Good catch! Look to be.

active/radiobuttons/RadioButtonsAPI.md
kikisaints: hmm, I already made this change... not sure why it didn't show up - but yes! it should all be muxc.

active/radiobuttons/RadioButtonsAPI.md
kikisaints: This should be in the examples above, *however* there appears to be a bug with it and I need @teaP to help address them before I add more examples to this spec.

active/radiobuttons/RadioButtonsAPI.md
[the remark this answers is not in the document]
teaP: It's not a bug, you just can't do this.

active/radiobuttons/RadioButtonsAPI.md
teaP: This scenario is not supported. Is this a new requirement?

active/RadioButtonsAPI.md
MikeHillberg: +1 to Primitives. (This Primitives pattern has issues, but nice to keep them all in one place and potentially fix them at the same time.)

active/radiobuttons/RadioButtonsAPI.md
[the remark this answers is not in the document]
MikeHillberg: I was just saying to have the "Selection is denoted by that item's RadioButton icon being checked" in both SelectedItem and SelectedIndex descriptions.

active/radiobuttons/RadioButtonsAPI.md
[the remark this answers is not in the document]
MikeHillberg: Does the scenario make sense?  It's what apps have to do today.

active/radiobuttons/RadioButtonsAPI.md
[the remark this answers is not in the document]
kikisaints: Ah! Okay will fix.

active/radiobuttons/RadioButtonsAPI.md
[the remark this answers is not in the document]
MikeHillberg: No, why is there a RadioButtonsListView and RadioButtonsAutomationPeer?  I think we should add a brief description of these in a "Remarks" section, after the Examples section.

active/radiobuttons/RadioButtonsAPI.md
```diff
@@ -0,0 +1,181 @@
...
+| SelectedIndex | Gets or sets the selected index. Selection is denoted by that item's RadioButton icon being checked.|
+| Items | Gets or sets an object source used to generate the content of the ItemsControl. All items placed within the RadioButtons control will get the content of a generated RadioButton. |
+| ItemsSource | Gets or sets an object source used to generate the content of the ItemsControl. |
+| ItemTemplate | The template set on any generated RadioButton elements within the RadioButtons control. If the root of the ItemTemplate is a RadioButton it will be used as the generated container. |
+| MaxColumns | Defines the number of columns used to determine the control's layout |
+
+## RadioButton Elements in the RadioButtons controls
+Properties set on a RadioButton that is placed in a RadioButtons control are ignored by Narrator.
+
+For example, in this scenario:
```
MikeHillberg: Why do we have an example that's not in the examples section?

active/radiobuttons/RadioButtonsAPI.md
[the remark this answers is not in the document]
MikeHillberg: No, I mean we're not giving any explanation for the RibbonsListView API, or the Ribbon...Peer APIs.  I know they're mostly implementation details, I would just describe them in a "Remarks" section after the Examples section.

active/radiobuttons/RadioButtonsAPI.md
MikeHillberg: Why do we have an example not in the Examples section?

active/radiobuttons/RadioButtonsAPI.md
```diff
@@ -0,0 +1,181 @@
...
+
+```xaml
+<muxc:RadioButtons Header="App Mode">
+    <RadioButton GroupName="group test">test 1</RadioButton>
```
MikeHillberg: We don't need to demonstrate the useless case of setting GroupName when using RadioButtons.  I would take these out, but move the note to the Remarks section.

active/radiobuttons/RadioButtonsAPI.md
```diff
@@ -0,0 +1,181 @@
...
+| ItemTemplate | The template set on any generated RadioButton elements within the RadioButtons control. If the root of the ItemTemplate is a RadioButton it will be used as the generated container. |
+| MaxColumns | Defines the number of columns used to determine the control's layout |
+
+## RadioButton Elements in the RadioButtons controls
+Properties set on a RadioButton that is placed in a RadioButtons control are ignored by Narrator.
+
+For example, in this scenario:
+
+```xaml
+<muxc:RadioButtons Header="App Mode">
```
MikeHillberg: To make this more realistic, each RadioButton should have its IsChecked property set to a different property (and be a Mode=TwoWay x:Bind)

active/radiobuttons/RadioButtonsAPI.md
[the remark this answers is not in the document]
kikisaints: It's apparently not a scenario we can target at this time. Removing references to in in API doc.

active/radiobuttons/RadioButtons_Spec.md
```diff
@@ -0,0 +1,252 @@
...
+
+![alt text](keyboardnav_multicol.png)
+
+## Selection follows Focus
+When you are navigating a RadioButtons list via the keyboard, as focus is placed on a previous or next RadioButton item, that item will also get selected/checked. This means the previous item that was selected/checked will be unselected, and the currently focused one will be instead.
+
+|Before keyboard navigation | After keyboard navigation|
+|:--|:--|
+| ![alt text](2selected_nonav.png) | ![alt text](3selected_yesnav.png)|
+| Focus is on the "2" RadioButton, and it is shown as selected | The down or right arrow key has been pressed, so focus was moved to the "3" RadioButton, thus selected 3 and unselected 2. |
```
StephenLPeters: >right [](start = 77, length = 5)

In this single column case, pressing right should not change focus, correct?

kikisaints: I believe that that is what we settled on in the last meeting @chigy to confirm.

Once I get confirmation, I'll edit this API.

chigy: > I believe that that is what we settled on in the last meeting @chigy to confirm.
> 
> Once I get confirmation, I'll edit this API.

Based on the old thread, this is what we settled last so sounds right to me.
-	Have up arrow move previous and down arrow move next
-	Left and right arrow keys move spatially

active/radiobuttons/RadioButtonsAPI.md
kikisaints: It should be now with the migration from ListView to ItemsRepeater. Resolving this.

active/radiobuttons/RadioButtonsAPI.md
kikisaints: I believe I added this. Resolving.

active/RadioButtonsAPI.md
kikisaints: Added.

active/radiobuttons/RadioButtons_Spec.md
```diff
@@ -169,84 +169,120 @@ If you find that you need wrapping behavior, perhaps a RadioButtons group isn't
...
+    Object ItemTemplate;
 
-    Windows.UI.Xaml.DependencyObject ContainerFromItem(Object item);
-    Windows.UI.Xaml.DependencyObject ContainerFromIndex(Int32 index);
+    Windows.UI.Xaml.UIElement ContainerFromIndex(Int32 index);
 
+    [MUX_DEFAULT_VALUE("-1")]
     Int32 SelectedIndex;
-    Object SelectedItem;
+    Object SelectedItem{ get; };
```
jevansaks: Why is this get-only?

active/radiobuttons/RadioButtons_Spec.md
```diff
@@ -169,84 +169,120 @@ If you find that you need wrapping behavior, perhaps a RadioButtons group isn't
...
-```
-namespace Microsoft.UI.Xaml.Controls.Primitives
+``` c++
+namespace MU_XCP_NAMESPACE
 {
+[WUXC_VERSION_PREVIEW]
 [webhosthidden]
-unsealed runtimeclass RadioButtonsListView : Windows.UI.Xaml.Controls.ListView
+[default_interface]
+unsealed runtimeclass RadioButtonsElementFactory : MU_XC_NAMESPACE.ElementFactory
```
jevansaks: @StephenLPeters does this type need to be public? If so, why? What do people do with it?

active/radiobuttons/RadioButtons_Spec.md
```diff
@@ -169,84 +169,120 @@ If you find that you need wrapping behavior, perhaps a RadioButtons group isn't
...
 {
-    RadioButtonsListView();
+    RadioButtonsElementFactory();
 }
 
+[WUXC_VERSION_PREVIEW]
 [webhosthidden]
 [default_interface]
-unsealed runtimeclass RadioButtonsListViewItem : Windows.UI.Xaml.Controls.ListViewItem
+unsealed runtimeclass ColumnMajorUniformToLargestGridLayout : MU_XC_NAMESPACE.NonVirtualizingLayout
```
jevansaks: From the PR my understanding was this was internal, do we think it's necessary to be public? Can we just create and use it from code-behind for now?

active/radiobuttons/RadioButtons_Spec.md
StephenLPeters: ItemsRepeater does not  have a way to get a container from a data object. This is by design as a data object might be duplicated in your data set and then repeater would not know which items to return.   Technically we can do this by either walking the tree looking for a radiobutton who's content matches the supplied item, or by accessing private repeater headers, But in general I think collection controls should not expose selection by data item. given that repeater does not support  it.

active/radiobuttons/RadioButtons_Spec.md
StephenLPeters: It is the default value for the ItemTemplate property on RadioButtons. I think we could get away with moving this to the private namespace.

active/radiobuttons/RadioButtons_Spec.md
StephenLPeters: In the PR I'm setting the repeater's layout in the RadioButtons template. We could move this to code behind and not expose this type, but we'd be preventing the consumer from setting their own layout by retemplating. (unless we can use the facny new ReadLocalValue api I just learned about to only set the layout if the consumer hasn't already set one).

StephenLPeters: The other aspect to this, which hasn't been updated in the xaml file cause I didn't give kiki an new one yet (sorry!) is that we are setting the layouts column and row spacing properties to theme resources in the controls resource dictionary. This would mean we'd have to do these theme resource look ups from code behind i think and that's been problematic in the past.

active/radiobuttons/RadioButtons_Spec.md
jevansaks: I think that's a feature gap for Repeater that we should fill. It's ambiguous for ItemsControl too but it's still offered. Binding to SelectedItem is a very popular ask, I don't think we should take that feature away.

active/radiobuttons/RadioButtons_Spec.md
jevansaks: But would it still be in the public API? You couldn't use it from XAML then, but RadioBUttons should be able to set it from its code-behind.

active/radiobuttons/RadioButtons_Spec.md
jevansaks: I see, then having it be public in primitives seems fine.

active/radiobuttons/RadioButtons_Spec.md
StephenLPeters: I think it needs to be a winrt type to be assignable to the dependency property?

active/radiobuttons/RadioButtons_Spec.md
Felix-Dev: @jevansaks @StephenLPeters Absolutely! For proper MVVM use cases, being able to bind to the SelectedItem property I feel would be very much appreciated by us customers. At least I would definitely use it much rather than using the SelectedIndex property. And as you have pointed out, history (Treeview SelectedItem property) shows us that there is a good chance an issue will be opened if the RadioButtons control won't ship with a settable (and bindable) SelectedItem property 😉

active/radiobuttons/RadioButtons_Spec.md
```diff
@@ -0,0 +1,288 @@
...
+```
+
+# API Notes
+
+| Name | Description |
+|:-:|:--|
+| Header | Places a label object above the group of RadioButton elements and is exposed to the UIA tree via the DataContext property on the RadioButtons control's ContentPresenter. |
+| SelectedItem | Gets the content of the selected RadioButton item. Selection is denoted by that item's RadioButton icon being checked. |
+| SelectedIndex | Gets or sets the selected RadioButton item by index. Selection is denoted by that item's RadioButton icon being checked.|
+| Items | Gets or sets an object source used to generate the content of the ItemsControl. All items placed within the RadioButtons control will get the content of a generated RadioButton. |
```
MikeHillberg: This description sounds like ItemsSource, not Items.

active/radiobuttons/RadioButtons_Spec.md
```diff
@@ -0,0 +1,288 @@
...
+unsealed runtimeclass ColumnMajorUniformToLargestGridLayout : MU_XC_NAMESPACE.NonVirtualizingLayout
+{
+    ColumnMajorUniformToLargestGridLayout();
+
+    [MUX_PROPERTY_CHANGED_CALLBACK(TRUE)]
+    Int32 MaximumColumns{ get; set; };
+    static Windows.UI.Xaml.DependencyProperty MaximumColumnsProperty{ get; };
+
+    [MUX_PROPERTY_CHANGED_CALLBACK(TRUE)]
+    Int32 ColumnSpacing{ get; set; };
```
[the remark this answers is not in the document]
jevansaks: Also should the type be Int or Float?

active/radiobuttons/RadioButtons_Spec.md
```diff
@@ -0,0 +1,288 @@
...
+
+The RadioButtons control in this spec is introduced to help solve those issues.
+It's like a ListView control, but the items in the list are RadioButton controls.
+
+# Description
+
+RadioButtons is a new control that enables you to create groups of RadioButton elements easily,
+while also correctly supporting keyboarding and narrator functionality.
+
+# Examples
```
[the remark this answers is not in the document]
kikisaints: I believe so, @StephenLPeters that should be possible, correct? If so I'll go ahead and add an example.

active/radiobuttons/RadioButtons_Spec.md
```diff
@@ -0,0 +1,288 @@
...
+        </Setter.Value>
+    </Setter>
+</Style>
+```
+
+# API Notes
+
+| Name | Description |
+|:-:|:--|
+| Header | Places a label object above the group of RadioButton elements and is exposed to the UIA tree via the DataContext property on the RadioButtons control's ContentPresenter. |
```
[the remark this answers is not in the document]
kikisaints: @StephenLPeters can you confirm whether or not RadioButtons is still doing this in your new implementation? If not, I can remove/update this.

active/radiobuttons/RadioButtons_Spec.md
```diff
@@ -0,0 +1,288 @@
...
+- The up or down arrow keys will move to the "next" or "previous" logical item (defined in markup)
+- The left or right arrow keys will move spatially
+
+![alt text](keyboardnav_singlecol.png)
+
+This means that in purely vertical RadioButton group lists, left/right arrow keys will not do anything, but up/down will still navigate as expected. However, in purely horizontal RadioButton group lists, left and right will navigate the same way as up and down.
+
+>This behavior is defined more below when we discuss multi-column RadioButton group lists.
+
+## Vertical and horizontal RadioButton lists
```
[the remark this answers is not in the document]
kikisaints: Fixed!

active/radiobuttons/RadioButtons_Spec.md
```diff
@@ -0,0 +1,288 @@
...
+    RadioButtons();
+
+    Object ItemsSource;
+    Windows.Foundation.Collections.IVector<Object> Items{ get; };
+    [MUX_DEFAULT_VALUE("winrt::RadioButtonsElementFactory{}")]
+    Object ItemTemplate;
+
+    Windows.UI.Xaml.UIElement ContainerFromIndex(Int32 index);
+
+    [MUX_DEFAULT_VALUE("-1")]
```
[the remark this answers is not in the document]
kikisaints: Removed.

active/radiobuttons/RadioButtons_Spec.md
```diff
@@ -0,0 +1,288 @@
...
+    [MUX_PROPERTY_CHANGED_CALLBACK(TRUE)]
+    Int32 RowSpacing{ get; set; };
+    static Windows.UI.Xaml.DependencyProperty RowSpacingProperty{ get; };
+}
+
+}
+
+```
+
+## RadioButtons.xaml
```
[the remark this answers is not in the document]
kikisaints: renamed.

active/radiobuttons/RadioButtons_Spec.md
```diff
@@ -0,0 +1,288 @@
...
+                        </local:ItemsRepeater.Layout>
+                    </local:ItemsRepeater>
+                </StackPanel>
+                </ControlTemplate>
+        </Setter.Value>
+    </Setter>
+</Style>
+```
+
+# API Notes
```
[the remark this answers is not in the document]
kikisaints: Move to top of API Details section.

active/radiobuttons/RadioButtons_Spec.md
kikisaints: Attempted to redefine it. Please double check for clarity @StephenLPeters 😄

active/radiobuttons/RadioButtons_Spec.md
```diff
@@ -0,0 +1,288 @@
...
+
+    Windows.UI.Xaml.UIElement ContainerFromIndex(Int32 index);
+
+    [MUX_DEFAULT_VALUE("-1")]
+    Int32 SelectedIndex;
+    Object SelectedItem{ get; };
+    event Windows.UI.Xaml.Controls.SelectionChangedEventHandler SelectionChanged;
+
+    [MUX_DEFAULT_VALUE("1")]
+    Int32 MaximumColumns;
```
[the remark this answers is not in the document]
kikisaints: I changed all references from `MaximumColumns` to `MaxColumns` in this API - but please note this does not reflect the implementation yet. @StephenLPeters  please change as necessary. Thanks!

active/radiobuttons/RadioButtons_Spec.md
```diff
@@ -0,0 +1,288 @@
...
+
+## Selection follows Focus
+When you are navigating a RadioButtons list via the keyboard, as focus is placed on a previous or next RadioButton item, that item will also get selected/checked. This means the previous item that was selected/checked will be unselected, and the currently focused one will be instead.
+
+|Before keyboard navigation | After keyboard navigation|
+|:--|:--|
+| ![alt text](2selected_nonav.png) | ![alt text](3selected_yesnav.png)|
+| Focus is on the "2" RadioButton, and it is shown as selected | The down or right arrow key has been pressed, so focus was moved to the "3" RadioButton, thus selected 3 and unselected 2. |
+
+## Narrator behavior
```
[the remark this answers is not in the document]
kikisaints: Renamed section. The keyboard covers ideal keyboarding accessibility behavior, the only other outstanding behavior is narrator/screen reader.

active/radiobuttons/RadioButtons_Spec.md
```diff
@@ -0,0 +1,271 @@
+# Background
+Radio buttons are used for mutually exclusive options in an app. When the users selects
```
YuliKl:
```suggestion
Radio buttons are used for mutually exclusive options in an app. When the user selects
```

active/radiobuttons/RadioButtons_Spec.md
```diff
@@ -0,0 +1,271 @@
...
+one of a group, the previously-checked button is unchecked. Consequently, radio buttons
```
YuliKl:
```suggestion
one in a group, the previously-checked button is unchecked. Consequently, radio buttons
```

active/radiobuttons/RadioButtons_Spec.md
```diff
@@ -0,0 +1,271 @@
...
+That mechanism is awkward, and doesn't by default give the right keyboarding or
+accessibility behavior.
+
+The RadioButtons control in this spec is introduced to help solve those issues.
+It's like a ListView control, but the items in the list are RadioButton controls.
+
+# Description
+
+RadioButtons is a new control that enables you to create groups of RadioButton elements easily,
+while also correctly supporting keyboarding and narrator functionality.
```
YuliKl:
```suggestion
while also correctly supporting keyboarding and Narrator functionality.
```

active/radiobuttons/RadioButtons_Spec.md
```diff
@@ -0,0 +1,271 @@
...
+
+# Examples
+
+The following example demonstrates how to create a small collective group of RadioButton elements
```
YuliKl:
```suggestion
The following example demonstrates how to create a group of RadioButton elements
```

active/radiobuttons/RadioButtons_Spec.md
```diff
@@ -0,0 +1,271 @@
...
+
+## Vertical and horizontal RadioButton lists
+In instances where there is more that one column of RadioButton items within a single RadioButtons list control, the list will flow a column-major layout.
+
+|MaxColumns not specified <br>OR<br> MaxColumns = 1 |  MaxColumns = 4 |
+|:--:|:--:|
+| ![alt text](singlecolumn_example.png) | ![alt text](multicolumn_example_1.png) |
+
+```xml
+<RadioButtons Header="Select Number" MaxColumns="2">
```
YuliKl: Does this belong in the next section, since it specifies 2 columns for 4 items?

active/radiobuttons/RadioButtons_Spec.md
```diff
@@ -0,0 +1,271 @@
...
+        radioButtonItems = new List<OptionDataModel>();
+        radioButtonItems.Add(new OptionDataModel() { Label = "Item 1" });
+        radioButtonItems.Add(new OptionDataModel() { Label = "Item 2" });
+        radioButtonItems.Add(new OptionDataModel() { Label = "Item 3" });
+    }
+}
+```
+
+## Vertical and horizontal RadioButton lists
+In instances where there is more that one column of RadioButton items within a single RadioButtons list control, the list will flow a column-major layout.
```
YuliKl:
```suggestion
In instances where there is more that one column of RadioButton items within a single RadioButtons list control, the list will flow in a column-major layout.
```

Maybe I'm the only one confused by this column-major wording... I find the "fill itself vertically" statement easier to understand.

active/radiobuttons/RadioButtons_Spec.md
```diff
@@ -0,0 +1,271 @@
...
+    <x:String>3</x:String>
+    <x:String>4</x:String>
+</RadioButtons>
+```
+
+By default, the RadioButtons list will fill itself vertically, meaning if no ``MaxColumns`` value is set, it will be assumed to be 1 and orient itself vertically.
+
+### Multiple columns (uncommon case)
+
+If you specify a ``MaxColumns`` value that is not equal to the number of items in the RadioButtons list, the list will arrange  itself in column-major order and put any uneven/remaining items in the first column, in the case of an uneven ``MaxColumns`` value or list items defined.
```
YuliKl: I was following along but got lost in this sentence by the time I got to "in the case of an uneven ``MaxColumns`` value or list items defined." 😅

active/radiobuttons/RadioButtons_Spec.md
```diff
@@ -0,0 +1,271 @@
...
+    <x:String>3</x:String>
+    <x:String>4</x:String>
+    <x:String>5</x:String>
+    <x:String>6</x:String>
+    <x:String>7</x:String>
+</RadioButtons>
+```
+
+# Remarks
+The RadioButtons control has special navigation behavior that helps not only keyboard accelerant users, but also accessibility users to navigate the list more quickly and more easily.
```
YuliKl:
```suggestion
The RadioButtons control has special navigation behavior that helps keyboard users navigate the list more quickly and more easily.
```

active/radiobuttons/RadioButtons_Spec.md
```diff
@@ -0,0 +1,271 @@
...
+
+## Putting focus into the RadioButtons list
+There are two states that the RadioButtons control can have with it's RadioButton content:
```
YuliKl:
```suggestion
There are two states that the RadioButtons control can have with its RadioButton content:
```

active/radiobuttons/RadioButtons_Spec.md
```diff
@@ -0,0 +1,271 @@
...
+When you have a single column list of RadioButton control options and you have already put focus onto an item, you can navigate the RadioButtons list in logical sequential order using the keyboard.
+
+- The up or down arrow keys will move to the "next" or "previous" logical item (defined in markup)
+- The left or right arrow keys will move spatially
+
+![alt text](keyboardnav_singlecol.png)
+
+This means that in purely vertical RadioButton group lists, left/right arrow keys will not do anything, but up/down will still navigate as expected. However, in purely horizontal RadioButton group lists, left and right will navigate the same way as up and down.
+
+>This behavior is defined more below when we discuss multi-column RadioButton group lists.
```
YuliKl: This note doesn't really feel necessary to me.

active/radiobuttons/RadioButtons_Spec.md
```diff
@@ -0,0 +1,271 @@
...
+
+### Keyboard navigation with multiple columns
+The keyboarding behavior is the same as the single-column navigation, it just wraps to the next column when there is more than one defined.
+
+![alt text](keyboardnav_multicol.png)
+
+## Selection follows focus
+When you are navigating a RadioButtons list via the keyboard, as focus is placed on a previous or next RadioButton item, that item will also get selected/checked. This means the previous item that was selected/checked will be unselected, and the currently focused one will be instead.
```
YuliKl:
```suggestion
When you are navigating a RadioButtons list via the keyboard, as focus is placed on a previous or next RadioButton item, that item will also get selected/checked. This means the previous item that was selected/checked will be unselected.
```

active/radiobuttons/RadioButtons_Spec.md
```diff
@@ -0,0 +1,271 @@
...
+
+|Before keyboard navigation | After keyboard navigation|
+|:--|:--|
+| ![alt text](2selected_nonav.png) | ![alt text](3selected_yesnav.png)|
+| Focus is on the "2" RadioButton, and it is shown as selected | The down or right arrow key has been pressed, so focus was moved to the "3" RadioButton, thus selected 3 and unselected 2. |
```
YuliKl:
```suggestion
| Focus is on the "2" RadioButton, and it is shown as selected | The down arrow key has been pressed, so focus was moved to the "3" RadioButton, thus selected 3 and unselected 2. |
```

Right arrow in this example will do nothing, correct?

active/radiobuttons/RadioButtons_Spec.md
```diff
@@ -0,0 +1,271 @@
...
+
+### No Selection follows focus with gamepad
```
YuliKl:
```suggestion
### Navigating with gamepad
```

active/radiobuttons/RadioButtons_Spec.md
```diff
@@ -0,0 +1,271 @@
...
+If you're navigating the RadioButtons container with a gamepad controller, the selection follows focus behavior will be disabled. When navigating a RadioButtons control, no gamepad engagement is necessary to initiate navigation or to select RadioButton elements within the group.
+
+## Accessibility behavior
+
+Below is a detailed table of, depending on the user interaction, what the narrator is expected to say.
+
+|Upon *first* putting focus into the group | Navigating to a selected item |
+|:--|:--|
+| "Group name" RadioButton collection, x of N selected | RadioButton "name" selected, x of N |
+|"Group name" RadioButton collection control, none selected| RadioButton "name" not selected, x of N <br> *(If navigating with shift-arrow keys, meaning no selection following focus)* |
```
YuliKl: Is the word "control" deliberately included in this example but not the one above it?

active/radiobuttons/RadioButtons_Spec.md
```diff
@@ -0,0 +1,271 @@
...
+
+## No Wrapping
```
YuliKl: This is kind of an odd section title, don't you think? Maybe should be included as a subsection in a Keyboard Navigation section instead?

active/radiobuttons/RadioButtons_Spec.md
```diff
@@ -0,0 +1,271 @@
...
+The RadioButtons group does not wrap. This is because when using a screen reader, a sense of boundary and clear indication of ends versus beginnings is lost and thus making it difficult for vision accessible users to navigate the list easily. There is also no enumeration with a RadioButtons group, since groups like this are meant to have a resonable number of items within.
```
YuliKl:
```suggestion
The RadioButtons group does not wrap. This is because when using a screen reader, a sense of boundary and clear indication of ends versus beginnings is lost and thus making it difficult for blind users to navigate the list easily. There is also no enumeration with a RadioButtons group, since groups like this are meant to have a resonable number of items within.
```

I've heard that "blind" is the more appropriate term

YuliKl: What does the "no enumeration" comment mean? Is that not the "x of N" mentioned previously?

active/radiobuttons/RadioButtons_Spec.md
```diff
@@ -0,0 +1,271 @@
...
+
+If you find that you need wrapping behavior, perhaps a RadioButtons group isn't for you. Consider using a [ListView with multiple selection enabled](https://docs.microsoft.com/en-us/windows/uwp/design/controls-and-patterns/listview-and-gridview#item-selection-and-interaction).
```
YuliKl: Not sure I would give this advice... maybe just remove this entirely?

active/radiobuttons/RadioButtons_Spec.md
```diff
@@ -0,0 +1,271 @@
...
+| Name | Description |
+|:-|:--|
+| Header | Places a label object above the group of RadioButton elements and is exposed to the UIA tree via the DataContext property on the RadioButtons control's ContentPresenter. |
+| SelectedItem | Gets the content of the selected RadioButton item. Selection is denoted by that item's RadioButton icon being checked. |
+| SelectedIndex | Gets or sets the selected RadioButton item by index. Selection is denoted by that item's RadioButton icon being checked.|
+| Items | Gets or sets an object source within the RadioButtons group. All items placed within the RadioButtons control will be in the content of a generated RadioButton. |
+| ItemsSource | Gets or sets an object source used to generate the content of the ItemsControl. |
+| ItemTemplate | The template set on any generated RadioButton elements within the RadioButtons control. If the root of the ItemTemplate is a RadioButton it will be used as the generated container. |
+| MaxColumns | Defines the number of columns used to determine the control's layout |
+| ColumnMajorUniformToLargestGridLayout | To be defined. |
```
YuliKl: ?

active/radiobuttons/RadioButtons_Spec.md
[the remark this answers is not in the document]
MikeHillberg: Should be a double like elsewhere (like [Grid.ColumnSpacing](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.Controls.Grid.ColumnSpacing)).

active/radiobuttons/RadioButtons_Spec.md
kikisaints: We define and use column-major in the docs [here](https://docs.microsoft.com/en-us/windows/uwp/design/input/keyboard-interactions#wrapping-homogeneous-list-and-grid-view-items) - so I'm attempting to use the same terminology in this doc as well.

active/radiobuttons/RadioButtons_Spec.md
kikisaints: I need to sync with Stephen on what this does.

active/radiobuttons/RadioButtons_Spec.md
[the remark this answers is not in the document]
MikeHillberg: Sorry, my question has a typo and I'm not sure what the original question was.

active/radiobuttons/RadioButtons_Spec.md
```diff
@@ -0,0 +1,288 @@
...
+That mechanism is awkward, and doesn't by default give the right keyboarding or
+accessibility behavior.
+
+The RadioButtons control in this spec is introduced to help solve those issues.
+It's like a ListView control, but the items in the list are RadioButton controls.
+
+# Description
+
+RadioButtons is a new control that enables you to create groups of RadioButton elements easily,
+while also correctly supporting keyboarding and narrator functionality.
```
MikeHillberg: I'm sure we don't have any ties specifically to Narrator in the implementation, so we shouldn't say so here. (Talk about accessibility in general, not Narrator in particular.)

active/radiobuttons/RadioButtons_Spec.md
```diff
@@ -0,0 +1,282 @@
...
+
+## Keyboard navigation
+When you have a single column list of RadioButton control options and you have already put focus onto an item, you can navigate the RadioButtons list in logical sequential order using the keyboard.
+
+- The up or down arrow keys will move to the "next" or "previous" logical item (defined in markup)
+- The left or right arrow keys will move spatially
+
+![alt text](keyboardnav_singlecol.png)
+
+This means that in purely vertical RadioButton group lists, left/right arrow keys will not do anything, but up/down will still navigate as expected. However, in purely horizontal RadioButton group lists, left and right will navigate the same way as up and down.
```
jesbis: What's the goal of this restriction? 
I tried this out and it feels weird to me that up-down navigates for a purely horizontal group but left-right does nothing for a purely vertical group. It's logically consistent as an implementation detail but as an end-user the behavior difference feels a bit weird.

PR #8, opened 2019-02-25
NavigationView top whitespace

YuliKl:
Tiny spec to address https://github.com/Microsoft/microsoft-ui-xaml/issues/47 and a related [internal issue](https://microsoft.visualstudio.com/OS/_workitems/edit/18004204)


active/NavigationView/TopPadding/TopPaddingMargins.md
```diff
@@ -0,0 +1,72 @@
...
+To further adjust the position of NavigationView's header area, override the *NavigationViewMinimalHeaderContentMargin* or *NavigationViewMinimalWithBackButtonHeaderContentMargin* XAML theme resources.  
+
+
+# API Notes
+<!-- Give a one or two line description of each API (type
+and member), or at least the ones that aren't obvious
+from their name.  These descriptions are what show up
+in IntelliSense. -->
+
+IsTitleBarPaddingEnabled - Specifies whether to automatically push NavigationView's interactive content down by the height of the title bar when the NavigationView is a root element of an app that extends into the title bar. The default value is True.
```
YuliKl: Other proposed names for the property: 
IsTitleBarAutoPaddingEnabled
IsExtendedTitleBarPaddingEnabled

active/NavigationView/TopPadding/TopPaddingMargins.md
```diff
@@ -0,0 +1,72 @@
...
+<!-- For example, this section is a place to explain why you're adding this API rather than
+modifying an existing API. -->
+
+<!-- For example, this is a place to provide a brief explanation of some dependent
+area, just explanation enough to understand this new API, rather than telling
+the reader "go read 100 pages of background information posted at ...". -->
+
+Some app developers choose to extend their app content into the title bar using the `CoreApplication.GetCurrentView().TitleBar.ExtendViewIntoTitleBar = true;` API. When developers do so, the shell continues to reserve a visually transparent title bar for window drag operations. This has the effect of making UI elements in the top 32 pixels of the app window impossible to click with pointer input. [NavigationView](https://docs.microsoft.com/en-us/windows/uwp/design/controls-and-patterns/navigationview) draws a back button and/or its hamburger button near the top of the control, potentially underneath the transparent title bar. To help app developers optimize interactivity and avoid frustrated users, starting in RS5 we added logic to automatically provide top padding and avoid hit test occlusion by the system title bar.
```
MikeHillberg: MD is easier if you put it lots of manual carriage returns.  That way diffing works better, and you can put a comment on a line rather than on a paragraph.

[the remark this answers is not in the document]
YuliKl: This is the background paragraph that won't make it into public documentation. We don't hardcode the 32px value anywhere and I don't think mentioning the number in this context does any harm.

active/NavigationView/TopPadding/TopPaddingMargins.md
```diff
@@ -0,0 +1,44 @@
...
+When developers do so, the shell continues to reserve a visually transparent title bar for window drag operations. 
+This has the effect of making UI elements in the top 32 pixels of the app window impossible to click with pointer input. [NavigationView](https://docs.microsoft.com/en-us/windows/uwp/design/controls-and-patterns/navigationview) draws a back button and/or its hamburger button near the top of the control, potentially underneath the transparent title bar. 
+To help app developers optimize interactivity and avoid frustrated users, starting in RS5 we added logic to automatically provide top padding and avoid hit test occlusion by the system title bar.
+
+While this feels like the correct default behavior, we do not currently provide a mechanism for app developers to opt out of this automatic padding. 
+App devs can `Window.Current.SetTitleBar` to their prefered UIElement, thus removing the hit test occlution by the transparent title bar across the entire window. 
+And although this resolves the interaction problem, NavigationView's padding remains, providing now unwanted whitespace.
+
+
```
licanhua: It will be great if we adding a picture of auto padding look like.

active/NavigationView/TopPadding/TopPaddingMargins.md
```diff
@@ -0,0 +1,44 @@
...
+## Top whitespace
+Some apps choose to [customize their window's title bar](https://docs.microsoft.com/en-us/windows/uwp/design/shell/title-bar), potentially extending their app content into the title bar area. 
+When NavigationView is the root element in apps that extend into the title bar, the control automatically adjusts the position of its interactive elements to prevent overlap with [the draggable region](https://docs.microsoft.com/en-us/windows/uwp/design/shell/title-bar#draggable-regions). 
+If your app handles the draggable region of the title bar and you would prefer to have the back and menu buttons draw closer to the top of your app window, set `IsTitleBarPaddingEnabled` to False.
+
+````Xaml
+<muxc:NavigationView x:Name="NavView" IsTitleBarPaddingEnabled="False">
+````
```
licanhua: What's the guidance/Example if IsTitleBarPaddingEnabled=False? Backbutton may be not clickable, we need to workaround the problem. I suggest we provide an example here which will make BackButton clickable even if it's on titlebar area.

active/NavigationView/TopPadding/TopPaddingMargins.md
```diff
@@ -0,0 +1,72 @@
+> See comments in Markdown for how to use this spec template
```
StephenLPeters: This line is showing up in your rich diff.

active/NavigationView/TopPadding/TopPaddingMargins.md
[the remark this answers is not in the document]
MikeHillberg: This won't go into the DMC documentation, but it will continue to live here.  We just don't want to encourage hard-coding.

active/NavigationView/TopPadding/TopPaddingMargins.md
```diff
@@ -0,0 +1,44 @@
...
+
+# Description
+
+App developers have greater control over [customizing](https://docs.microsoft.com/en-us/windows/uwp/design/controls-and-patterns/navigationview#navigation-view-customization) the margins and padding near the top of their instance of NavigationView. 
+
+# Navigation view customization
+
+
+## Top whitespace
+Some apps choose to [customize their window's title bar](https://docs.microsoft.com/en-us/windows/uwp/design/shell/title-bar), potentially extending their app content into the title bar area.
```
MikeHillberg: It doesn't "potentially" extend, it really extends, right?

active/NavigationView/TopPadding/TopPaddingMargins.md
```diff
@@ -0,0 +1,44 @@
...
+
+App developers have greater control over [customizing](https://docs.microsoft.com/en-us/windows/uwp/design/controls-and-patterns/navigationview#navigation-view-customization) the margins and padding near the top of their instance of NavigationView. 
+
+# Navigation view customization
+
+
+## Top whitespace
+Some apps choose to [customize their window's title bar](https://docs.microsoft.com/en-us/windows/uwp/design/shell/title-bar), potentially extending their app content into the title bar area. 
+When NavigationView is the root element in apps that extend into the title bar, the control automatically adjusts the position of its interactive elements to prevent overlap with [the draggable region](https://docs.microsoft.com/en-us/windows/uwp/design/shell/title-bar#draggable-regions). 
+If your app handles the draggable region of the title bar and you would prefer to have the back and menu buttons draw closer to the top of your app window, set `IsTitleBarPaddingEnabled` to False.
```
MikeHillberg: What does it mean for my app to "handle the draggable region of the title bar"?  I think it means if my app "calls Window.SetTitleBar".  We should be explicit, it's not terribly intuitive.

active/NavigationView/TopPadding/TopPaddingMargins.md
```diff
@@ -0,0 +1,44 @@
...
+
+
+# Description
+
+App developers have greater control over [customizing](https://docs.microsoft.com/en-us/windows/uwp/design/controls-and-patterns/navigationview#navigation-view-customization) the margins and padding near the top of their instance of NavigationView. 
+
+# Navigation view customization
+
+
+## Top whitespace
```
MikeHillberg: Picture(s) would be wonderful

active/NavigationView/TopPadding/TopPaddingMargins.md
```diff
@@ -0,0 +1,44 @@
...
+````Xaml
+<muxc:NavigationView x:Name="NavView" IsTitleBarPaddingEnabled="False">
+````
+
+To further adjust the position of NavigationView's header area, override the *NavigationViewMinimalHeaderContentMargin* or *NavigationViewMinimalWithBackButtonHeaderContentMargin* XAML theme resources.  
+
+
+# API Notes
+
+IsTitleBarPaddingEnabled - Specifies whether to automatically push NavigationView's interactive content down by the height of the title bar when the NavigationView is a root element of an app that extends into the title bar.
```
MikeHillberg: Please call out somewhere that this is the NavigationView class.

active/NavigationView/TopPadding/TopPaddingMargins.md
MikeHillberg: I think I like IsTitleBarPaddingEnabled.  But I think we should fix the underlying problem so that this isn't typically necessary.  Would this then need to become a three state, like a nullable<bool>?

active/NavigationView/TopPadding/TopPaddingMargins.md
YuliKl: The article I've linked to describes other customization options, such as specifying the colors of the title bar and caption controls. Extending into the title bar is not the only type of customization. Can you suggest a less-clumsy way of trying to get that point across?

active/NavigationView/TopPadding/TopPaddingMargins.md
YuliKl: @StephenLPeters - seeing this line show up in the doc confused me, so I deleted it along with all the comments. This line isn't part of the spec text.

active/NavigationView/TopPadding/TopPaddingMargins.md
YuliKl: @MikeHillberg - I'm sorry, I don't follow why this may need to be a nullable. NavigationView is already doing the right thing much of the time - getting out of the way of the title bar's back button and ensuring that its own back/hamburger buttons remain clickable. But NavigationView isn't currently allowing apps to get rid of this intelligent auto-padding, not even if the app author really, really wants to. So to me this feels like a binary choice: let NavigationView do its default thing or tell it to stop doing the default thing. What's the third state you think we might need?

active/NavigationView/TopPadding/TopPaddingMargins.md
YuliKl: Adding picture to actual documentation section below.

active/NavigationView/TopPadding/TopPaddingMargins.md
YuliKl: There's an [article](https://docs.microsoft.com/en-us/windows/uwp/design/shell/title-bar) which explains title bar customization in detail. I don't want to duplicate the same information in the NavigationView article.

Personally, I don't recommend extending into the title bar. That was our guidance when NavigationView's Expanded pane used acrylic, which it doesn't anymore. The title bar provides a variety of user benefits (familiarity and consistency with other apps, color hints about whether the app has focus, a place for the app title) and apps shouldn't undertake the decision to overwrite it lightly. We still don't have good design guidance of where the app title ought to appear in a NavigationView that's drawing into the title bar.

I think that documenting how to perform this customization encourages apps to extend into the title bar. And I would prefer not to encourage this.

active/NavigationView/TopPadding/TopPaddingMargins.md
YuliKl: Updated wording.

active/NavigationView/TopPadding/TopPaddingMargins.md
YuliKl: Added a screenshot of the current behavior with title bar padding.

active/NavigationView/TopPadding/TopPaddingMargins.md
licanhua: It's possible that Auto, NoPadding, and AlwayPadding.

MikeHillberg: Yes, I was wondering if there should be TopPadding = on/off/auto.  But think it makes more sense the way you have it; AutoPadding = on/off.

active/NavigationView/TopPadding/TopPaddingMargins.md
MikeHillberg: I see, that makes sense.

active/NavigationView/TopPadding/TopPaddingMargins.md
YuliKl: Proposed name: IsTitleBarAutoPaddingEnabled
Keep as bool

active/NavigationView/TopPadding/TopPaddingMargins.md
```diff
@@ -0,0 +1,46 @@
...
+Some apps choose to [customize their window's title bar](https://docs.microsoft.com/en-us/windows/uwp/design/shell/title-bar), potentially extending their app content into the title bar area. 
+When NavigationView is the root element in apps that extend into the title bar, the control automatically adjusts the position of its interactive elements to prevent overlap with [the draggable region](https://docs.microsoft.com/en-us/windows/uwp/design/shell/title-bar#draggable-regions). 
+![An app extending into the title bar](NavigationView_WithTitleBarPadding.png)
+If your app specifies the draggable region by calling the [Window.SetTitleBar](/uwp/api/windows.ui.xaml.window.settitlebar.md) method and you would prefer to have the back and menu buttons draw closer to the top of your app window, set `IsTitleBarPaddingEnabled` to False.
+
+````Xaml
+<muxc:NavigationView x:Name="NavView" IsTitleBarPaddingEnabled="False">
+````
+
+To further adjust the position of NavigationView's header area, override the *NavigationViewMinimalHeaderContentMargin* or *NavigationViewMinimalWithBackButtonHeaderContentMargin* XAML theme resources.
```
YuliKl: Look into combining this into a single resource so app devs don't need to do math.

jevansaks: Suggest: NavigationViewHeaderContentMargin as the resource name.

active/NavigationView/TopPadding/TopPaddingMargins.md
```diff
@@ -0,0 +1,46 @@
...
+App devs can `Window.Current.SetTitleBar` to their prefered UIElement, thus removing the hit test occlution by the transparent title bar across the entire window. 
+And although this resolves the interaction problem, NavigationView's padding remains, providing now unwanted whitespace.
+
+
+# Description
+
+App developers have greater control over [customizing](https://docs.microsoft.com/en-us/windows/uwp/design/controls-and-patterns/navigationview#navigation-view-customization) the margins and padding near the top of their instance of NavigationView. 
+
+# Navigation view customization
+
```
MikeHillberg: Could you add:

(The following is an update to the [Navigation View Customization](https://docs.microsoft.com/en-us/windows/uwp/design/controls-and-patterns/navigationview#navigation-view-customization) guide.)

active/NavigationView/TopPadding/TopPaddingMargins.md
```diff
@@ -0,0 +1,46 @@
...
+
+## Top whitespace
+Some apps choose to [customize their window's title bar](https://docs.microsoft.com/en-us/windows/uwp/design/shell/title-bar), potentially extending their app content into the title bar area. 
+When NavigationView is the root element in apps that extend into the title bar, the control automatically adjusts the position of its interactive elements to prevent overlap with [the draggable region](https://docs.microsoft.com/en-us/windows/uwp/design/shell/title-bar#draggable-regions).
```
MikeHillberg: "... extend into the title bar **using the [ExtendViewIntoTitleBar](uwp/api/Windows.ApplicationModel.Core.CoreApplicationViewTitleBar.ExtendViewIntoTitleBar.md) API**, the control ..."

active/NavigationView/TopPadding/TopPaddingMargins.md
```diff
@@ -0,0 +1,46 @@
...
+
+App developers have greater control over [customizing](https://docs.microsoft.com/en-us/windows/uwp/design/controls-and-patterns/navigationview#navigation-view-customization) the margins and padding near the top of their instance of NavigationView. 
+
+# Navigation view customization
+
+
+## Top whitespace
+Some apps choose to [customize their window's title bar](https://docs.microsoft.com/en-us/windows/uwp/design/shell/title-bar), potentially extending their app content into the title bar area. 
+When NavigationView is the root element in apps that extend into the title bar, the control automatically adjusts the position of its interactive elements to prevent overlap with [the draggable region](https://docs.microsoft.com/en-us/windows/uwp/design/shell/title-bar#draggable-regions). 
+![An app extending into the title bar](NavigationView_WithTitleBarPadding.png)
```
MikeHillberg: I can't understand the picture, there's no whitespace at the top (the "App Extending into Title Bar" text is right up at the top of the window?)

active/NavigationView/TopPadding/TopPaddingMargins.md
```diff
@@ -0,0 +1,46 @@
...
+If your app specifies the draggable region by calling the [Window.SetTitleBar](/uwp/api/windows.ui.xaml.window.settitlebar.md) method and you would prefer to have the back and menu buttons draw closer to the top of your app window, set `IsTitleBarAutoPaddingEnabled` to False.
+
+````Xaml
+<muxc:NavigationView x:Name="NavView" IsTitleBarAutoPaddingEnabled="False">
+````
+
+To further adjust the position of NavigationView's header area, override the *NavigationViewHeaderMargin* XAML theme resource.
```
MikeHillberg: I like calling this out here. This should be in a "Remarks" section (just add a "# Remarks" header line above this). Is there any more detail we can reference?

active/NavigationView/TopPadding/TopPaddingMargins.md
YuliKl: This is what customers are asking for, yes. The app title "App Extending into Title bar", is positioned as close to the top of the window as if that text was actually in the title bar and drawn by shell.

active/NavigationView/TopPadding/TopPaddingMargins.md
YuliKl: Updated in #37

active/NavigationView/TopPadding/TopPaddingMargins.md
YuliKl: Updated in #37

active/NavigationView/TopPadding/TopPaddingMargins.md
YuliKl: I actually haven't tried out the changes @RBrid made to the control template so don't know exactly what other details to add.

PR #5, opened 2019-01-23
Teaching tip spec

SavoySchuler:
Merging in TeachingTip Spec.

PR #17, opened 2019-03-07
Updated code samples to use .Target property for attaching pointing t…

SavoySchuler:
…ips and moved non-point tips out of Button.

This change is coming through as a separate pull request as I am waiting on the corresponding update to the implementation. This keeps the shared documentation consistent with the publicly available version of the control.

### DIFF
--- a/active/teachingtip/TeachingTipSpec.md
+++ b/active/teachingtip/TeachingTipSpec.md
@@ -438,7 +438,7 @@ enum TeachingTipHeroPlacementMode
     Auto,
 };
 
-enum TeachingTipHeroPointerMode
+enum TeachingTipPointerMode
 {
     Auto,
     On,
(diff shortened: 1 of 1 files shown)


MikeHillberg:
[At this line](https://github.com/Microsoft/microsoft-ui-xaml-specs/pull/17/files#diff-c15898e1aa4a9b40aea964878c93c520R31)

What does it mean to require manual acknowledgement?  Is this picture supposed to have an "OK" button or some such?

SavoySchuler:
@MikeHillberg, both samples above have an [X] in the top right corner whereas the light-dismiss tip does not. Does the follow edit read more clearly?

"A teaching tip can require the user to dismiss it via an "X" button in a top corner or a "Close" button at the bottom. However, a teaching tip may also be light-dismiss enabled in which case there is no dismiss button and the teaching tip will instead dismiss when a user scrolls or interacts with other elements of the application."

SavoySchuler:
@MikeHillberg I have moved this discussion over to our [main PR](https://github.com/Microsoft/microsoft-ui-xaml-specs/pull/5) as this PR is just for the .Target update.

PR #20, opened 2019-03-13
TeachingTip API Review

MikeHillberg:


active/teachingtip/TeachingTipSpec.md
```diff
@@ -0,0 +1,565 @@
...
+
+A teaching tip can point to a specific UI element to enhance contextual clarity of the information it is presenting. 
+
+![A sample app with a teaching tip pointing at the save button. The tip title reads "Saving automatically" and the subtitle reads "We save your changes as you go - so you never have to." There is a close button on the top right corner of the teaching tip.](TeachingTipPointerSample.jpg)
+
+A teaching tip can be used without a pointer when the information that it is presenting is not specific to a UI element. 
+
+![A sample app with a teaching tip in the bottom right corner. The tip title reads "Saving automatically" and the subtitle reads "We save your changes as you go - so you never have to." There is a close button on the top right corner of the teaching tip.](TeachingTipSampleApp.jpg)
+
+A teaching tip can require manual acknowledgement before dismissing or a teaching tip may also be light-dismiss enabled so that it will dismiss when a user scrolls or interacts with other elements of the application.
```
oldnewthing: This paragraph is redundant with the one following.

active/teachingtip/TeachingTipSpec.md
```diff
@@ -0,0 +1,565 @@
...
+        </muxc:TeachingTip>
+    </Button.Resources>
+</Button>
+```
+
+![A sample app with a teaching tip pointing at the save button. The tip title reads "Saving automatically" and the subtitle reads "We save your changes as you go - so you never have to." In the content area of the teaching tip is a CheckBox labeled "Don't show tips at startup" and underneath is text that reads "You can change your tip preferences in Settings if you change your mind" where "Settings" is a link to the app's settings page. There is a close button on the top right corner of the teaching tip.](TeachingTipContentSample.jpg)
+
+### Add buttons
+
+A standard "X" close button is automatically added next to the title of a teaching tip. However, a custom close button can be added to the bottom of a teaching tip by setting the CloseButtonContent property.
```
oldnewthing: This doesn't "add" a custom Close button. It converts the X button into a Close button. Perhaps

> By default, a standard "X" close button is shown next to the title of a teaching tip. The Close button can be customized with the CloseButtonContent property, in which case the button is moved to the bottom of the teaching tip.

active/teachingtip/TeachingTipSpec.md
```diff
@@ -0,0 +1,565 @@
...
+    Subtitle="Please close this tip and review our updated privacy policy and privacy settings."
+    Closing="OnTipClosing">
+</muxc:TeachingTip>
+```
+
+C#
+```C#
+public void OnTipClosing(object sender, TeachingTipClosingEventArgs args)
+{
+    if (args.Reason == TeachingTipCloseReason.CloseButton)
```
oldnewthing: Can a Programmatic close be canceled?

active/teachingtip/TeachingTipSpec.md
```diff
@@ -0,0 +1,565 @@
...
+    static Windows.UI.Xaml.DependencyProperty ActionButtonCommandParameterProperty{ get; };
+
+    static Windows.UI.Xaml.DependencyProperty CloseButtonContentProperty{ get; };
+    static Windows.UI.Xaml.DependencyProperty CloseButtonStyleProperty{ get; };
+    static Windows.UI.Xaml.DependencyProperty CloseButtonCommandProperty{ get; };
+    static Windows.UI.Xaml.DependencyProperty CloseButtonCommandParameterProperty{ get; };
+
+    static Windows.UI.Xaml.DependencyProperty TargetOffsetProperty{ get; };
+    static Windows.UI.Xaml.DependencyProperty IsLightDismissEnabledProperty{ get; };
+	static Windows.UI.XAML.DependencyProperty ShouldConstrainToRootBounds{ get; };
```
oldnewthing: XAML autocorrect strikes again. Also, alignment.

active/teachingtip/TeachingTipSpec.md
```diff
@@ -0,0 +1,565 @@
...
+                <StackPanel>
+                    <CheckBox x:Name="HideTipsCheckBox" Content="Don't show tips at start up" IsChecked="{x:Bind HidingTips, Mode=TwoWay}" />
+                    <TextBlock>You can change your tip preferences in <Hyperlink NavigateUri="app:/item/SettingsPage">Settings</Hyperlink> if you change your mind.</TextBlock>
+                </StackPanel>
+        </muxc:TeachingTip>
+    </Button.Resources>
+</Button>
+```
+
+![A sample app with a teaching tip pointing at the save button. The tip title reads "Saving automatically" and the subtitle reads "We save your changes as you go - so you never have to." In the content area of the teaching tip is a CheckBox labeled "Don't show tips at startup" and underneath is text that reads "You can change your tip preferences in Settings if you change your mind" where "Settings" is a link to the app's settings page. There is a close button on the top right corner of the teaching tip.](TeachingTipContentSample.jpg)
```
oldnewthing: In this example, the Close button gets the "default button" treatment. But in the appendix ("Action button" section), the Action button gets the "default button" treatment. Does the app get to control which button looks default-like?

active/teachingtip/TeachingTipSpec.md
```diff
@@ -0,0 +1,565 @@
...
+}
+```
+
+Here's the result when this Page containing the button and teaching tip shown:
+
+![A sample app with a teaching tip pointing at the save button. The tip title reads "Saving automatically" and the subtitle reads "We save your changes as you go - so you never have to." There is a close button on the top right corner of the teaching tip.](TeachingTipPointerSample.jpg)
+
+### Non-pointing tips
+
+Not all tips relate to an element onscreen. For these scenarios, do not set the Target property and the teaching tip will instead display relative to the edges of the XamlRoot. However, a teaching tip can have the pointer removed while retaining placement relative to a UI element by setting the PointerMode property to "Off".
```
oldnewthing: What if I want the tip to appear in the corner of a container that doesn't fill the root window? This is a common layout for an app. We use it in the XAML Controls Gallery, for example. XAML Controls Gallery Observe that the list of samples does not fill the window. Suppose I want to put a tip in the upper left corner of the samples listview to say "Select a sample from this list by clicking on it."

I cannot use a null Target because that will go to the corner of the XAML root. But if I set `Target=MyPanel` and `PointerMode=Off`, the tip will appear *outside* the target. There are no "inside" alignments. Do I have to specify `TopEdgeAlignedLeft` and set a negative `TargetOffset.Bottom` equal to the `ActualHeight` of the TeachingTip? Or do I set `Target=HeroImage` and ask for `BottomEdgeAlignedLeft`? (Would that mess up accessibility, setting an artificial target?)

active/teachingtip/TeachingTipSpec.md
```diff
@@ -0,0 +1,565 @@
...
+
+A teaching tip can require the user to dismiss it via an "X" button in a top corner or a "Close" button at the bottom. A teaching tip may also be light-dismiss enabled in which case there is no dismiss button and the teaching tip will instead dismiss when a user scrolls or interacts with other elements of the application. Because of this behavior, light-dismiss tips are the best solution when a tip needs to be placed in a scrollable area. 
+
+![A sample app with a light-dismiss teaching tip in the bottom right corner. The tip title reads "Saving automatically" and the subtitle reads "We save your changes as you go - so you never have to."](TeachingTipLightDismissSample.jpg)
+
+
+### Create a teaching tip
+
+Here's the XAML for a pointing teaching tip control that demonstrates the default look of the TeachingTip with a title and subtitle. 
+Note that the teaching tip can appear anywhere in the element tree. In this example below, it's located in a ResourceDictionary.
```
oldnewthing: Later, under "Non-pointing tips", we say that the ResourceDictionary is not part of the element tree. Who's right?

active/teachingtip/TeachingTipSpec.md
```diff
@@ -0,0 +1,565 @@
...
+</muxc:TeachingTip>
+```
+
+Note that in this example the TeachingTip is in the element tree rather than in a ResourceDictionary. This has no effect on behavior; the TeachingTip only displays when opened, and takes up no layout space.
+![A sample app with a teaching tip in the bottom right corner. The tip title reads "Saving automatically" and the subtitle reads "We save your changes as you go - so you never have to." There is a close button on the top right corner of the teaching tip.](TeachingTipSampleApp.jpg)
+
+### Add an offset 
+
+You can control how far a pointing teaching tip is set apart from its target and how far a non-pointing teaching tip is set apart from the the edges of the XamlRoot using the TargetOffset property. Like [Margin](https://docs.microsoft.com/en-us/dotnet/api/system.windows.frameworkelement.margin?view=netframework-4.7.2), TargetOffset has four values – left, right, top, and bottom – so only the relevant values are used. For example, if the target is left of the teaching tip, the TargetOffset.Left will be used. If the teaching tip has no target, the TargetOffset’s Right and Bottom will be used to space it from the bottom right corner of the XamlRoot.
```
oldnewthing: The name "TargetOffset" is strange. It's not the offset of the target. And it's not the offset of the tip relative to the target. And it's not the offset of the tip relative to its default placement. Here, the TargetOffset is positive 80, so I expect that to mean that the tip is 80 pixels southeast of "something", but instead it moves northwest from the corner of the XAML root. Should we just call this `Margin`? Because it describes "how much space to put between the tip and its reference point."

active/teachingtip/TeachingTipSpec.md
```diff
@@ -0,0 +1,565 @@
...
+| Hero Content | * Hero Content is media that stretches to the edges of a tip. <br> * Can be placed at the top or bottom of a tip. <br><br> ![A teaching tip with a sample title and subtitle. Above the title and subtitle is a large border-to-border picture of a giraffe.](HeroContent.PNG) |
+| Scroll Bar | * If the tip's contents are large enough to require scrolling, a scrollbar which will not intersect the Close (X) Button will be added to the content area. <br><br> ![A tall teaching tip with a scrollbar on the right a long string of "Lorem ipsum" text in the body.](ScrollBar.PNG) |
+
+### Behavioral Components
+
+ | Property | Notes |
+|:---:|:---|
+| Opening | * A tip is shown by setting its IsOpen property to true. <br> * Tips will animate on opening. <br> * When a tip does not have enough available window space to fully show in any location [see Placement], it will not open and will instead overwrite IsOpen to false. |
```
oldnewthing: Is there an obvious way to know that the tip refused to open? I can imagine people writing code like this:

```C#
var expensiveResource = CreateExpensiveResource();
tipContent.PutInDancingLights(expensiveResource);
myTip.Closed += (s, e) => expensiveResource.Dispose();
myTip.IsOpen = true;
```

This leaks the expensive resource because the `Closed` event never happens.

active/teachingtip/TeachingTipSpec.md
```diff
@@ -0,0 +1,565 @@
...
+| Closing | * There are three ways to close a tip: set the IsOpen property to false, the user invokes a close button, or the tip is closed via light dismiss. These will return the method used in TeachingTipCloseReason.  <br> * Closing can be deferred by taking a handle to the deferral object in the closing event args. |
```
oldnewthing: Grammar nitpick: Parallel construction.

> There are three ways a tip can close: The program sets the `IsOpen` property to `false`, the user invokes the Close button, or the tip closes due to light dismiss. Use the `TeachingTipCloseReason` to determine which case has occurred.

Second sentence needs some wordsmithing too, because the deferral doesn't defer closing. It defers the decision whether or not to close.

> Closing can be prevented by setting the `Cancel` property to `true`. You can use a deferral to respond asynchronously to the event.

active/teachingtip/TeachingTipSpec.md
```diff
@@ -0,0 +1,565 @@
...
+| Placement | * Placement modes for pointing teaching tips will follow the precedent of Flyout. Full placement mode will be replaced by Center which positions Pointer at the center of the element. <br> * Placement modes for non-pointing tips will include each side, corner, and center of the application window. <br> * The following properties are not preferred in tip placement: <br> &nbsp;&nbsp;&nbsp;&nbsp; * There is not enough space for the tip to show without clipping. <br> &nbsp;&nbsp;&nbsp;&nbsp; * The target is not large enough to maintain the tip's alignment and the Pointer's 12px margin from the edge of the tip. <br> &nbsp;&nbsp;&nbsp;&nbsp; * The target element is too large to maintain edge alignment while keeping the Pointer centered on the target. |
```
oldnewthing: Text here says that `Center` is a placement option, but it's not present in the IDL.

active/teachingtip/TeachingTipSpec.md
```diff
@@ -0,0 +1,565 @@
...
+### Some APIs shown in the following examples are not yet available in the prerelease version of this control. Please see the [Teaching Tip Sample in the XAML Sample Gallery](https://github.com/Microsoft/Xaml-Controls-Gallery/blob/w/saschule/TeachingTipSample/XamlControlsGallery/ControlPages/TeachingTipPage.xaml) for implementation examples that are accurate with the version of this control currently available through prerelease. 
+
+
+The teaching tip has several states, including these notable ones.
+
+A teaching tip can point to a specific UI element to enhance contextual clarity of the information it is presenting. 
+
+![A sample app with a teaching tip pointing at the save button. The tip title reads "Saving automatically" and the subtitle reads "We save your changes as you go - so you never have to." There is a close button on the top right corner of the teaching tip.](TeachingTipPointerSample.jpg)
+
+A teaching tip can be used without a pointer when the information that it is presenting is not specific to a UI element.
```
oldnewthing: Wikipedia says that the pointy thing in a speech balloon is called a "[tail](https://en.wikipedia.org/wiki/Speech_balloon#Speech_bubbles)". This agrees with the terminology used in [our own UI guidelines text](https://docs.microsoft.com/en-us/windows/desktop/uxguide/ctrl-balloons).

active/teachingtip/TeachingTipSpec.md
```diff
@@ -0,0 +1,565 @@
...
+
+#### Keyboarding and Narrator Experience 
+| State | Action | Narrator |
+|:---|:---|:---|
+| Tip first appears and is not in focus | No action is needed invoke the tip. |  Persistent tip: <br> “New tip. It says..." + Title + Subtitle <br><br> Light-dismiss tip: <br> “New light-dismiss tip. It says..." Title + Subtitle | 
+| Tip is first focused on | Tip steals focus and returns focus to the previous element upon closing. | Title + Button in Focus |
+| Tip is tabbed through | Tab Button: <br> Will go through all actionable items, regardless of group, in order. When tab is pressed on the last element in the teaching tip, focus will cycle to the first element in the teaching tip.  <br> <br> Left + Right Arrow Keys: <br> Can be used to navigate between the footer Action and Close buttons if both are present. <br><br> Escape: <br> Will result in closing the tip. | | 
+| Tip is dismissed | 1. X Button is pressed. <br> 2. Close Button is pressed. <br> 3. Action Button is pressed. <br> Tab focus returns to the predecessor. | |
```
oldnewthing: Do we need to inform Narrator what object the tip is pointing to?

active/teachingtip/TeachingTipSpec.md
```diff
@@ -0,0 +1,565 @@
...
+
+C#
+```C#
+public MainPage()
+{
+    this.InitializeComponent();
+
+    if(!HaveExplainedAutoSave())
+    {
+        AutoSaveTip.IsOpen = true;
```
oldnewthing: Can I reconfigure the tip while it is open?

* Change title, subtitle, content, hero content, hero placement.
* Change target, placement, offset, root bounds constraint.
* Switch between light dismiss and explicit dismiss.
* Switch between "X" button and full Close button.
* Add/remove an action button.
* Change the content of the buttons.

active/teachingtip/TeachingTipSpec.md
```diff
@@ -0,0 +1,565 @@
...
+A teaching tip can require manual acknowledgement before dismissing or a teaching tip may also be light-dismiss enabled so that it will dismiss when a user scrolls or interacts with other elements of the application.
+
+A teaching tip can require the user to dismiss it via an "X" button in a top corner or a "Close" button at the bottom. A teaching tip may also be light-dismiss enabled in which case there is no dismiss button and the teaching tip will instead dismiss when a user scrolls or interacts with other elements of the application. Because of this behavior, light-dismiss tips are the best solution when a tip needs to be placed in a scrollable area. 
+
+![A sample app with a light-dismiss teaching tip in the bottom right corner. The tip title reads "Saving automatically" and the subtitle reads "We save your changes as you go - so you never have to."](TeachingTipLightDismissSample.jpg)
+
+
+### Create a teaching tip
+
+Here's the XAML for a pointing teaching tip control that demonstrates the default look of the TeachingTip with a title and subtitle.
```
MikeHillberg: Can we standardize on "targeted/untargeted" rather than "pointing/non-pointing" (to match the API name)?

active/teachingtip/TeachingTipSpec.md
MikeHillberg: "XamlRoot" is still difficult to understand. Maybe we should work with DMC to create a special page on the topic, refer to it more as a description with "the xaml root", and make it a hyperlink?

active/teachingtip/TeachingTipSpec.md
```diff
@@ -0,0 +1,565 @@
...
+
+Here's the result when this Page containing the button and teaching tip shown:
+
+![A sample app with a teaching tip pointing at the save button. The tip title reads "Saving automatically" and the subtitle reads "We save your changes as you go - so you never have to." There is a close button on the top right corner of the teaching tip.](TeachingTipPointerSample.jpg)
+
+### Non-pointing tips
+
+Not all tips relate to an element onscreen. For these scenarios, do not set the Target property and the teaching tip will instead display relative to the edges of the XamlRoot. However, a teaching tip can have the pointer removed while retaining placement relative to a UI element by setting the PointerMode property to "Off". 
+
+Unlike pointing teaching tips which can only be added to a resource dictionary, non-pointing teaching tips can also be added through the element tree, as shown below.
```
MikeHillberg: Targeted or un-targeted, a tip can be in a ResourceDictionary or in the element tree.

active/teachingtip/TeachingTipSpec.md
```diff
@@ -0,0 +1,565 @@
...
+
+XAML
+```XAML
+<Button x:Name="SaveButton" Content="Save" />
+
+<muxc:TeachingTip x:Name="AutoSaveTip"
```
MikeHillberg: The "muxc" xmlns is left as a mystery. I think for all the WinUI examples we need to include more context, to get the xmlns definition in.

active/teachingtip/TeachingTipSpec.md
MikeHillberg: Bad hyperlink in here for Margin; it's pointing to WPF's.

active/teachingtip/TeachingTipSpec.md
SavoySchuler: Test.

active/teachingtip/TeachingTipSpec.md
MikeHillberg: Test

active/teachingtip/TeachingTipSpec.md
SavoySchuler: Fixed by both this [commit ](https://github.com/Microsoft/microsoft-ui-xaml-specs/commit/e28ae9aed3521b2ba28f107fb6c11e73cced415f) and [this ](https://github.com/Microsoft/microsoft-ui-xaml-specs/commit/76b323b15c5d313e086c4702e6275c510360df8b) one.

active/teachingtip/TeachingTipSpec.md
SavoySchuler: [Resolved by deleting Line 31.](https://github.com/Microsoft/microsoft-ui-xaml-specs/commit/15c824806d05a7ffb8634de3869f44abf4b966e0)

active/teachingtip/TeachingTipSpec.md
SavoySchuler: Resolved with [this commit](https://github.com/Microsoft/microsoft-ui-xaml-specs/commit/5e0603d2992cedd4000c838a21f7375d56316942).

active/teachingtip/TeachingTipSpec.md
SavoySchuler: [Resolved.](https://github.com/Microsoft/microsoft-ui-xaml-specs/commit/1656bd37a8ce535be679e8caaae89e577148b106)

active/teachingtip/TeachingTipSpec.md
SavoySchuler: Default button property was cut. All PM art will be updated with real screenshots shortly.

active/teachingtip/TeachingTipSpec.md
SavoySchuler: Resolved with [this commit](https://github.com/Microsoft/microsoft-ui-xaml-specs/commit/48bde1f087b8fc83369eb4c0004f92b20009461f). 

@jevansaks, @MikeHillberg, @StephenLPeters Is there a traditional or precedented place in the enum for "Center" to be at? I couldn't make sense of any particular order with [Flyout's Placement Mode](https://docs.microsoft.com/en-us/uwp/api/Windows.UI.Xaml.Controls.Primitives.FlyoutPlacementMode), but we already appear to disregard that with our current ordering. 

I added it at the end because, from the user perspective, I anticpate Center placement to be the least common/preferred in placement iteration order.

active/teachingtip/TeachingTipSpec.md
SavoySchuler: Fixed by both this [commit ](https://github.com/Microsoft/microsoft-ui-xaml-specs/commit/e28ae9aed3521b2ba28f107fb6c11e73cced415f) and [this ](https://github.com/Microsoft/microsoft-ui-xaml-specs/commit/76b323b15c5d313e086c4702e6275c510360df8b) one.

active/teachingtip/TeachingTipSpec.md
SavoySchuler: Fixed [here](https://github.com/Microsoft/microsoft-ui-xaml-specs/commit/c7b9379cbbe6f11b422d9bd033533bd185998c7b).

active/teachingtip/TeachingTipSpec.md
SavoySchuler: Resolved [here](https://github.com/Microsoft/microsoft-ui-xaml-specs/commit/5be7d4bab21c5c74909927055faaa81fa2eff706).

SavoySchuler: Fixed [here](https://github.com/Microsoft/microsoft-ui-xaml-specs/commit/016cfd4cc7cb29b9935826614a598e5135d034a5).

active/teachingtip/TeachingTipSpec.md
SavoySchuler: @MikeHillberg  I specified that a TeachingTip can appear anywhere in the element tree or code behind [here](https://github.com/Microsoft/microsoft-ui-xaml-specs/commit/e3a4e0e756b69052210f911a1b3728f5a2a81626). Please let me know if I missed anything.

active/teachingtip/TeachingTipSpec.md
SavoySchuler: Added [this ](https://github.com/Microsoft/microsoft-ui-xaml-specs/commit/3ec3514c3e68902feb6b1d54a6c003817f7c78fe) warning under the Placement and Deferring Close examples. Also added P0 telemetry request.

active/teachingtip/TeachingTipSpec.md
SavoySchuler: [Fixed](https://github.com/Microsoft/microsoft-ui-xaml-specs/commit/f6f1c1d4d93efaeec1d4ac32526925eddc16fdcb)!

active/teachingtip/TeachingTipSpec.md
SavoySchuler: [Added a "Reconfiguring an open teaching tip" section to Remarks](https://github.com/Microsoft/microsoft-ui-xaml-specs/commit/b493afb1d70535508101aaeefaac1a922d3a7bcf).

active/teachingtip/TeachingTipSpec.md
SavoySchuler: Added the explanation [here](https://github.com/Microsoft/microsoft-ui-xaml-specs/commit/52887dcce86fc52198705b34d3b6e137a4688ee9).

active/teachingtip/TeachingTipSpec.md
SavoySchuler: Changed "XamlRoot" to "the xaml root" [here](https://github.com/Microsoft/microsoft-ui-xaml-specs/commit/2bf44703a703275f75a8ea53e1ca5ced6009ba95). I agree that a page special page on the topic may increase clarity. @MikeHillberg, anyone from DMC we can tag here?

active/teachingtip/TeachingTipSpec.md
SavoySchuler: Added requirement to inform Narrator what object the tip is pointing to [here](https://github.com/Microsoft/microsoft-ui-xaml-specs/commit/71881ccf08475b55e37409ec3f9e590e2e8c441d).

SavoySchuler:
I have updated TeachingTip's spec with all the feedback from this review. Thanks again for your feedback, @oldnewthing. 

I have left a few remarks open for further feedback from @MikeHillberg, @jevansaks, and @StephenLPeters.

PR #26, opened 2019-04-05
User/lucashaines/inputvalidation

LucasHaines:
adding the first draft of Input Validation for review


active/InputValidation/InputValidation.md
```diff
@@ -0,0 +1,572 @@
...
+that want to evaluate and give feedback on the API, as part of
+the submission process.  When it's complete
+it will be incorporated into the public documentation at
+docs.microsoft.com (http://docs.microsoft.com/uwp/toolkits/winui/).
+Hopefully we'll be able to copy it mostly verbatim.
+So the second audience is everyone that reads there to learn how
+and why to use this API. -->
+
+# Background
+In WPF Developers had access to Input Validation via [INotifyDataErrorInfo](https://docs.microsoft.com/en-us/dotnet/api/system.componentmodel.inotifydataerrorinfo?view=netframework-4.7.2) We are adding the same capability to UWP.
```
mrlacey: "We are adding the same capability to UWP" exactly the same or with some differences/modifications?
If identical, what's the need for this spec?
If not identical, what are the differences? If that's this doc, please be explicit about this.

active/InputValidation/InputValidation.md
```diff
@@ -0,0 +1,572 @@
...
+    [MaxLength(6, ErrorMessage = "Name must be no more than 6 characters")]
+    public string Name
+    {	
+        get { return GetValue<string>("Name"); }
+        set { SetValue("Name", value); }
+    }
+}
+```
+
+#### Code – INotifyDataErrorInfo base class
```
mrlacey: Shouldn't this header be `NotifyDataErrorInfo` not `INotifyDataErrorInfo`?

active/InputValidation/InputValidation.md
```diff
@@ -0,0 +1,572 @@
...
+The NotifyDataErrorInfoBase class (app code) provides a reusable implementation of INotifyDataErrorInfo (new API).  It’s the base of ValidationWrapper, which is the base class for Person.
```
mrlacey: It's not clear if this is a class that will be provided as part of the solution or if this is something that a developer would write themselves.

active/InputValidation/InputValidation.md
```diff
@@ -0,0 +1,572 @@
...
+            ? Errors[propertyName]
+            : new ObservableCollection<ValidationError>();
+    }
+
+    protected virtual void OnErrorsChanged(string propertyName)
+    {
+        ErrorsChanged?.Invoke(this, new DataErrorsChangedEventArgs(propertyName));
+    }
+
+    protected void ClearErrors()
```
mrlacey: This naming risks being confused with `Validation.ClearErrors()` which is used elsewhere in the samples in this document.

active/InputValidation/InputValidation.md
```diff
@@ -0,0 +1,572 @@
...
+    {
+        foreach (var propertyName in Errors.Keys.ToList())
+        {
+            Errors.Remove(propertyName);
+            OnErrorsChanged(propertyName);
+        }
+    }
+}
+```
+#### Code – ValidationWrapper (no new APIs)
```
mrlacey: Again, is this what will be provided? or is it an example of something a developer would write themselves? It seems generic enough that it could be used by anyone and not need recreating by each developer.

active/InputValidation/InputValidation.md
```diff
@@ -0,0 +1,572 @@
...
+ValidationWrapper derives from the above NotifyDataErrorInfoBase, and provides GetValue/SetValue methods for storage of subclass’ properties.  The SetValue performs validation.  This also provides an implementation of INotifyPropertyChanged.
```
mrlacey: Saying "This also provides an implementation of INotifyPropertyChanged." implies this implements this directly, rather than inheriting it from the base class. I suggest removing this sentence to avoid confusion.

active/InputValidation/InputValidation.md
```diff
@@ -0,0 +1,572 @@
...
+
+``` csharp
+public class ValidationWrapper : NotifyDataErrorInfoBase
+{
+
+    #region get/set
```
mrlacey: Are regions in this code sample needed? 
They add more lines to the code sample but don't add a lot of value.

active/InputValidation/InputValidation.md
```diff
@@ -0,0 +1,572 @@
...
+    protected T GetValue<T>(string propertyName)
```
mrlacey: Adding `[CallerMemberName]` here would be helpful.

active/InputValidation/InputValidation.md
```diff
@@ -0,0 +1,572 @@
...
+        }
+
+        object value;
+        value = default(T);
+
+        return (T)value;
+    }
+
+    // Validate() gets called from inside SetValue
+    protected void SetValue<T>(string propertyName, T value)
```
mrlacey: Adding `[CallerMemberName]` here would be helpful.

active/InputValidation/InputValidation.md
```diff
@@ -0,0 +1,572 @@
...
+        {
+            PropertyChanged?.Invoke(
+                    this, new PropertyChangedEventArgs(propertyName));
+            Validate(propertyName);
+        }
+    }
+    #endregion
+
+    #region Populate errors (C# - DataAnnotations)
+    private Task Validate(string propertyName)
```
mrlacey: The implementation of this method looks to have errors.
It clears all errors--which may come from multiple properties.
It then only adds errors for the current property.
If multiple properties had errors before validating a property, you'd end up with an error collection that doesn't contain all the errors.

active/InputValidation/InputValidation.md
```diff
@@ -0,0 +1,572 @@
...
+
+                OnErrorsChanged(propertyName);
+            }
+
+        }
+    }
+    #endregion
+
+    #region privates
+    private object GetPropertyValue([CallerMemberName] string propertyName = "")
```
mrlacey: Using `[CallerMemberName]` here is unnecessary as this si only called in one place and you'd never want the name of the caller passed.

active/InputValidation/InputValidation.md
```diff
@@ -0,0 +1,572 @@
...
+                </Border>
+                <TextBlock Text="{x:Bind ErrorMessage, Mode=OneWay}"
+                           />
+            </StackPanel>
+        </DataTemplate>
+    </ItemsControl.ItemTemplate>
+</ItemsControl>
+```
+
+## Customizing the error template
```
mrlacey: How does this work with different `InputValidationKind` options?

active/InputValidation/InputValidation.md
```diff
@@ -0,0 +1,572 @@
...
+    <Color x:Key="SystemErrorColor">#CE3880</Color>
+    <SolidColorBrush x:Key="SystemControlErrorTextForegroundBrush" Value="SystemErrorColor"/>
+
+    <!--Brush for description messages-->
+    <SolidColorBrush x:Key="SystemControlDescriptionTextForegroundBrush" Value="SystemControlPageTextBaseMediumBrush"/>
+
+</Application.Resources>
+```
+
+## Change description text for validation
```
mrlacey: This seems like something that warrants a spec in its own right as it seems bigger than is covered here.
Is this displayed only when there's an error?
How to display it based on custom logic?
How to restyle this?
How can this be incorporated in custom controls?

active/InputValidation/InputValidation.md
```diff
@@ -0,0 +1,572 @@
...
+
+Description text provides input to the end user on what format or content is expected from the control when it is not clear from the header and placeholder text.
```
mrlacey: This sounds useful even if there isn't an error. Why make it related to validation?

active/InputValidation/InputValidation.md
```diff
@@ -0,0 +1,572 @@
...
+
+Developers can choose to either show the description text for a control or not. Default is that description text is off.
```
mrlacey: How is this controlled?

active/InputValidation/InputValidation.md
```diff
@@ -0,0 +1,572 @@
...
+## Change description text for validation
+
+Description text provides input to the end user on what format or content is expected from the control when it is not clear from the header and placeholder text.
+
+Developers can choose to either show the description text for a control or not. Default is that description text is off. 
+
+Description text should be used to complement headers and placeholders. Guidance on what each corresponds to is as follows: 
+1.	Headers give you an indication of what the control represents. It is used to catch the eye of a user scanning your form or to refer to a field.
+2.	Placeholders provide a succinct description of what should be input in the box. However, they aren’t an ideal pattern for guiding users input because user testing continually shows that placeholders in form fields often hurt usability more than help it. This is because: 
+    - Disappearing placeholder text strains users’ short-term memory.
```
mrlacey: This is an argument for making it always visible. Can it be?

active/InputValidation/InputValidation.md
```diff
@@ -0,0 +1,572 @@
...
+    -	Users may mistake a placeholder for data that was automatically filled in.
+    -	The default light-grey color of placeholder text has poor color contrast against most backgrounds.
+    -	Not all screen readers read placeholder text aloud
```
mrlacey: How will the DescriptionText work with screen readers?

active/InputValidation/InputValidation.md
```diff
@@ -0,0 +1,572 @@
...
+            Header="First Name:" 
+            Text="{x:Bind   ViewModel.Person.FirstName, 
+                            UpdateSourceTrigger=LostFocus, 
+                            Mode=TwoWay}"
+            DescriptionText="Use alphabetical characters (a-Z)."
+        />
+```
+
+#### C# (Model)
+```csharp
```
mrlacey: Is this meant to work with the XAML directly above or an example of the C# way of doing the same thing?
That the `DescriptionText` in the XAML and the `ErrorMessage` in the C# use the same text is potentially confusing. Are developers expected to have this duplication?

active/InputValidation/InputValidation.md
```diff
@@ -0,0 +1,572 @@
...
+    public class Person : ValidationWrapper
+    {
+        [Required]
```
mrlacey: Will there be any UI affordance or indicator that works with required properties? 
If not, why not?
If so, how can this be controlled beyond using the attribute?

active/InputValidation/InputValidation.md
```diff
@@ -0,0 +1,572 @@
...
+Code a view-centric (Winforms-like) developer would write to enable Validation.
+
+
+### 1.10.1	C# Sample
+#### XAML
+
+```xaml
+<TextBox x:Name="Email" 
+         Header="Email" 
+         Validation.IsRequired="True"
```
mrlacey: Does this indicate that the field is a required field? or that validation should be performed?

If this means that entry in this textbox is required
- How is that displayed?
- Can it be re-templated?
- How should a developer support this in custom controls?

active/InputValidation/InputValidation.md
```diff
@@ -0,0 +1,572 @@
...
+}
+
+private void OnValidationError(object sender, ValidationErrorEventArgs e)
+{
+    if (e.Action == ValidationErrorEventAction.Added)
+        // Do something when an error message was added
+    else
+        // Do something when an error was removed.
+
+        SaveButton.IsEnabled = Validation.GetErrors(e.Source);
```
mrlacey: The indentation on this line risk confusing it as being part of the `else` clause.

active/InputValidation/InputValidation.md
```diff
@@ -0,0 +1,572 @@
...
+in IntelliSense. For properties, specify the default value of the property if it
+isn't the type's default (for example an int-typed property that doesn't default to zero.) -->
+
+<!-- Option 2: Put these descriptions in the below API Details section,
+with a "///" comment above the member or type. -->
+
+# API Details
+<!-- The exact API, in MIDL3 format (https://docs.microsoft.com/en-us/uwp/midl-3/) -->
+
+## InputPropertyAttribute
```
mrlacey: This title is very vague and non-descriptive of what it does.
There are lots of properties that capture input but don't generate any validation.
Why not use a name that communicates the intent better, such as `GeneratesValidationAttribute` or `InputValidationAttribute`?

active/InputValidation/InputValidation.md
```diff
@@ -0,0 +1,572 @@
...
+The `InputPropertyAttribute` helps us ensure that we only generate validation code for the property specified by this attribute. This will help reduce code bloat by not generating code for properties that will never perform validation (like width). Controls can specify multiple properties as their input property.
```
mrlacey: Why not apply this to the properties that generate validation, rather than at the class level?
This would remove the need to specify the name. It would also link the property to the fact that is has validation, by putting the attribute next to the property, as this can be easy to miss with large code files or controls that have code over multiple partial files.

active/InputValidation/InputValidation.md
```diff
@@ -0,0 +1,572 @@
...
+    /// <summary>
+    /// Name of the property that is associated with user input.
+    /// </summary>
+    attribute InputPropertyAttribute
+    {
+         string Name;
+    }
+
+## InputValidationKind  
+Determines how the control displays validation visuals.
```
mrlacey: How would a custom control implement this?

active/InputValidation/InputValidation.md
```diff
@@ -0,0 +1,572 @@
...
+        /// </summary>
+        Compact,
+        /// <summary>
+        /// Text for the error messages is displayed underneath the control.  
+        /// </summary>
+        Inline,
+        /// <summary>
+        /// Validation visuals are not displayed at all.
+        /// </summary>
+        Hidden,
```
mrlacey: Can you include an example of why this is relevant/useful?
I assume that it's for scenarios where the app will show errors for the view in a list together, rather than inline, but it would be good to include an explanation.

active/InputValidation/InputValidation.md
```diff
@@ -0,0 +1,572 @@
...
+    };
+
+## IInputValidationControl
+Any control that wants to participate in input validation should derive from this interface. This, among other things, is what allows the markup compiler to know when to generate the code necessary for listening to the `INotifyDataErrorInfo.ErrorsChanged` event and propagate the results to the `IInputValidationControl.Errors` property.
```
mrlacey: Should controls "derive" from this interface or _implement_ it?

active/InputValidation/InputValidation.md
```diff
@@ -0,0 +1,572 @@
...
+
+When the `ValidationCommand` property is non-null, the following will be true:
+1. When the `InputValidationKind` property is set to `Hidden`, the control will *not* invoke the `ValidationCommand.CanValidate` or `ValidationCommand.Validate` methods
```
mrlacey: How would this work where a developer doesn't want errors indicated inline but in a list elsewhere in the view?

active/InputValidation/InputValidation.md
```diff
@@ -0,0 +1,572 @@
...
+2. When the `InputValidationKind` property is set to `Auto`, the control will use the value of the `ValidationCommand.InputValidationKind` property for the value. If both are `Auto` the default bevahior should be `Inline`.
```
mrlacey: "When the `InputValidationKind` property is set to `Auto`" on what? the control?

active/InputValidation/InputValidation.md
```diff
@@ -0,0 +1,572 @@
...
+3. When focus has been lost, validation should be attempted by calling the `ValidationCommand.CanValidate` method.
```
mrlacey: What if `UpdateSourceTrigger` is set to `TextChanged`? - In this case I'd expect the validation to be triggered on TextChanged too. If a develoepr wanted that would they need to invoke validation themselves?
Or will validation always be called on LostFocus, regardless of if it's called elsewhere too?

active/InputValidation/InputValidation.md
```diff
@@ -0,0 +1,572 @@
...
+4. If the `IInputValidationControl.ValidationErrors` collection is non-empty, validation should be attempted by calling the `ValidationCommand.CanValidate` method.
+
+The markup compiler uses this interface to properly generate code when using x:Bind iff:
```
mrlacey: "iff" - typo?

active/InputValidation/InputValidation.md
```diff
@@ -0,0 +1,572 @@
...
+            Windows.UI.Xaml.Controls.InputValidationKind InputValidationKind;
+            /// <summary>
+            /// Command associated with the IInputValidationControl. See InputValidationCommand for more
+            /// information.
+            /// </summary>
+            Windows.UI.Xaml.Controls.InputValidationCommand ValidationCommand;
+         };
+   
+## InputValidationCommand
+The command allows for easy use of input validation without the need for a tight dependency on x:Bind.  This enables developers who aren't striclty using MVVM to use input validation (i.e ReactiveUI).  The InputValidationCommand is what allows for the correct-by-design approach where controls initially validate on focus lost, but then will query for validation on property changed once in an invalid state.  This wouldn't be for free, but the amount of code here is pretty trivial. We could provide some attribute or pattern that tells the markup compiler how to generate that code automatically.
```
mrlacey: "striclty" - typo?

active/InputValidation/InputValidation.md
```diff
@@ -0,0 +1,572 @@
...
+            /// information.
+            /// </summary>
+            Windows.UI.Xaml.Controls.InputValidationCommand ValidationCommand;
+         };
+   
+## InputValidationCommand
+The command allows for easy use of input validation without the need for a tight dependency on x:Bind.  This enables developers who aren't striclty using MVVM to use input validation (i.e ReactiveUI).  The InputValidationCommand is what allows for the correct-by-design approach where controls initially validate on focus lost, but then will query for validation on property changed once in an invalid state.  This wouldn't be for free, but the amount of code here is pretty trivial. We could provide some attribute or pattern that tells the markup compiler how to generate that code automatically.  
+
+The command can shared between different IInputValidationControls, as would most likely be the case in the scenario of a forms control.  A custom command can be made that listens for changes to all errors and knows when a form has been completed and ready for submition. The control will query the command as to whether they should perform validation.
```
mrlacey: The command can **be** shared between ...

active/InputValidation/InputValidation.md
```diff
@@ -0,0 +1,572 @@
...
+Developers can override any of this behavior by creating their own InputValidationCommand.
+ 
+ 
+    /// <summary>
+    /// Command that controls how an IInputValidationControl does validation.
+    /// </summary>
+    [webhosthidden]
+    unsealed runtimeclass InputValidationCommand
+    {
+        [method_name("CreateInstance")] InputValidationCommand();
```
mrlacey: This formatting looks weird. Should `[method_name("CreateInstance")]` be on its own line, as an attribute on the method?

active/InputValidation/InputValidation.md
```diff
@@ -0,0 +1,572 @@
...
+        Windows.UI.Xaml.Controls.InputValidationKind InputValidationKind;
+        /// <summary>
+        /// Called by a control to determine if it should call the Validate method.
+        /// </summary>
+        Boolean CanValidate(Windows.UI.Xaml.Controls.IInputValidationControl validationControl);
```
mrlacey: Can you include an explanation of why this is needed?
I can see lots of implementations that involve some of the logic necessary for validation. If that's not desirable or the intent this needs to be made clear.
Also, are there considerations that need to be made for the implementation here? such as, this needs to be really performant and so must return quickly as it may get called a lot.

active/InputValidation/InputValidation.md
```diff
@@ -0,0 +1,572 @@
...
+        /// </summary>
+        overridable Boolean CanValidateCore(Windows.UI.Xaml.Controls.IInputValidationControl validationControl);
+        /// <summary>
+        /// Method that derived classes implement to provide their own implementation of the Validate method.
+        /// </summary>
+        overridable void ValidateCore(Windows.UI.Xaml.Controls.IInputValidationControl validationControl);
+    };
+
+## ValidationContext
+The `ValidationContext` class provides context as to what is being required by validation. This class is auto-generated by the XAML markup compiler in certain scenarios (see [IInputValidationControl]()). Controls can use the provided properties on the `ValidationContext` to customize their appearance.  For example, the IsInputRequired field is can allow controls to put the little * next to a header.
```
mrlacey: For example, the IsInputRequired field ~is~ can allow controls to put the little * next to a header.

active/InputValidation/InputValidation.md
```diff
@@ -0,0 +1,572 @@
...
+Note: Intended to be extended for things like IsPhoneNumber or IsEmail that will allow controls to have other behavior (like mask text box).
+
+    
+    /// <summary>
+    /// Provides more contextual information about how to validate, display, and/or format input.
+    /// </summary>
+    [webhosthidden]
+    unsealed runtimeclass InputValidationContext
+    {
+        [method_name("CreateInstance")] InputValidationContext(String memberName, Boolean isRequired);
```
mrlacey: This formatting looks weird. Should `[method_name("CreateInstance")]` be on its own line, as an attribute on the method?

active/InputValidation/InputValidation.md
```diff
@@ -0,0 +1,572 @@
...
+        Boolean IsInputRequired{ get; };
+        String MemberName{ get; };
+    };
+    
+## InputValidationError
```
mrlacey: Can you explain why this is needed?

Why not just have `IInputValidationControl .ValidationErrors` return an `Windows.Foundation.Collections.IObservableVector<string>`?

active/InputValidation/InputValidation.md
GeertvanHorrik: Why not IDataErrorInfo, it's part of .net standard 2.0?

active/InputValidation/InputValidation.md
```diff
@@ -0,0 +1,572 @@
...
+
+Most applications that take user input need to have validation logic to ensure that the user has entered the expected information. The validation checks can be based on type, range, format, or other application-specific requirements. For example:
+![validation example](basicvalidationexample.png)
+
+
+# Examples
+
+## End to end input validation
+
+In this sample, we want to display a validation error on a textbox when user input is invalid. This is the recommended model for validating fields. The validation logic with DataAnnoations displayed in this sample is proper to managed apps. Native apps can use INotifyDataErrorInfo as described below and substitute their current validation logic in place of the .Net APIs.
```
GeertvanHorrik: Typo in DataAnnoations

active/InputValidation/InputValidation.md
```diff
@@ -0,0 +1,572 @@
...
+        Errors = new Dictionary<string, List<string>>();
+    }
+
+    public bool HasErrors => Errors.Any();
+
+    public event EventHandler<DataErrorsChangedEventArgs> ErrorsChanged;
+
+    public ObservableCollection<ValidationError> GetErrors(string propertyName)
+    {
+        return propertyName != null && Errors.ContainsKey(propertyName)
```
GeertvanHorrik: Since this serves as an example which many might just copy/paste, please consider using TryGetValue and a cached empty collection in case of no validation.

active/InputValidation/InputValidation.md
MikeHillberg: I agree, I'm not sure what this is doing in this spec.

active/InputValidation/InputValidation.md
```diff
@@ -0,0 +1,572 @@
...
+        [RegularExpression(@"^[a-zA-Z''-'\s]{1,40}$", ErrorMessage ="Use alphabetical characters (a-Z).")]
+        public string FirstName {
+            get { return GetValue<string>("FirstName"); }
+            set { SetValue("FirstName", value); }
+        }
+    }
+```
+
+## Validation for view-centric developers
+Code a view-centric (Winforms-like) developer would write to enable Validation.
```
MikeHillberg: This "WinForms" section is kind of out of the blue. So really there's three approaches: DataAnnotations, INDEI, and programmatic. Explaining this context would help.

active/InputValidation/InputValidation.md
MikeHillberg: Requiring an InputPropertyAttribute sounds difficult. Is there a way an x:Bind can specify it if the control author didn't realize it was required?

active/InputValidation/InputValidation.md
```diff
@@ -0,0 +1,572 @@
...
+<!-- Option 1: Give a one or two line description of each API (type
+and member), or at least the ones that aren't obvious
+from their name.  These descriptions are what show up
+in IntelliSense. For properties, specify the default value of the property if it
+isn't the type's default (for example an int-typed property that doesn't default to zero.) -->
+
+<!-- Option 2: Put these descriptions in the below API Details section,
+with a "///" comment above the member or type. -->
+
+# API Details
```
MikeHillberg: Almost none of this API is demonstrated or explained in the Examples

active/InputValidation/InputValidation.md
MikeHillberg: Or if means "if and only if", please write out the whole thing

active/InputValidation/InputValidation.md
stevenbrix: Some of the controls have a `Description` property:
https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.controls.textbox.description

active/InputValidation/InputValidation.md
stevenbrix: Great point! I think it's worth mentioning as part of the overall experience for building a form, but maybe it isn't worth calling out as part of validation

active/InputValidation/InputValidation.md
stevenbrix: One potential way we could do this is add a `Validation` value to the `BindingMode` enum:
`Text="{x:Bind Person.Name, Mode=Validation}"`

Or, we loosen the restriction. It might be unlikely that we'll be generating dead code, but it's hard to say with any certainty that we won't.

active/InputValidation/InputValidation.md
stevenbrix: The idea here was that the Description text will be replaced with the error text, but this does assume that `InputValidationKind` is set to `Inline`. When using `Compact` i'm pretty sure the `Desription` will still show

active/InputValidation/InputValidation.md
stevenbrix: Yes, the controls were intended to know they are required and put a '*' on the header.

It can be set using the `ValidationContext.IsInputRequired` field. This get's auto-generated and set by the markup compiler when using `System.ComponentModel.DataAnnotations`

active/InputValidation/InputValidation.md
stevenbrix: > How is that displayed?

By an '*' next to the header

> Can it be re-templated?

Yes, this will be part of the ControlTemplate

> How should a developer support this in custom controls?

By using the built-in controls as an example and following what they do

active/InputValidation/InputValidation.md
stevenbrix: That's a great point, thank you! The vagueness was a bit intentional, as I know we've discussed other uses for having this attribute in the past (I don't recall off the top of my head what those were). This attribute was not intended for validation only scenarios, but if we'd prefer it to be more scoped, that is a reasonable thing to do.

active/InputValidation/InputValidation.md
stevenbrix: It was meant to follow the same pattern that we did for `ContentPropertyAttribute`. I personally don't have much of a preference, and think you make some good points :)

active/InputValidation/InputValidation.md
stevenbrix: Great questions! So originally, we had to make `Compact` the default, since having them set to `Inline` would have been a breaking change.  Specifying `Inline` means that an extra padding is added below the control so that when the errors are shown, there is no affect to layout (assuming only a single line of text).

active/InputValidation/InputValidation.md
stevenbrix: Thanks for calling this out! We originally had a seperate enum called `InputValidationMode` that could be set to `Disabled`. Then `Hidden` would just mean "don't display" and we could display errors elsewhere like you suggest. I was trying to simplify the API as much as possible while retaining functionality, and this looks like something that was missed.

active/InputValidation/InputValidation.md
stevenbrix: yeah the control

active/InputValidation/InputValidation.md
stevenbrix: So my goal was to try and decouple the features that come from using `x:Bind` and how validation is designed to work. In my opinion, it is better for us to design input validation independently, so that developers who aren't using the MVVM pattern can use input validation nicely. I still need to validate this API would meet those goals for an app using a framework like ReactiveUI.

As part of the original design, there was an `InputValidationMode` enum that would allow more fine grained control over this. However, my thoughts are that the correct end-user experience for validation is that we should only do initial validation on FocusLost (once user has signaled they are done typing). Then, once the control is invalidated, we try to re-validate on PropertyChanged, so that the user knows as soon as they have fixed the issue. I couldn't really think of a scenario where you wouldn't want this to be the experience for the end-user, so i ommitted the enum from the spec.

The goal here was to describe how validation should work, regardless if `x:Bind` is being used or not.  The only thing that using `x:Bind` really gives you is the boilerplate code to listen to `INotifyDataErrorInfo.ErrorsChanged` and pass those errors to the control. Setting `UpdateSourceTrigger` only controls when the target tries to update the source, and in turn that means your model will raise the `ErrorsChanged` event (assuming invalid input)

active/InputValidation/InputValidation.md
stevenbrix: That's a great question, and i was wondering that myself a lot. A few reasons i kept it around:
1. Closer to what WPF has (maybe not a great reason?)
2. We might want to change the concept of an "error" to a "result" that can be either positive or negative, allowing controls to give visual feedback when a field is correctly filled out. This becomes a bit more complicated if we are relying on `INotifyDataErrorInfo` since it well, only notifies about errors :)
3. Easier to extend for the future if it turns out we want to add more functionality. This was a bigger deal when we were designing this feature as part of Windows, since we can't really change APIs once they ship. We probably can have more flexibility here and could just make this a string.

active/InputValidation/InputValidation.md
thomasclaudiushuber: @stevenbrix @MikeHillberg 

I associate the mode only with the direction of the binding, so a mode `Validation` looks a bit strange to me. Can't we have a `ValidatesOnNotifyDataErrors` property like on Binding in WPF:

```{x:Bind Person.Name, Mode=TwoWay,ValidatesOnNotifyDataErrors=True}```

Unlike in WPF, it could have the default value false. But: We could have the InputPropertyAttribute in addition. If you create a binding on such a property that has the InputPropertyAttribute, you have to explicitly set ValidatesOnNotifyDataErrors to false if you don't want to generate the binding code. By default, ValidatesOnNotifyDataErrors is true on an InputProperty binding

On the other side, if you have a property without the InputPropertyAttribute - the case that @MikeHillberg  mentioned - you could set ValidatesOnNotifyDataErrors to true to generate the validation code for that binding.

File: active/InputValidation/InputValidation.md (hunk 37 above)
Comment (mdtauk) [reply active/InputValidation/InputValidation.md]: There should probably be system brush resources added to the ThemeResource dictionaries for positive validation, as well as the negative error red and yellow, currently afforded

File: active/InputValidation/InputValidation.md (hunk 1 above)
Comment (LucasHaines) [reply active/InputValidation/InputValidation.md]: INotifyDataErrorInfo is a super set. It's a v2 of IDataErrorInfo, we fell that this covers all of the same scenarios and is async by nature. Are there any scenarios we are not thinking of that are supported in IDataErrorInfo?

File: active/InputValidation/InputValidation.md [hunk 42]
```diff
@@ -0,0 +1,572 @@
...
+        get { return GetValue<string>("Name"); }
+        set { SetValue("Name", value); }
+    }
+}
+```
+
+#### Code – INotifyDataErrorInfo base class
+The NotifyDataErrorInfoBase class (app code) provides a reusable implementation of INotifyDataErrorInfo (new API).  It’s the base of ValidationWrapper, which is the base class for Person.
+``` csharp
+public class NotifyDataErrorInfoBase : INotifyDataErrorInfo
```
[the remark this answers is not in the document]
LucasHaines: Thanks for feedback. @stevenbrix Is going to write some more coherent examples for this.

active/InputValidation/InputValidation.md
LucasHaines: This will be removed in the latest incarnation. With the new model we are updating to you can clear the collection of errors. Will update the samples to help call this out clearer.

active/InputValidation/InputValidation.md
LucasHaines: Thanks for the feedback. Yes this is boilerplate code that the developer would have to add. What if we added this as part of the community toolkit to make it easier?

active/InputValidation/InputValidation.md
LucasHaines: Thanks for the feedback. Yes this is boilerplate code that the developer would have to add. What if we added this as part of the community toolkit to make it easier?

active/InputValidation/InputValidation.md
LucasHaines: We are going to update the examples to clearly call out the work required from a developer. We will add in standard conventions to replace my bad PM code ;)

active/InputValidation/InputValidation.md
LucasHaines: No, they are not. We will remove in a better sample.

active/InputValidation/InputValidation.md
LucasHaines: Agree.

active/InputValidation/InputValidation.md
LucasHaines: Agree. Will add in update coming soon.

active/InputValidation/InputValidation.md
LucasHaines: The input validation kind determine if the template is displayed inline or in tooltip.

active/InputValidation/InputValidation.md
```diff
@@ -0,0 +1,572 @@
...
+
+```xaml
+<Application.Resources>
+
+    <!--Color for error messages-->
+    <Color x:Key="SystemErrorColor">#CE3880</Color>
+    <SolidColorBrush x:Key="SystemControlErrorTextForegroundBrush" Value="SystemErrorColor"/>
+
+    <!--Brush for description messages-->
+    <SolidColorBrush x:Key="SystemControlDescriptionTextForegroundBrush" Value="SystemControlPageTextBaseMediumBrush"/>
```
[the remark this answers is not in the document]
LucasHaines: Good call. Will update the example to set SystemErrorColor.

active/InputValidation/InputValidation.md
LucasHaines: We will update this section to call out to use description with input validation and remove the overall description information.

active/InputValidation/InputValidation.md
LucasHaines: Are you suggesting that we make description always visible?

active/InputValidation/InputValidation.md
LucasHaines: We are going to update this section, to speak about how description text works with Input Validation, and remove the general Description text guidance. Taking a look at our documentation, we don't have this called out any where. I will follow up, description was added in 1809, I need to make sure we have this scenario covered.

active/InputValidation/InputValidation.md
LucasHaines: We should call this out in the samples.

active/InputValidation/InputValidation.md
LucasHaines: Good feedback. Will expand.

active/InputValidation/InputValidation.md
LucasHaines: We are updating to match the latest IDL provided and clean up the samples.

active/InputValidation/InputValidation.md
LucasHaines: We will add example that illustrates how a custom control can implement this.

active/InputValidation/InputValidation.md
LucasHaines: Thanks for the feedback, How about this... _"There are differences from the previous implementations, we are building off the of lessons learned. Which will results in differences in the new model."_

GeertvanHorrik: True, but if I understand correctly, IDataErrorInfo is part of .NET Standard 2.0, and it would be extremely confusing not to support it.

I think you can easily do something like this so it will never cost additional performance:

```
// Async first
var notifyDataErrorInfo = obj as INotifyDataErrorInfo;
if (!(notifyDataErrorInfo is null))
{
    // async validation here
    return;
}

var dataErrorInfo = obj as IDataErrorInfo;
if (!(dataErrorInfo is null))
{
    // sync validation here
    return;
}

// Validation not supported
```

active/InputValidation/InputValidation.md
thomasclaudiushuber: I think that's actually a great idea to add a basic implementation of the interface as part of the community toolkit.

active/InputValidation/InputValidation.md
thomasclaudiushuber: I think this should be "Sample INotifyDataErrorInfo implementation", as this base class won't be part of WinUI (See next comment from @LucasHaines)

active/InputValidation/InputValidation.md
mrlacey: Toolkit seems like a good place for this

active/InputValidation/InputValidation.md
mrlacey: So I'm clear that I understand correctly, you're saying that if this is specified it becomes the template that's used when in the error state, regardless of the `InputValidationKind`?

What then about needing to allow for coping with variations in the UI for different focus states when also in the error state?

mdtauk: Is there a need for a control primitive which can be styled for this Validation message?  It could also be used on other controls that are not TextBox?

active/InputValidation/InputValidation.md
mrlacey: Clarification needed: Is this adding a new property/ sub-element called `Description` everywhere or is this related to the property that already exists on some controls.

active/InputValidation/InputValidation.md
mrlacey: Some descriptions already are always visible.
As this is included in this spec I wondered how it was different from the description property that also exists elsewhere. 

Consider `PasswordBox.Description` that already exists. I might use this to include a password hint on the form, e.g. "Remember, this is different to your system login." If validation of this filed/control was failed, would an additional piece of descriptive text about the error be displayed? Would the description I provided previously be replaced? Something else?

active/InputValidation/InputValidation.md
mrlacey: This answers some of the additional comments I've just added above. Thanks for the clarification.

active/InputValidation/InputValidation.md
mrlacey: I'm not sure that saying developers wishing to build their own custom controls should copy the built-in ones is adequate. 
Lots of developers will want to use C#/XAML, rather than the C++ the built-in ones are created with.
Hopefully toolkit controls will be updated accordingly and so they can be used as a reference by 3rd parties but may this also serve as a reminder that samples/examples often need to be in more than one language and copying C++ to create C# equivalents can be daunting for many.

active/InputValidation/InputValidation.md
mrlacey: As it wasn't mentioned explicitly I just wanted to make sure it had been considered and wasn't likely to cause any unforeseen problems if attempted. Thanks for clarifying.

active/InputValidation/InputValidation.md
mrlacey: The prospect of adding `Validation` as an option for  binding mode concerns me as it potentially causes the mode to have two meanings and would require changes in (especially 3rd party) tooling to handle the change in meaning.

active/InputValidation/InputValidation.md
stevenbrix: > Hopefully toolkit controls will be updated accordingly and so they can be used as a reference by 3rd parties but may this also serve as a reminder that samples/examples often need to be in more than one language and copying C++ to create C# equivalents can be daunting for many.

Fair point, thank you for bringing that up. I am working on a full end-to-end example for C# and will make sure it's made available

active/InputValidation/InputValidation.md
LucasHaines: Please check the latest spec and samples.

active/InputValidation/InputValidation.md
LucasHaines: Better samples updated.

active/InputValidation/InputValidation.md
```diff
@@ -0,0 +1,592 @@
...
+```xaml
+<TextBox 
+        x:Name="Name" 
+        Header="Name: " 
+        PlaceholderText="Enter Name..." 
+
+        Text="{x:Bind   ViewModel.Person.Name, 
+                        UpdateSourceTrigger=PropertyChanged, 
+                        Mode=TwoWay}" 
+         InputValidationKind="Hidden"
```
jevansaks: Will all controls that support input validation need this property added? If so it should be in the API section.

active/InputValidation/InputValidation.md
```diff
@@ -0,0 +1,592 @@
...
+            /// </summary>
+            Windows.UI.Xaml.Controls.InputValidationMode InputValidationMode;
+            /// <summary>
+            /// Command associated with the IInputValidationControl. See InputValidationCommand for more
+            /// information.
+            /// </summary>
+            Windows.UI.Xaml.Controls.InputValidationCommand ValidationCommand;
+         };
+   
+## InputValidationCommand
```
jevansaks: Did WPF have an equivalent of this? I know we got some pushback on having the framework define Commanding types in the framework layer -- I think folks were expecting these types to come out of the .NET layer. Now maybe .NET doesn't have any equivalent types that can stand in here, but I'd like to understand the need that this is satisfying. Or maybe just the "Command" part of the type name is throwing me off.

active/InputValidation/InputValidation.md
```diff
@@ -0,0 +1,592 @@
...
+            Windows.Foundation.Collections.IObservableVector<Windows.UI.Xaml.Controls.InputValidationError> ValidationErrors{ get; };
+            /// <summary>
+            ///  Extra information about the Source of the validation. 
+            ///  Note: This is automatically set by the XAML markup compiler. See summary above. 
+            /// </summary>
+            Windows.UI.Xaml.Controls.InputValidationContext ValidationContext;
+            /// <summary>
+            ///  DataTemplate that expresses how the errors are displayed.
+            /// </summary>
+            Windows.UI.Xaml.DataTemplate ErrorTemplate;
```
jevansaks: Why does ErrorTemplate need to be on this interface?

active/InputValidation/InputValidation.md
```diff
@@ -0,0 +1,592 @@
...
+         /// <summary>
+         /// Interface for Controls that participate in input validation.
+         /// <summary/>
+          [webhosthidden]
+         interface IInputValidationControl 
+         {
+             /// <summary>
+             /// Collection of errors to display based on the InputValidationKind property.
+            /// </summary>
+            Windows.Foundation.Collections.IObservableVector<Windows.UI.Xaml.Controls.InputValidationError> ValidationErrors{ get; };
```
jevansaks: Could you update this comment to describe where the errors come from? Does the control itself populate them? Does the XAML compiler generated code set them?

active/InputValidation/InputValidation.md
jevansaks: I don't think this attribute should be needed.

knightmeister:
I apologise, but I don't know how to reference the markdown file directly as others have done.

With respect to the [Custom Validation UI](https://github.com/microsoft/microsoft-ui-xaml-specs/blob/user/lucashaines/inputvalidation/active/InputValidation/InputValidation.md#custom-validation-ui), I'd love to see an attached property for UIElement controls added. I feel this would be less verbose than the sample linked above and also clearer to grok while reading the XAML. I do see value in supporting both methodologies.

What would be nice to see is something like
````
<TextBox ....>
    <TextBox.Validation DisplayValidationMessages="All|First|Last" Position="Right|Below">
        <DataTemplate x:DataType="ValidationError">
            <StackPanel Orientation="Horizontal" Spacing="5">
                <Border Background="Red" Margin="3" CornerRadius="10" 
                        Width="12" Height="12" VerticalAlignment="Top">
                    <TextBlock Text="!" Foreground="White" 
                               HorizontalAlignment="Center" VerticalAlignment="Center"/>
                </Border>
                <TextBlock Text="{x:Bind ErrorMessage, Mode=OneWay}" />
            </StackPanel>
        </DataTemplate>
    </TextBox.Validation>
</TextBox>
`````
Being able to build a validation message as I've proposed above would allow the validation UI for each form element to be defined in the XAML tree for the relevant control.

I don't know whether attached properties allow attributes, but I feel that if possible those two attributes would be quite useful.

**Comment by mdtauk** [comment]
I think there needs to be some thought about the ErrorTextColor in both light and dark themes.

Looking at the iconography used by Windows defender, and across the OS - as well as the colours for the proposed Status Banner #913...

The Dark Theme error colour is a yellow colour, and light theme uses red.

Should the red, green, and yellow colours, used in these icons, not be added as theme resources, and used for error and confirmation validation?  They would then be consistent with Light and Dark themes?

**Comment by kronic** [comment]
Why only Mode=TwoWay. I would like to see errors for all modes.

**Comment by thomasclaudiushuber** [comment]
> Why only Mode=TwoWay. I would like to see errors for all modes.

Hmmm, I'm not sure if I understand the reason why you want it for all modes.

Normally you show errors to tell the user what they have to enter and in which form they have to enter the data. To allow a user to enter data, you have to use a TwoWay Data Binding -> assuming you're using Data Bindings to connect the UI to your Data/ViewModel.

What is the scenario to show errors in a User Interface where a user can't change anything, because there are for example only OneWay Data Bindings? This would be a readonly User Interface.

**Comment by kronic** [comment]
> 
> 
> > Why only Mode=TwoWay. I would like to see errors for all modes.
> 
> Hmmm, I'm not sure if I understand the reason why you want it for all modes.
> 
> Normally you show errors to tell the user what they have to enter and in which form they have to enter the data. To allow a user to enter data, you have to use a TwoWay Data Binding -> assuming you're using Data Bindings to connect the UI to your Data/ViewModel.
> 
> What is the scenario to show errors in a User Interface where a user can't change anything, because there are for example only OneWay Data Bindings? This would be a readonly User Interface.

In the current project, I have errors in the datagrid in oneway mode

File: active/InputValidation/InputValidation.md (hunk 45 above)
Comment (stevenbrix) [reply active/InputValidation/InputValidation.md]: no they don't, we're going to remove this type and come up with a different, and simpler model for achieving what it's purpose was

File: active/InputValidation/InputValidation.md (hunk 47 above)
Comment (stevenbrix) [reply active/InputValidation/InputValidation.md]: they are populated by the app - either via the compiler if using x:Bind, or the app author

**Comment by stevenbrix** [comment]
> > > Why only Mode=TwoWay. I would like to see errors for all modes.
> > 
> > 
> > Hmmm, I'm not sure if I understand the reason why you want it for all modes.
> > Normally you show errors to tell the user what they have to enter and in which form they have to enter the data. To allow a user to enter data, you have to use a TwoWay Data Binding -> assuming you're using Data Bindings to connect the UI to your Data/ViewModel.
> > What is the scenario to show errors in a User Interface where a user can't change anything, because there are for example only OneWay Data Bindings? This would be a readonly User Interface.
> 
> In the current project, I have errors in the datagrid in oneway mode

I'm with @thomasclaudiushuber on this, but would like to understand your use cases more. How are you propagating data back to your model and performing validation?

**Comment by stevenbrix** [comment]
> I think there needs to be some thought about the ErrorTextColor in both light and dark themes.
> 
> Looking at the iconography used by Windows defender, and across the OS - as well as the colours for the proposed Status Banner #913...
> 
> The Dark Theme error colour is a yellow colour, and light theme uses red.
> 
> Should the red, green, and yellow colours, used in these icons, not be added as theme resources, and used for error and confirmation validation? They would then be consistent with Light and Dark themes?

@mdtauk if it's not in the spec then that was unintentionally left out, there will be theme resources made available for the error color (not sure we have a confirmation color?)

**Comment by kronic** [comment]
@stevenbrix 
A model with data can be checked not only by the user's water, but for example by a button. for example, we drive field a, then we drive field b, and we check field c for read-only, which is calculated on the basis of a and b.

File: active/InputValidation/InputValidation.md [hunk 48]
```diff
@@ -0,0 +1,592 @@
...
+There are a set of requirements that enable input validation to occur:
+
+1. The target property of the x:Bind matches the value of the `InputPropertyAttribute` of the target
+2. The x:Bind has a `Mode` of `TwoWay`
+3. The target of the bind must implement `IInputValidationControl`
+4. The source of the x:Bind implements `INotifyPropertyChanged` as well as `INotifyDataErrorInfo`
+
+### [App Code] Proper Xaml usage
+
+Lets first start with some simple XAML that illustrates how this is done from markup. This example satisfies requirements 1-3 described above, since the `TextBox` implements the proper interfaces and specifies that the `Text` property is the input property, and we are using the proper `BindingMode`. It's important to note that the default `UpdateSourceTrigger` for TextBox is `FocusLost`, which is desirable. This ensures that the initial validation occurs after the user has finished entering input before trying to validate.
```
MikeHillberg, line 59: What determines that the UpdateSourceTrigger for TextBox is FocusLost?

active/InputValidation/InputValidation.md
Noemata: Having worked on several complex WPF and Silverlight forms applications, INotifyDataErrorInfo was sufficient for validation purposes.  There was no need for IDataErrorInfo given it was a subset.  There a lot of discussion here around this without making reference to Silverlight.  Silverlight had/has the most evolved validation support given how it nicely tied into UI controls.  Just replicating Silverlight's validation features would be a big step forward for UWP/WPF/Xamarin.

Noemata:
Before committing to a final solution, please review the validation features that were present in Silverlight.  These worked very well in complex forms applications.  Also, I always wondered why Adorners were not added to Silverlight and now UWP.  An Adorner allows you to add a layer of functionality over an existing control, so could also be used for validation.

PR #27, opened 2019-04-26
RadialGradientBrush

SavoySchuler:
Spec review for RadialGradientBrush


active/RadialGradientBrush/RadialGradientBrush.md
```diff
@@ -0,0 +1,167 @@
...
+
+
+# Description
+<!-- Use this section to provide a brief description of the feature.
+For an example, see the introduction to the PasswordBox control 
+(http://docs.microsoft.com/windows/uwp/design/controls-and-patterns/password-box). -->
+
+Paints an area with a radial gradient. A focal point defines the beginning of the gradient, and an ellipse defines the end point of the gradient. Stops can be used to mark the points at which the gradient changes from one color to the next. 
+
+RadialGradientBrush can be animated by changing the location and color of stops over time.
```
jevansaks: Can you provide an example of animating the stops?

active/RadialGradientBrush/RadialGradientBrush.md
```diff
@@ -0,0 +1,167 @@
...
+isn't the type's default (for example an int-typed property that doesn't default to zero.) -->
+
+<!-- Option 2: Put these descriptions in the below API Details section,
+with a "///" comment above the member or type. -->
+
+# API Details
+<!-- The exact API, in MIDL3 format (https://docs.microsoft.com/en-us/uwp/midl-3/) -->
+
+```C++
+runtimeclass RadialGradientBrush : Windows.UI.Xaml.Media.GradientBrush
```
jevansaks: What's the advantage of deriving from GradientBrush? The reason I ask is that I think we'll actually have to derive from XamlCompositionBrushBase since I think we'd be talking about building a composition effect to do this. So we should just include all the API from that control on here as well.

active/RadialGradientBrush/RadialGradientBrush.md
```diff
@@ -0,0 +1,167 @@
...
+{
+    RadialGradientBrush();
+
+    Windows.Foundation.Numerics.Vector2 EllipseCenter; 
+    Windows.Foundation.Numerics.Vector2 EllipseRadius; 
+    Windows.Foundation.Numerics.Vector2 GradientOrigin; 
+
+    Windows.UI.Xaml.Media.GradientStopCollection GradientStopCollection
```
jevansaks:
```suggestion
    Windows.UI.Xaml.Media.GradientStopCollection GradientStopCollection;
```

active/RadialGradientBrush/RadialGradientBrush.md
```diff
@@ -0,0 +1,167 @@
...
+        <GradientStop Color="White" Offset="0.875" />
+        <GradientStop Color="Black" Offset="1.0" />
+      </RadialGradientBrush>
+    </Rectangle.Fill>
+  </Rectangle>
+</StackPanel>
+```
+
+# Remarks
+<!-- Explanation and guidance that doesn't fit into the Examples section. -->
```
jevansaks: Is there an animated stops example?

active/RadialGradientBrush/RadialGradientBrush.md
jesbis: Added a section on animation support. I don't think we can readily support independent animation of stops on pre-RS5 so it seems worth deferring to WinUI 3+

PR #1, opened 2018-11-29
API spec template

MikeHillberg:

PR #28, opened 2019-05-28
TabControl Spec

stmoy:
Spec review for TabControl.


active/tab/tabcontrol.md
```diff
@@ -0,0 +1,245 @@
...
+<!-- Give a one or two line description of each API (type
+and member), or at least the ones that aren't obvious
+from their name.  These descriptions are what show up
+in IntelliSense. -->
+
+### TabControl properties and events
+
+| Property | Type | Description |
+|:-------- |:---- |:----------- |
+| CanCloseTabs | bool | Default value for the item if it doesn't specify a IsClosable value. |
```
YuliKl: It took me several attempts to understand the meaning of this description. Now that I get it, perhaps AreItemsClosable would be a better name? But really, I continue to question the need for this property. What scenario requires that all tab items not be closable?

active/tab/tabcontrol.md
```diff
@@ -0,0 +1,245 @@
...
+| CloseButtonOverlay | enum | Describes the behavior of the close button. Values are {Auto, OnHover, Persistent} |
```
YuliKl: What's the Auto behavior? And is this property required? (Again, primarily asking because of my earlier comment about touch devices.)

active/tab/tabcontrol.md
```diff
@@ -0,0 +1,245 @@
...
+| ItemHeaderTemplate | DataTemplate | Default template for the item if no template specified. |
+| SelectedTabWidth | double | The size of the selected tab header. |
+| TabHeader | object | Content to the left of the tab strip. |
```
YuliKl: Maybe it's just me, but the TabHeader and ItemHeader(Template) concepts sound very similar and potentially confusing. I know you're proposing names that are independent of reading order. But as I've learned from NavigationView, header gets overloaded and overused very quickly.

active/tab/tabcontrol.md
```diff
@@ -0,0 +1,245 @@
...
+## To replicate the behavior of Microsoft Edge
+
+``` xml
+<TabControl TabWidthBehavior="Equal"
+            CanCloseTabs="True"
+            CloseButtonOverlay="OnHover"
+            CanDragItems="True"
+            CanReorderItems="True"
+            TabDraggedOutside="OpenTabInNewWindow">
+    <TabControl.TabFooter>
```
YuliKl: If this is replicating Edge, should the + button be in the TabActionContent instead of TabFooter?

active/tab/tabcontrol.md
```diff
@@ -0,0 +1,245 @@
...
+| CloseButtonOverlay | enum | Describes the behavior of the close button. Values are {Auto, OnHover, Persistent} |
+| ItemHeaderTemplate | DataTemplate | Default template for the item if no template specified. |
+| SelectedTabWidth | double | The size of the selected tab header. |
+| TabHeader | object | Content to the left of the tab strip. |
+| TabHeaderTemplate | DataTemplate | Template for the Header. |
+| TabFooter | object | Content to the far right of the tab strip. |
+| TabFooterTemplate | DataTemplate | Template for the Footer. |
+| TabActionContent | object | Content immediately to the right of the tabs |
+| TabActionContentTemplate | DataTemplate | Template for the ActionContent. |
+| TabWidthBehavior | enum | Specifies how the tabs should be sized. Values are {Actual, Equal, Compact} |
```
YuliKl: Would be useful to list the default value for TabWidthBehavior.

active/tab/tabcontrol.md
```diff
@@ -0,0 +1,245 @@
...
+| TabClosing | Fires when a tab is about to be closed. Can be cancelled to prevent closure. |
+| TabDraggedOutside | Fires when a Tab is dragged outside of the Tab bar. |
+
+### TabItem properties and events
+
+| Property | Type | Description |
+|:-------- |:---- |:----------- |
+| Content | object | The main content that appears in the tab. |
+| Header | object | The content that appears inside the tab itself.  |
+| HeaderTemplate | DataTemplate | Template for the header object. |
```
YuliKl: How is this different from ItemHeaderTemplate? (Other than being a member of a different class.) When would I use HeaderTemplate and when would I use ItemHeaderTemplate?

active/tab/tabcontrol.md
```diff
@@ -0,0 +1,245 @@
...
+| Icon | IconElement | Icon for the tab. |
+| IsClosable | bool | Determines if the tab shows a close button. |
```
YuliKl: Will this property end up with strange use cases of arbitrary tab items not showing a close button while others do? I'm guessing the use case here is to dynamically prevent the last open tab from being closable. But how important is this scenario?

active/tab/tabcontrol.md
```diff
@@ -0,0 +1,245 @@
...
+}
+```
+
+## Databind to a set of tabs 
+
+``` xml
+<TabControl ItemsSource="{x:Bind TabItemCollection}" />
+```
+
+# Remarks
```
YuliKl: I would be interested in learning more about TabControl's resizing behavior. The API includes TabWidthBehavior, which is part of this story. I'm also curious about what happens to the space allocated for TabHeader, TabFooter, and TabActionContent as the tab content grows or window width shrinks.

What if I put a CommandBar with dynamic overflow into the TabFooter, how will the elements behave? What if I put a stack panel with multiple items in the TabFooter? Do I need to wrap it in a ScrollViewer to ensure all the items remain accessible?

active/tab/tabcontrol.md
```diff
@@ -0,0 +1,245 @@
...
+
+``` xml
+<TabControl TabWidthBehavior="Equal"
+            CanCloseTabs="True"
+            CloseButtonOverlay="OnHover"
+            CanDragItems="True"
+            CanReorderItems="True"
+            TabDraggedOutside="OpenTabInNewWindow">
+    <TabControl.TabFooter>
+        <Button Content="+" Click="NewTab_Click" />
```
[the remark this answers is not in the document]
stmoy: Agreed. Since we're not providing the button as part of the control, it has been suggested that we provide a "NewTabButtonStyle" or something. Will update the sample.

active/tab/tabcontrol.md
stmoy: This supports the "pinned tab" scenario, like in Edge. The Pinned tab cannot be closed via a close button.

active/tab/tabcontrol.md
```diff
@@ -0,0 +1,245 @@
...
+```
+
+# Remarks
+<!-- Explanation and guidance that doesn't fit into the Examples
+section.  For example, see the Remarks for the MediaPlayerElement 
+(https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.Controls.MediaPlayerElement#remarks). -->
+
+## Keyboarding behavior
+### TAB and Arrow key behavior
+The TabControl has areas containing left content, right content, and the Tab content (containing 0 or more Tabs). When the user hits the TAB key, focus moves between the content areas. When focus moves into the Tab content area, the selected Tab gains focus. Once the user is in the Tab content area, pressing the arrow key will move focus between the tabs. Arrow focus is trapped inside the Tab content area.
```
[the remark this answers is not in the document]
stmoy: I will follow up with @chigy about the keyboarding behavior. The behavior as spec'd matches our overall keyboarding guidance. 

Since the + button is developer provided, if we decide that the arrowing behavior suggested in Edge (dev) is the way to go, we will want to provide a sample about how to make this work.

active/tab/tabcontrol.md
```diff
@@ -0,0 +1,245 @@
...
+
+<!-- As an example of this section, see the Examples section for the PasswordBox control 
+(https://docs.microsoft.com/windows/uwp/design/controls-and-patterns/password-box#examples). -->
+
+## To replicate the behavior of Microsoft Edge
+
+``` xml
+<TabControl TabWidthBehavior="Equal"
+            CanCloseTabs="True"
+            CloseButtonOverlay="OnHover"
```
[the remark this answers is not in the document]
stmoy: The selected tab always shows the X-to-close button (assuming it is closable). CloseButtonOverlay only affects the unselected tabs. For a touch-only device, that means that the tab must be selected before it is closed.

active/tab/tabcontrol.md
YuliKl:
```suggestion
| CloseButtonOverlay | enum | Describes the behavior of the close button in unselected tabs. Values are {Auto, OnHover, Persistent} |
```

active/tab/tabcontrol.md
```diff
@@ -0,0 +1,245 @@
...
+            CanDragItems="True"
+            CanReorderItems="True"
```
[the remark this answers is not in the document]
stmoy: Agreed, this inconsistency is a bit strange. In general I prefer the word "Tabs", but unfortunately CanDragItems and CanReorderItems are funneled through from ListViewBase. Since this feedback is specifically related to naming and not functionality, I think I'm going to hold off making any changes until the API folks take a look. It's possible that this inconsistency isn't actually that strange.

active/tab/tabcontrol.md
stmoy: TabItem.HeaderTemplate takes precedence over TabControl.ItemHeaderTemplate. You'd use TabItem.HeaderTemplate if you wanted to specify the template of a single Tab, whereas you'd use ItemHeaderTemplate if you wanted to apply the same template to every Tab.

active/tab/tabcontrol.md
stmoy: Good catch - yes.

active/tab/tabcontrol.md
```diff
@@ -0,0 +1,245 @@
...
+* Consider maintaining a list of recently closed Tabs. If the user presses CTRL+SHIFT+T, they expect recently closed tabs to be reopened.
+* If users can perform more commands on a Tab than just closing the Tab (for example, pinning a Tab or duplicating a Tab), consider adding a context menu to the Tab. 
+
+# API Notes
+<!-- Give a one or two line description of each API (type
+and member), or at least the ones that aren't obvious
+from their name.  These descriptions are what show up
+in IntelliSense. -->
+
+### TabControl properties and events
```
YuliKl: Thinking out-loud if Tab**s**Control might be a better name than TabControl. Because there's going to be a TabItem class, I think it might feel more intuitive to place multiple TabItem controls into a TabsControl.

active/tab/tabcontrol.md
stmoy: There are a lot of "Headers", I agree. This largely comes from WPF, which had TabControl and TabItem and their corresponding Header properties as well. 

I'm open to suggestions here. One consideration is the discussion about "TabStripPlacement" which could support Tabs on the left. We're not supporting placement for the time being, but I do want to avoid calling the content "LeftTabContent/RightTabContent" because that won't make sense if we do support TabStripPlacement.

active/tab/tabcontrol.md
stmoy: Auto by default, Auto maps to Persistent.

active/tab/tabcontrol.md
stmoy: Part of the reason we chose "TabControl" specifically was because that's what WPF has. I suspect the API folks have a point of view on this topic, so I'll lean on them for naming.

active/tab/tabcontrol.md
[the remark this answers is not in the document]
MikeHillberg: Typical tab control behavior is to switch tabs with Control+Tab, but the unmodified tab key moves keyboard focus like most other cases.

active/tab/tabcontrol.md
```diff
@@ -0,0 +1,252 @@
...
+### Selecting a Tab
+When a Tab has focus, pressing SPACE or ENTER will select that Tab.
+
+### Shortcuts for moving between tabs
+Ctrl+TAB will select the next Tab.
+Ctrl+Shift+TAB will select the previous Tab.
+Tab selection cycles (meaning if the user has selected the last tab and presses Ctrl+TAB, the first tab becomes selected).
+
+Ctrl+1 through 8 will automatically select the corresponding Tab.
+Ctrl+9 will automatically select the last Tab.
```
MikeHillberg: This is non-standard tab control behavior (Control+1...9). If an app wants to have this behavior, it can always be implemented in the app code.

active/tab/tabcontrol.md
YuliKl: How about using prefix/suffix terminology? Something like calling the TabHeader property TabsPrefix or TabStripPrefix, and similar with Suffix instead of Footer.

active/tab/tabcontrol.md
```diff
@@ -0,0 +1,252 @@
...
+<!-- Give a one or two line description of each API (type
+and member), or at least the ones that aren't obvious
+from their name.  These descriptions are what show up
+in IntelliSense. -->
+
+### TabControl properties and events
+
+| Property | Type | Description |
+|:-------- |:---- |:----------- |
+| CanCloseTabs | bool | Default value for the item if it doesn't specify a IsClosable value. |
```
teaP: I wouldn't say this value is a default -- if it's False it overrides the IsClosable value on the tabitems.

active/tab/tabcontrol.md
```diff
@@ -0,0 +1,252 @@
...
+| CloseButtonOverlay | enum | Describes the behavior of the close button of unselected tabs. Values are {Auto, OnHover, Persistent}. Default is Auto. In WinUI 2.2, Auto maps to Persistent.  |
```
teaP: What is the reason for having this? I would suggest we just pick one behavior for now and implement only that.

active/tab/tabcontrol.md
```diff
@@ -0,0 +1,252 @@
...
+| ItemHeaderTemplate | DataTemplate | Default template for the item if no template specified. |
+| SelectedTabWidth | double | The size of the selected tab header. |
```
teaP: I don't think this is needed, even if we have compact tabs; the tab would just size to content and respect the max/min tab sizes defined in the resources.

active/tab/tabcontrol.md
```diff
@@ -0,0 +1,252 @@
...
+| TabHeader | object | Content to the left of the tab strip. |
+| TabHeaderTemplate | DataTemplate | Template for the Header. |
+| TabFooter | object | Content to the far right of the tab strip. |
+| TabFooterTemplate | DataTemplate | Template for the Footer. |
+| TabActionContent | object | Content immediately to the right of the tabs |
```
teaP: I don't quite see the reason for this -- if we give them the Footer area, they can align items inside of there however they want.

active/tab/tabcontrol.md
```diff
@@ -0,0 +1,252 @@
...
+| TabActionContentTemplate | DataTemplate | Template for the ActionContent. |
+| TabWidthBehavior | enum | Specifies how the tabs should be sized. Values are {Actual, Equal, Compact}. Default is Actual. |
```
teaP: So, based on our last meeting with the terminal folks, I'm still not sure which of these is necessary. I would like to err on the side of doing less for v1 and only adding things later as there is need for it.

active/tab/tabcontrol.md
```diff
@@ -0,0 +1,252 @@
...
+|:-------- |:---- |:----------- |
+| Content | object | The main content that appears in the tab. |
+| Header | object | The content that appears inside the tab itself.  |
+| HeaderTemplate | DataTemplate | Template for the header object. |
+| Icon | IconElement | Icon for the tab. |
+| IsClosable | bool | Determines if the tab shows a close button. |
+
+| Event | Description |
+|---|---|
+| TabClosing | Fires when a tab's close button is clicked. |
```
teaP: I don't think this event is necessary, since there's one on the TabControl itself.

active/tab/tabcontrol.md
```diff
@@ -0,0 +1,245 @@
...
+  etc.-->
+
+<!-- As an example of this section, see the Examples section for the PasswordBox control 
+(https://docs.microsoft.com/windows/uwp/design/controls-and-patterns/password-box#examples). -->
+
+## To replicate the behavior of Microsoft Edge
+
+``` xml
+<TabControl TabWidthBehavior="Equal"
+            CanCloseTabs="True"
```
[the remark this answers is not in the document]
mdtauk: The last remaining Tab should probably close the window, and so should not display a close button.

Also if Tabs are used for docked toolbars or a property window just like in Visual Studio - you would probably want to drag a tab off, into a floating window/dialog, which could then be closable or dockable.

active/tab/tabcontrol.md
mdtauk: Rather than Header and Footer, perhaps it should be **Before** and **After** which will be contextually relevant whether Tabs are at the Top, Bottom, Left or Right - and would also work for localised situations.

active/tab/tabcontrol.md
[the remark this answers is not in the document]
mdtauk: Tabs could be called **TabItem**, and so the Items name would make sense

active/tab/tabcontrol.md
```diff
@@ -0,0 +1,252 @@
...
+This spec therefore covers "document-style tabs" specifically as this platform gap isn't easily solved with an app-side workaround. 
+
+# Description
+<!-- Use this section to provide a brief description of the feature.
+For an example, see the introduction to the PasswordBox control 
+(http://docs.microsoft.com/windows/uwp/design/controls-and-patterns/password-box). -->
+
+The Tab control is a collection of tabs that each represents a new page or document in your app. Tab controls are useful when your app has several pages of content and the user expects to be able to add, close, and rearrange the pages.
+
+**Is this the right control?**
```
mrlacey: Is this the right control if I wanted to create an older style MDI UI where the page has multiple open documents but it's not possible to see the tabs for each document?

active/tab/tabcontrol.md
```diff
@@ -0,0 +1,252 @@
...
+```
+
+## Create a new window when tearing out tabs
+
+![An example picture showing that a tab can be torn out and moved to its own window.](images/tab-new-window.png)
+
+See the [TabView tear out sample](https://github.com/windows-toolkit/Sample-TabView-TearOff/tree/master/TabViewTear) for a more complete sample.
+
+``` xml
+<TabControl CanDragItems="True"
```
mrlacey: `CanDragItems` is not listed as a property below. Is it meant to be?

active/tab/tabcontrol.md
```diff
@@ -0,0 +1,252 @@
...
+            CanReorderItems="True"
```
mrlacey: `CanReorderItems` is not listed as a property below. Is it meant to be?

active/tab/tabcontrol.md
```diff
@@ -0,0 +1,252 @@
...
+
+### Selecting a Tab
+When a Tab has focus, pressing SPACE or ENTER will select that Tab.
+
+### Shortcuts for moving between tabs
+Ctrl+TAB will select the next Tab.
+Ctrl+Shift+TAB will select the previous Tab.
+Tab selection cycles (meaning if the user has selected the last tab and presses Ctrl+TAB, the first tab becomes selected).
+
+Ctrl+1 through 8 will automatically select the corresponding Tab.
```
mrlacey: Is this visible tabs?
I'm assuming there will be a way to support more tabs than can be see on screen at any one time and a way to switch between tabs that aren't visible--both through keys and directly through the UI. What will this be?

active/tab/tabcontrol.md
mrlacey: Will `Ctrl+9` select the last tab if there are more than 9?

active/tab/tabcontrol.md
```diff
@@ -0,0 +1,252 @@
...
+Ctrl+9 will automatically select the last Tab.
+
+### Closing a Tab 
+Hitting CTRL+W or CTRL+F4 will automatically close the selected Tab (assuming it is closable). 
+
+### Keyboard guidance for App Developers
+The above sections outline built-in keyboarding behavior provided by the TabControl. However, there are certain expected keyboard shortcuts that you will be responsible for implementing.
```
mrlacey: Will samples be provided for these too?

active/tab/tabcontrol.md
```diff
@@ -0,0 +1,252 @@
...
+Ctrl+9 will automatically select the last Tab.
+
+### Closing a Tab 
+Hitting CTRL+W or CTRL+F4 will automatically close the selected Tab (assuming it is closable). 
+
+### Keyboard guidance for App Developers
+The above sections outline built-in keyboarding behavior provided by the TabControl. However, there are certain expected keyboard shortcuts that you will be responsible for implementing. 
+* If your Tab control supports adding a new Tab, users expect CTRL+T to open a new tab
+* Consider maintaining a list of recently closed Tabs. If the user presses CTRL+SHIFT+T, they expect recently closed tabs to be reopened.
+* If users can perform more commands on a Tab than just closing the Tab (for example, pinning a Tab or duplicating a Tab), consider adding a context menu to the Tab.
```
mrlacey: For an app to be able to pin a TabItem there are things that may be wanted that aren't covered in this spec.
- Can TabItem order be changed programmaticly? (As the underlying type for the Items isn't specified I don't want to assume.)
- Can the ability to reorder individual Items be prevented? (So can do things like put all the pinned tabs together.)

active/tab/tabcontrol.md
```diff
@@ -0,0 +1,252 @@
...
+| CanCloseTabs | bool | Default value for the item if it doesn't specify a IsClosable value. |
+| CloseButtonOverlay | enum | Describes the behavior of the close button of unselected tabs. Values are {Auto, OnHover, Persistent}. Default is Auto. In WinUI 2.2, Auto maps to Persistent.  |
+| ItemHeaderTemplate | DataTemplate | Default template for the item if no template specified. |
+| SelectedTabWidth | double | The size of the selected tab header. |
+| TabHeader | object | Content to the left of the tab strip. |
+| TabHeaderTemplate | DataTemplate | Template for the Header. |
+| TabFooter | object | Content to the far right of the tab strip. |
+| TabFooterTemplate | DataTemplate | Template for the Footer. |
+| TabActionContent | object | Content immediately to the right of the tabs |
+| TabActionContentTemplate | DataTemplate | Template for the ActionContent. |
```
mrlacey: If `TabActionContent` can take any object, does it really need a separate template property?
What will happen to anything specified as content without a template?

active/tab/tabcontrol.md
```diff
@@ -0,0 +1,252 @@
...
+| TabWidthBehavior | enum | Specifies how the tabs should be sized. Values are {Actual, Equal, Compact}. Default is Actual. |
+
+| Event | Description |
```
mrlacey: If reordering is allowed, will there be an event to indicate `TabOrderChanged`?
This may be needed when wanting to connect the display order with the underlying data order.

A `TabOrderChanging` event (with the ability to cancel) where the app wants all TabItems of a certain type (e.g. pinned) to be kept together would also be useful.

active/tab/tabcontrol.md
```diff
@@ -0,0 +1,252 @@
...
+|---|---|
+| TabClosing | Fires when a tab is about to be closed. Can be cancelled to prevent closure. |
```
mrlacey: Will there also be a corresponding `TabClosed` event?
Having both is a common convention and allows separating the logic or determining whether closing is allowed and what to do after a tab has been closed.

active/tab/tabcontrol.md
```diff
@@ -0,0 +1,252 @@
...
+| TabDraggedOutside | Fires when a Tab is dragged outside of the Tab bar. |
```
mrlacey: What about if a tab that was previously dragged outside and launched in a new Window is dragged back to the view. Will it be possible to drag that tab back into the original control? If so, will custom code need to be written for that which will require an equivalent event? (and sample)

active/tab/tabcontrol.md
```diff
@@ -0,0 +1,252 @@
...
+
+### TabItem properties and events
+
+| Property | Type | Description |
+|:-------- |:---- |:----------- |
+| Content | object | The main content that appears in the tab. |
+| Header | object | The content that appears inside the tab itself.  |
```
mrlacey: What happens regarding the display of the `Header` and `Icon` in a `TabItem` when there is insufficient space to display everything for the number of open tabs and the size of the containing Window? 
It Header content truncated? (remembering that it might be more that text if supporting custom templates.)
Will it slim down to just Icons?
If partially obscured/truncated or just showing the Icon, will there be a tool tip on mouse over to see the full value?

active/tab/tabcontrol.md
```diff
@@ -0,0 +1,252 @@
...
+
+# API Details
+<!-- The exact API, in MIDL3 format (https://docs.microsoft.com/en-us/uwp/midl-3/) -->
+
+# Appendix
+<!-- Anything else that you want to write down for posterity, but 
+that isn't necessary to understand the purpose and usage of the API.
+For example, implementation details. -->
+
+![Parts of a tab item](https://user-images.githubusercontent.com/25991996/53277900-35beed00-36bb-11e9-910e-36119f7ad4af.png)
```
mrlacey: These images include the text `TabView Control Header` and `TabView Control Footer` at the top and bottom of each image respectively. What are these, as they're not covered in the spec? Please remove if not actually part of the spec as they're potentially confusing.

active/tab/tabcontrol.md
```diff
@@ -0,0 +1,252 @@
...
+
+See the [TabView tear out sample](https://github.com/windows-toolkit/Sample-TabView-TearOff/tree/master/TabViewTear) for a more complete sample.
+
+``` xml
+<TabControl CanDragItems="True"
+            CanReorderItems="True"
+            TabDraggedOutside="TabControl_TabDraggedOutside">
+```
+``` csharp
+// NOTE: The app is responsible for writing this code. We will provide a sample that may look something like:
```
mrlacey: What will happen if this code isn't written? (Or the event handled in general?)
Will the tab just disappear? Or will it simply return to where it was before?

In what scenarios might a different behavior than opening in a new window be wanted? Are there considerations that need to be made for these other scenarios?

Will a sample also be included that shows how to handle tearing off a tab that was created from an tab in a data bound collection?

active/tab/tabcontrol.md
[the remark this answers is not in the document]
mrlacey: Having the app close when the app is closed should be a decision made on a per app basis. 
There are reasons for some apps to have other content displayed when no tabs/documents are open.
The control also needs to be able to support showing no tabs, such as if bound to an empty collection.

active/tab/tabcontrol.md
mdtauk: If you are dragging back onto the TabControl bar, then the control could automatically detect that.  Dragging onto the window itself - maybe not.

active/tab/tabcontrol.md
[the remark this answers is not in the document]
Poopooracoocoo: > Having the app close when the _app_ is closed should be a decision made on a per app basis.

 Do you mean when the _tab_ is closed?

active/tab/tabcontrol.md
Poopooracoocoo: On browsers it does I'd imagine that it would do that here too.

active/tab/tabcontrol.md
Poopooracoocoo: and what about scrollbars?

active/tab/tabcontrol.md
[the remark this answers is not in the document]
mdtauk: I don't mean the TabControl should close the app when the final tab is closed by default - but that should be the official guidance, and expected behaviour.  This is when the Tabs are used as the shell.  When used for toolbars and property inspectors etc, then this is different.

active/tab/tabcontrol.md
stmoy: This will be inherited from ListViewBase (or more specifically, ItemsControl) which has an OnItemsChanged event. 

ItemsControl doesn't have the equivalent of *Changing, but this can be accomplished by listening to other events (like the DragItemsStarted and Completed events) and preventing the reorder.

active/tab/tabcontrol.md
stmoy: The Tab strip area will be contained within a scrollviewer. The tabs themselves will layout relative to their mode (Actual, Equal, Compact). For example, in the scenario where the tabs are Actual, and the app has too many of them, they will each shrink to content and any remaining tabs will clip behind the scrollviewer's bounds. 

The header content will not be truncated unless the app explicitly sets the width of the tab to be less than the natural width of the content. Similarly, it will only slim down to just the icons in the Compact mode.

active/tab/tabcontrol.md
stmoy: Good catch. I will remove. (Leaving this comment open until I do so.)

stmoy:
@YuliKl @mdtauk @mrlacey @Poopooracoocoo @MikeHillberg @teaP 

Thank you all for the great feedback, suggestions, and ideas. I replied to a bunch of the open threads in the PR, but there are still a handful that are outstanding. The ones that are open but haven't been replied to are the ones that I need to think more about -- working on it!

Additionally I've compiled some open questions at the end of the doc for discussion. We'll be discussing internally (and I'll be sure to reply with our thoughts!), but I'd also love to know your thoughts on the following open questions:
* Should TabControl support tabs that don't close?
* Does TabControl need two areas to the right of the tab pane (as spec'd) or can there just be one content area?
* The spec outlines several keyboard behaviors which may be considered "non-standard". Should those be built into the control directly?
* Compact mode isn't required for Terminal v1 - can we push this feature out to v2? 
* Do we need a TabClosed event? It would be useful from a consistency point of view, but is it required for v1?

active/tab/tabcontrol.md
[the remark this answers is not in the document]
Felix-Dev: @stmoy
Having tabs which can never be closed just sounds weird to me. Just as as user can easily add new tabs, the user should also be able to close tabs whenever he/she likes. There shouldn't be a case of having tabs which can never be closed. In all cases, a user should be able to request closing a tab with the tabClosing event handling temporary cases where closing might come with some consequence - such as lost data or an aborted operation - and provide an option to keep the tab alive.

active/tab/tabcontrol.md
cinnamon-msft: I agree that we should pick one behavior and only implement that. Preferably, the behavior is 'persistent' so it aligns with the new version of Edge using Chromium.

mdtauk: > I agree that we should pick one behavior and only implement that. Preferably, the behavior is 'persistent' so it aligns with the new version of Edge using Chromium.

The Edgium tabs, obviously derive from the Chromium project, so it may be worth asking someone on that team to provide some guidance.

There are also the WPF Tabs, and the Win32 Tabs to derive from.

Felix-Dev: What is the reason for the change from Edge UWP tabs (where the closeButtonOverlay is of type OnHover)? I'm just looking at Firefox, Edge Chromium,...and I have to say having a close button on *every* tab seems like overkill to me. In edge UWP, the additional space is used to display more of the tab title which seems far more useful to me then having a close button on every tab.

The only reason I could see right now is better touch support, i.e. you can directly close unselected tabs with touch without having to first select them so that the close button will appear. Not sure if that is a valid point or how Edge UWP is handling this case (no touch inout capable display here).

active/tab/tabcontrol.md
cinnamon-msft: Eventually, we'd like to have tab width as a setting users can choose between. For now, this feature is a "nice-to-have," but it's not required for our v1 😊

stmoy:
# Tab Control Team Spec Review Summary: # 

Hi all! Following up on the discussion topics from the Open Questions list. I'll be updating the spec to reflect the changes discussed here and feedback from the PR review. 

* TODO: Add sample for tab recombination. There is some nuance where tearing out the last tab of Window A and recombining it into an existing Window B should close the window A.

* Q: Should TabControl have a global knob for making the whole set of tabs not closeable?

A: No. No global knob is needed, as individual tabs can still mark themselves as not closeable. 

* Q: Could Ribbon be built with TabControl? Should it even be called TabControl?

A: TabControl is being built specifically to support document style tabs and the first customer is the Terminal. There are a set of things that the proposed control won’t support yet (like positioning, rows of tabs, etc.). Rather than sitting on the “good name” (TabControl) the proposed change was to update the name of the control to “DocumentTabControl” or something to help tip developers that this might not be the tabs they are looking for. To that end, Ribbon is an “out of scope” scenario.

* Q: Should we provide a “+” button in the control itself?

A: Yes. We will provide a “+” button in the control. It will have the following capabilities: 1) it can be turned off; 2) it will fire an event when clicked

* Q: Should we have two areas to the right of the tab pane (as spec’d) or can there just be one content area?

A: Inconclusive. We discussed that having two areas would be an easier app dev experience, but there are questions about how to properly lay out the two areas. We will do a side-by-side comparison between the two areas and the one area to figure out the delta in the developer experience.

* Q: The spec outlines several keyboard behaviors which may be considered "non-standard". Should those be built into the control directly?

A: Inconclusive. We definitely don’t want the tab control to “eat” keyboard presses that the app may use – and the CTRL+# combination is expected to be fairly common. Maybe we could have an API to opt-in to the keyboard behaviors? We will have another followup checkpoint specifically on keyboarding and accessibility. 

* Q: Compact mode isn't required for Terminal v1 - can we push this feature out to v2?

A: Yes. This isn’t a P0 requirement and can be cut for now.

* Q: Do we need a TabClosed event? It would be useful from a consistency point of view, but is it required for v1?

We didn’t discuss this.

ZodmanPerth:
Does code for the `TabControl` spec'd here exist yet or is this just a spec for a control to be written in the near future?  I'd like to take it for a spin, even if I need to compile from a branch to do so.

mdtauk:
> Does code for the `TabControl` spec'd here exist yet or is this just a spec for a control to be written in the near future? I'd like to take it for a spin, even if I need to compile from a branch to do so.

The code exists, but not in the publically released package.  The new Windows Terminal repo is using it and "dogfooding" it.

jesbis:
> Does code for the `TabControl` spec'd here exist yet or is this just a spec for a control to be written in the near future? I'd like to take it for a spin, even if I need to compile from a branch to do so.

@zodmanperth TabView should be available as a preview API in the current prerelease NuGet package:

https://www.nuget.org/packages/Microsoft.UI.Xaml/2.2.190611001-prerelease

Although it will be a bit behind source branches, e.g.:

https://github.com/microsoft/microsoft-ui-xaml/tree/user/teaP/TabViewAdditions

ZodmanPerth:
Thanks folks, I'm playing with the `TabView` control now.

I do have a follow-up question; is the `TabView` control in `Microsoft.UI.Xaml` the control where this spec will be applied, or can I expect to see a `TabControl` to emerge in the near future?

For background, we've got "urgent" need for multiview functionality and are evaluating tear-away tabs as a potential solution.  We would like to contribute and/or help guide development driven by our real-world use case to ensure success (for everyone).

stmoy:
@ZodmanPerth - thank you for trying out the preview control and "kicking the tires" 😀 I'd love to hear any feedback you have about TabControl.

> I do have a follow-up question; is the TabView control in Microsoft.UI.Xaml the control where this spec will be applied, or can I expect to see a TabControl to emerge in the near future?

The exact name/API is still being finalized. We will continue to develop and iterate on the TabView control (ie. there won't be both a TabView and a TabControl - we'll pick one name and update TabView to that name). 

> For background, we've got "urgent" need for multiview functionality and are evaluating tear-away tabs as a potential solution. We would like to contribute and/or help guide development driven by our real-world use case to ensure success (for everyone).

We'd love to hear your thoughts and get your feedback. TabView is still in preview as we iterate, so please be aware that functionality and names are subject to change. However, this is also a great time in the control development lifecycle to provide feedback/contributions.

active/tab/tabcontrol.md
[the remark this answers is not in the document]
stmoy: > Having tabs which can never be closed just sounds weird to me. 

The canonical example of this is pinning a tab in Edge. The tab then becomes a "permanent" tab until the user either unpins it or closes it from a context menu.

> Given your intro, is it really necessary to have a CanCloseTabs property?

Looping back the original question - we determined that the answer is "no". It's possible to have individual tabs that are not closeable, but there shouldn't be a control-wide knob to make them all not closeable. 

Speaking of closeable, I'm resolving this conversation as we've closed the original topic.

active/tab/tabcontrol.md
stmoy: Thinking more upon this - the WinUI framework needs to be sensitive to shortcuts that the app has built in. The CTRL+# combination (in addition to CTRL+_letter_ ) are very common app-side shortcuts that the control should not "stomp over". We will be thoughtful about which shortcuts to include, and for the ones that we don't build into the control, we will provide a sample if the app developer wants to emulate that behavior.

active/tab/tabcontrol.md
stmoy: > Rather than Header and Footer, perhaps it should be Before and After which will be contextually relevant whether Tabs are at the Top, Bottom, Left or Right - and would also work for localised situations.

Interesting idea! Although we have precedent for generic areas before/after above/below left/right of a control, we don't have precedent for the wording "Before" and "After". Instead, we traditionally use the words "Header" and "Footer". 

Given that context, we are leaning towards having a single content area to the right of the Tab strip area called the "TabStripFooter".

active/tab/tabcontrol.md
stmoy: Sounds good - thanks! I will remove this from the v1 spec.

active/tab/tabcontrol.md
stmoy: Will remove.

active/tab/tabcontrol.md
stmoy: > I agree that we should pick one behavior and only implement that. Preferably, the behavior is 'persistent' so it aligns with the new version of Edge using Chromium.

Thanks @cinnamon-msft! For v1 we'll focus on persistent tabs and leave the hover tabs for v2. 

There's still an open question about visualization, but we are actively iterating on visual design, so lets push pause on that topic until the team has a baseline point of view.

ZodmanPerth:
> We'd love to hear your thoughts and get your feedback. TabView is still in preview as we iterate, so please be aware that functionality and names are subject to change. However, this is also a great time in the control development lifecycle to provide feedback/contributions.

Thank @stmoy .  I've had a good look at the [TabView TearOff Sample](https://github.com/windows-toolkit/Sample-TabView-TearOff) and initial impressions are good.  I like how customisable the UI of the control is, and the events for managing tear off seem logical.  Progress so far looks promising.

At present we have two concerns:

* We want to provide the same tear-away tab experience as Edge does; being able to drag and position the new window on a secondary monitor.  We're concerned about [Known issue #1](https://github.com/windows-toolkit/Sample-TabView-TearOff#known-issues) which states it's not currently possible.  What are the chances of this being possible before Win UI 2.2 drops?  Or Win UI 3.0?  Or beyond?

* `AppWindow` does not have native support for DirectX ([see Remarks from the docs](https://docs.microsoft.com/en-us/uwp/api/windows.ui.windowmanagement.appwindow#remarks)) which may prevent some of our drawing scenarios.  Further investigation needs doing on our side, however development of tear-off windows for the `TabControl` appears to be moving with `AppWindow` as a focus.  I do hope support for this edge case is not lost in the rush to complete `TabControl`; it would be disappointing to see our first concern addressed only to find we cannot use `TabControl` because of dependence on `AppWindow` to complete the use case.

I know you're all working hard to get this ready; we really appreciate your effort!

We hope you will take our concerns with you as you make decisions for this exciting control.  Users of our software have been asking for this feature for some time and we would love to be able to provide it for them using the `TabControl` (or whatever it ends up being named).

stmoy:
@ZodmanPerth - thank you for the feedback! Here are my thoughts on the two concerns raised: 

> We want to provide the same tear-away tab experience as Edge does; being able to drag and position the new window on a secondary monitor. We're concerned about Known issue #1 which states it's not currently possible. What are the chances of this being possible before Win UI 2.2 drops? Or Win UI 3.0? Or beyond?

Tab tear-away is an important feature for this control and is considered a P0 scenario for v1. We just haven't built it in yet. 

> AppWindow does not have native support for DirectX (see Remarks from the docs) which may prevent some of our drawing scenarios. Further investigation needs doing on our side, however development of tear-off windows for the TabControl appears to be moving with AppWindow as a focus. I do hope support for this edge case is not lost in the rush to complete TabControl; it would be disappointing to see our first concern addressed only to find we cannot use TabControl because of dependence on AppWindow to complete the use case.

I'll bubble this concern up to the AppWindow team. The control itself doesn't have a coupling to AppWindow - I believe that you could achieve a similar experience using [CoreApplicationView](https://docs.microsoft.com/en-us/windows/uwp/design/layout/show-multiple-views). @michael-hawker , do you have an example of this?

ZodmanPerth:
Thanks @stmoy.

I've now completed a spike for how we would use `AppWindow` and it looks like we may be OK.  It would be great to discuss our needs directly with the AppWindow team if anyone from that team wants to reach out.

All our concerns appear to be being addressed.

Best of luck with the development of the `TabControl`.  I look forward to seeing what you come up with.

vitorgrs:
Wanted to give my feedback on the name.  
Currently XAML Controls don't use Control for names. There's no NavigationControl, FlipControl, ListControl, GridControl, etc. It's all NavigationView, FlipView, ListView, GridView.  

So I believe TabView would be more intuitive and more consistent with other controls name.

micahl:
> Wanted to give my feedback on the name.
> Currently XAML Controls don't use Control for names. There's no NavigationControl, FlipControl, ListControl, GridControl, etc. It's all NavigationView, FlipView, ListView, GridView.
> 
> So I believe TabView would be more intuitive and more consistent with other controls name.

Agreed.  I was about to say that for the cases that do have a name that ends in "Control" are more low-policy controls (i.e. ContentControl, ItemsControl), but then I remembered we had a RatingsControl.  I'll defer to @MikeHillberg on naming conventions.  :)  I suspect that one was a case of aligning the name with what was already the prevailing name used elsewhere.

stmoy:
@vitorgrs @micahl - very good points. This control is expected to take up the entire page, similar to NavigationView, so I agree that TabView is probably the way to go!

active/tab/tabcontrol.md
```diff
@@ -0,0 +1,401 @@
...
+<!-- For example, this section is a place to explain why you're adding this API rather than
+modifying an existing API. -->
+
+<!-- For example, this is a place to provide a brief explanation of some dependent
+area, just explanation enough to understand this new API, rather than telling
+the reader "go read 100 pages of background information posted at ...". -->
+
+> See [proposal document](https://github.com/Microsoft/microsoft-ui-xaml/issues/304) for more background
+
+The Tab control is a way to display a set of tabs and their respective content. Tab controls are useful for displaying several pages (or documents) of content while giving a user the capability to rearrange, open, or close new tabs.
```
MikeHillberg: Tab => TabView (throughout)

active/tab/tabcontrol.md
```diff
@@ -0,0 +1,401 @@
...
+For an example, see the introduction to the PasswordBox control 
+(http://docs.microsoft.com/windows/uwp/design/controls-and-patterns/password-box). -->
+
+The Tab control is a collection of tabs that each represents a new page or document in your app. Tab controls are useful when your app has several pages of content and the user expects to be able to add, close, and rearrange the pages.
+
+**Is this the right control?**
+
+Use a TabView to help the user manage multiple app pages or documents within the same window. 
+
+Do not use a TabView to display a static set of tabs that the user cannot rearrange, open, or close. Use a Pivot or top NavigationView instead.
```
MikeHillberg: The name "TabView" doesn't imply this restriction. Have you discussed some more specific name?

active/tab/tabcontrol.md
```diff
@@ -0,0 +1,401 @@
...
+            AddTabButtonClick="NewTab_Click">
+    ...
+</TabView>
+```
+
+## Put tabs in the titlebar area
+
+![A sample app that shows the Tab control can extend into the titlebar area](images/tab-extend-to-title.png)
+
+This sample demonstrates how to extend the TabView into the title bar area and also specify a portion of the UI as the draggable region. Per the [titlebar customization guidelines](https://docs.microsoft.com/en-us/windows/uwp/design/shell/title-bar), we must set a [specific drag region](https://docs.microsoft.com/en-us/windows/uwp/design/shell/title-bar#draggable-regions). If we don't specify the drag region, the entire titlebar area remains draggable (meaning input won't be routed to the tabs, making the tabs unclickable).
```
MikeHillberg:
```suggestion
This sample demonstrates how to extend the TabView into the title bar area and also specify a portion of the UI as the draggable region, using [CoreApplicationViewTitleBar.ExtendViewIntoTitleBar](https://docs.microsoft.com/uwp/api/Windows.ApplicationModel.Core.CoreApplicationViewTitleBar.ExtendViewIntoTitleBar). Per the [titlebar customization guidelines](https://docs.microsoft.com/en-us/windows/uwp/design/shell/title-bar), we must set a [specific drag region](https://docs.microsoft.com/en-us/windows/uwp/design/shell/title-bar#draggable-regions) (the space that a user can use to drag the window around the screen using mouse or touch). If we don't specify the drag region, the entire titlebar area remains draggable (meaning input won't be routed to the tabs, making the tabs unclickable).
```

active/tab/tabcontrol.md
```diff
@@ -0,0 +1,401 @@
...
+    x:Name="TabRoot"
+    AddTabButtonClick="AddTabButton_Click"
+    >
+    <!-- ... some tabs ... -->
+</controls:TabView>
+```
+
+``` csharp
+
+private void AddTabButton_Click(object sender, RoutedEventArgs e)
```
MikeHillberg: Sender type should be TabView, event arg type should be TabViewAddTabButtonClickEventArgs (yes, it's a long name).

active/tab/tabcontrol.md
```diff
@@ -0,0 +1,401 @@
...
+TODO -- should this be part of the "tab tear out" sample above?
+
+# Remarks
+<!-- Explanation and guidance that doesn't fit into the Examples
+section.  For example, see the Remarks for the MediaPlayerElement 
+(https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.Controls.MediaPlayerElement#remarks). -->
+
+## Keyboarding behavior
+### TAB and Arrow key behavior
+The TabView has three content areas: TabStrip, TabStripHeader (to the left of the TabStrip) and TabStripFooter (to the right of the TabStrip). When the user hits the TAB key, focus moves between the content areas. When focus moves into the Tab content area, the selected Tab gains focus. Once the user is in the Tab content area, pressing the arrow key will move focus between the tabs. Arrow focus is trapped inside the Tab content area.
```
MikeHillberg: This could use a picture and simple example.

MikeHillberg: I'm having trouble parsing this. It doesn't necessarily need to be longer, or if it is, maybe move it to the appendix. Right now I'm just not sure what it's saying.

"When the user hits the TAB key, focus moves between the content areas." ... Doesn't it move _within_ the content area too?

"When focus moves into the Tab content area, the selected Tab gains focus" ... How do you move into the tab content area without selecting the tab first.  And does "tab content area" mean the Content of a TabViewItem? In Edge you use arrows to tab between the tab item headings, then Enter/Space to select a tab.

active/tab/tabcontrol.md
```diff
@@ -0,0 +1,401 @@
...
+<!-- Give a one or two line description of each API (type
+and member), or at least the ones that aren't obvious
+from their name.  These descriptions are what show up
+in IntelliSense. -->
+
+### TabView properties, events, methods
+
+| Property | Type | Description |
+|:-------- |:---- |:----------- |
+| AddButtonCommand | ICommand | Gets or sets the command to invoke when the Add button is tapped. |
```
MikeHillberg: Is this the same as the AddTabButton? So this should be AddTabButtonCommand?

active/tab/tabcontrol.md
```diff
@@ -0,0 +1,401 @@
...
+| TabDraggedOutside | Fires when a Tab is dragged outside of the Tab bar. |
+
+| Method | Description |
+|---|---|
+| DependencyObject ContainerFromItem(object item) | Returns the generated container for the specified item. |
+| DependencyObject ContainerFromIndex(Int32 index) | Returns the generated container for the specified index. |
+
+### TabViewItem properties and events
+
+| Property | Type | Description |
```
MikeHillberg: Once we have the IDL, we should remove this Type column. It's not entirely correct, and once we have the IDL it just creates the overhead of staying in sync.

active/tab/tabcontrol.md
```diff
@@ -0,0 +1,401 @@
...
+| Event | Description |
+|---|---|
+| AddTabButtonClick | Fires when the + button is clicked. |
+| SelectionChanged | Occurs when the currently selected tab changes. |
+| TabClosing | Fires when a tab is about to be closed. Can be cancelled to prevent closure. |
+| TabDraggedOutside | Fires when a Tab is dragged outside of the Tab bar. |
+
+| Method | Description |
+|---|---|
+| DependencyObject ContainerFromItem(object item) | Returns the generated container for the specified item. |
```
MikeHillberg: Is TabView an ItemsControl, so this is inherited? If so, no need to call it out here. If not, do we need these?

active/tab/tabcontrol.md
stmoy: > "When the user hits the TAB key, focus moves between the content areas." ... Doesn't it move within the content area too?

In the left and right content areas which have app-provided content, TAB will naively do whatever the app says (which will be moving between the elements in those areas by default). However, the behavior is different in the tab strip area. When focus is outside of the tab strip area and the user presses TAB, it moves focus to the selected tab. If the user presses TAB again, it will move focus out of the tab strip area. If focus is within the tab strip area and the user presses LEFT/RIGHT arrows, it will move focus (not selection). Once focus has moved, the user can press ENTER/SPACE to select the focused item. 

I shouldn't have said "Tab content area" - I meant "tab strip". I'll clarify. 

The TL;DR is: we're trying to mimic what Edge does.

active/tab/tabcontrol.md
stmoy: TabView contains a ListView and is not an ItemsControl. 

Since several of the events on TabView return the TabViewItem (and not the underlying data), we needed a way to get the specific data item from the container - hence these APIs.

mdtauk:
The ListView within the TabView, should be given appropriate Add, Remove, Re-arrange animations for TabViewItems being added, removed or dragged.

Edge Canary and Dev has a flag for a new opening animation - which should probably be the basis for the transitions used by the XAML control.

**edge://flags/#new-tab-loading-animation**

active/tab/tabcontrol.md
MikeHillberg: Isn't TabViewItem.Content the data item?

stmoy: TabViewItem.Content is the body of the tab: https://github.com/microsoft/microsoft-ui-xaml-specs/blob/882f458c966600673f0fd8822aba058bf6b4e127/active/tab/images/tab-control-parts-2.PNG

So often it will be a Frame or other containing element.

stmoy:
> The ListView within the TabView, should be given appropriate Add, Remove, Re-arrange animations for TabViewItems being added, removed or dragged.
> 
> Edge Canary and Dev has a flag for a new opening animation - which should probably be the basis for the transitions used by the XAML control.
> 
> **edge://flags/#new-tab-loading-animation**
> 
> https://youtu.be/jn_pUsnKwgM

Thank you for calling out the animations and also the video. Naively, the Tab control will end up having the animations from ListView by default, since TabView contains a ListView. That said, the Edge Canary animations look much more appropriate than the ListView animations - we'll need to investigate how we can make these animations happen.

active/tab/tabcontrol.md
MikeHillberg: Still, TabViewItem.Content is the data item. But I see now, we don't have an ItemFromContainer API here, just the ContainerFrom* APIs. So when you're using ItemsSource, this lets you get the generated TreeViewItem.

active/tab/tabcontrol.md
stmoy: > What will happen if this code isn't written? 

The tab will simply return to where it was before. 

> In what scenarios might a different behavior than opening in a new window be wanted?

We expect that most apps will want to spawn new windows when tearing out; the motivation for this code to be handled by the app and not the control itself is to maintain a separation of concerns. The UWP windowing story is nuanced (with multi-threaded CoreApplicationView) and evolving (with AppWindow), so rather than coupling the control implementation with the windowing implementation, we'll provide samples for apps to use as a starting point.

> Will a sample also be included that shows how to handle tearing off a tab that was created from an tab in a data bound collection?

Yes, if this would be helpful and worthwhile.

active/tab/tabcontrol.md
stmoy: By having this event on TabViewItem, apps can localize their logic to the item itself. I believe there are scenarios for both handling the closing globally (ie. on TabView) and handling the closing locally (ie. on TabViewItem) so even though it's not strictly necessary, I think it is worthwhile to have in both places.

active/tab/tabcontrol.md
stmoy: For now, we'll just have TabClosing. We can add TabClosed later if needed.

active/tab/tabcontrol.md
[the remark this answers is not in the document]
stmoy: We've moved away from TabView **being** a ListView to **containing** a list view, so we have a little more flexibility with how we name these APIs. To that end, we're considering consolidating the Drag/Reorder APIs and renaming them to be more tab-centric.

Poopooracoocoo:
@mdtauk. That's not an Edge exclusive flag. That flag changes the loading spinner on the favicon so you can see it while it's loading. Before there was just a spinner. 

(I actually wrote this comment tons better but since I'm using firefox focus and opened your video in a new tab, my comment was cleared)

mdtauk:
> @mdtauk. That's not an Edge exclusive flag. That flag changes the loading spinner on the favicon so you can see it while it's loading. Before there was just a spinner.
> 
> (I actually wrote this comment tons better but since I'm using firefox focus and opened your video in a new tab, my comment was cleared)

@Poopooracoocoo There are slight tweaks to the movement, but also how the Icon animates into view.  The ProgressRing control is changing by WinUI 3.0 - so hopefully it will work a little better in the tab header, compared to the current version.

active/tab/tabcontrol.md
stmoy: This restriction is purely a guidance recommendation and not a restriction in the API. We briefly discussed other options like "DocumentTabView", "BrowserTabView", "DynamicTabView", etc. 

However, these more specific names have a number of drawbacks, including:
- Discoverability: not starting with the word "Tab" will make it more difficult to find this control
- Coupling with data type: The words "Document" and "Browser" suggest specific data items. For example, DocumentTabView.ItemsSource would probably only take a collection of <Document>. TabView doesn't have this restriction.

Perhaps the wording in the guidance could be updated, as there are legitimate "static tab" scenarios that _could_ be satisified by TabView instead of Pivot/Top NavView. 

To help demonstrate, I built a very basic Ribbon using TabView. (Ribbon has its own set of requirements and scenarios, but I built this to demonstrate that the current TabView is robust enough to be worthy of the name TabView 😉)
https://github.com/stmoy/TabViewSamples/tree/master/RibbonUsingTabView

image

active/tab/tabcontrol.md
stmoy: Sample added!

active/tab/tabcontrol.md
stmoy: As it turns out, tab tear-out and recombination is a non-trivial chunk of work. Who'd have guessed? 😅

To that end, I'll provide a full E2E sample in code rather than including it in the spec.

active/tab/tabcontrol.md
stmoy: Restructured to move this section to the Appendix. Added some pictures and examples.

stmoy:
Hi all! I wanted to provide a super quick update and let you know what's coming next. TL;DR - the feedback provided has helped evolve and solidify the spec to a point where we want to create a "checkpoint of truth" 😉, so I will be completing this PR and submitting the spec to our internal API review board. If you have more comments/feedback, you can still provide them here - we just wanted to give the review board a "clean" PR so they can focus specifically on the spec.

We are actively developing TabView (you can check out its progress on the main [WinUI repo](https://github.com/Microsoft/microsoft-ui-xaml)). Once the API review is completed, we can start the process of moving TabView from "preview" to "release". 

Thank you again for all of your comments/suggestions/feedback! I know there are some features that didn't make it in for v1, so keep an eye out for a v2 tracking issue soon!

mdtauk:
Will the updated spec be published on GitHub to view?

At the moment the current Spec uses images from UWP Edge, the current control design, and a design with selected indicator.

Will the proposed actual design be included, or is that beyond the scope of the spec document?

stmoy:
Good questions! This is a pretty new process for the team so we're still trying to flesh a few things out. With that in mind...

> Will the updated spec be published on GitHub to view?

The current spec (as it exists in this PR) is the most up-to-date spec for v1 (sans my answer to your next question 😉)

> At the moment the current Spec uses images from UWP Edge, the current control design, and a design with selected indicator.
Will the proposed actual design be included, or is that beyond the scope of the spec document?

Yup, you've touched on an area that is still being actively iterated upon. We're working with our design team to come up with better and more representative images for the spec and docs. Since the next step in the process is API review, and the API board isn't as concerned with the visual spec, we didn't want to block on that piece. Before we check into master, we'll have more representative images of what our actual POR is.

mdtauk:
@stmoy  Thank you for that.  I think I have seen others "cleaning" up Specs and Proposals for pitching to teams - so wasn't sure if that is what happens with API reviews also.

I am sure the Windows Terminal team are looking forward to getting their hands on a V1 release, and addressing the many design related feedback items they have received.

I look forward to seeing how First Party apps implement it, or even the Windows Shell itself (I'm looking at you Property Dialog) 😉

PR #35, opened 2019-07-12
Checkbox consistency detailed design PR

chigy:


active/checkboxes/checkbox.md
```diff
@@ -0,0 +1,71 @@
...
+1. The box fill
+2. The box border
+3. The glyph
+4. The indeterminate marker
+5. Text
+
+### Checkbox colors
+Following colors are used to denote the state of the checkbox:
+1. Fill color of the box
+2. Foreground color of the border
```
chrisglein:
```suggestion
2. Color of the border
```
There's no real foreground/background difference here.

active/checkboxes/checkbox.md
```diff
@@ -0,0 +1,71 @@
...
+Checkbox has mainly 3 styles:
+1. Outline unselected
+2. Fill selected
+3. Fill unselected
+
+![styles](https://github.com/microsoft/microsoft-ui-xaml-specs/blob/user/chigy/checkboxes/active/checkboxes/Images/BaseDesign.png)
+
+Visually, fill style is little more visible and highlights the controls better than the Outline style.
+
+For Windows, we default to using Outline style for Unselcted and Fill style for Selected with exception of GridView.
```
chrisglein:
```suggestion
For Windows, we default to using Outline style for Unselected and Fill style for Selected with exception of GridView.
```

active/checkboxes/checkbox.md
```diff
@@ -0,0 +1,71 @@
...
+3. The glyph
+4. The indeterminate marker
+5. Text
+
+### Checkbox colors
+Following colors are used to denote the state of the checkbox:
+1. Fill color of the box
+2. Color of the border
+3. Foreground color of the glyph
+4. The fill color
```
YuliKl: It's not clear to me fill color of what. All the other line items specify of "the box", "the glyph", etc.

chigy:
```suggestion
4. The fill color of the indeterminate indicator
```

chigy: @YuliKl , I wasn't sure how to call it... Does this sound OK? "The fill color of the indeterminate indicator" Didn't want call it a box since that assumes the visual will always be the same.

YuliKl: Sounds great.

chigy:
@YuliKl and @chrisglein FYI. I've updated the spec with latest visual agreement with the design.

chigy:
> Will the Unselected Outline styles be made consistent with the Outline Button styles, if the proposed Button styles are implemented? It appears as if the Filled styles will match the Primary/Accent button styles, including pressed and hover styles.
> 
> [microsoft/microsoft-ui-xaml#1054](https://github.com/microsoft/microsoft-ui-xaml/issues/1054)

@mdtauk , I do not believe they will be made exactly the same, though they should feel coherent. That said, the actual outline style is not being looked at right now so I'm unable to answer with confidence. If you have a passion, why don't you propose checkbox/button outline style that aligns so we can take a look, perhaps?

mdtauk:
> @mdtauk , I do not believe they will be made exactly the same, though they should feel coherent. That said, the actual outline style is not being looked at right now so I'm unable to answer with confidence. If you have a passion, why don't you propose checkbox/button outline style that aligns so we can take a look, perhaps?

I posted some thoughts in the [Proposal](microsoft/microsoft-ui-xaml#1054) I mentioned

Should there be Outlined style and Filled style for all controls including CheckBoxes?  I really don't know to be honest.

comparison-3
_Windows Phone 7 compared to Windows Phone 8_

Both these styles have existed in the past, but were there good reasons one was swapped for the other before?

Android has Outlined and Default text fields.
image


Fabric/Fast DNA has Outlined, Default, and Primary buttons.
image

image

----

It strikes me that outlined buttons, with a 1px border, would feel a lot more light weight on a form or page.  But Filled buttons may stand out a bit better.  I don't have a firm preference for one or the other, only that Fast DNA includes all three styles, and Windows UIs have moved between filled and outlined control designs in the past, so there is scope to include both styles - if there can be a simple way to switch for devs.

And what the Windows Shell's pick, will probably influence developers choices - but until the modern Shell is in dev's or user's hands - its all just discussion and hope that the design vision comes together.

PR #36, opened 2019-07-18
PR for ProgressRing determinate mode

kathyang:
See proposal documents for:
[determinate mode for ProgressRing](https://github.com/microsoft/microsoft-ui-xaml/issues/688) 
[ProgressRing style](https://github.com/microsoft/microsoft-ui-xaml/issues/837)
[Guidance for Progress controls](https://github.com/microsoft/microsoft-ui-xaml/issues/880)


active/progress/progress.md
```diff
@@ -0,0 +1,107 @@
...
+
+# Remarks
+<!-- Explanation and guidance that doesn't fit into the Examples section. -->
+
+<!-- APIs should only throw exceptions in exceptional conditions; basically,
+only when there's a bug in the caller, such as argument exception.  But if for some
+reason it's necessary for a caller to catch an exception from an API, call that
+out with an explanation either here or in the Examples -->
+
+# API Notes
```
YuliKl: ProgressRing derives from Control but ProgressBar derives from RangeBase. Is there value to changing ProgressRing's base class? And if there is value, is it possible to do so at this time?

active/progress/progress.md
```diff
@@ -0,0 +1,107 @@
...
+
+<!-- As an example of this section, see the Examples section for the PasswordBox control 
+(https://docs.microsoft.com/windows/uwp/design/controls-and-patterns/password-box#examples). -->
+
+The following examples show how to use the isIndeterminate property to change the mode of the ProgressRing and the Value property to change the proportionate amount indicated. 
+
+## Indeterminate ProgressRing
+
+```xml
+<ProgressRing IsActive="True" Height="100" Width="100" IsIndeterminate="True"/>
```
YuliKl: The way this is written, there's an implication that the default value for IsIndeterminate is False. Is that true? If so, will that break existing ProgressRing uses that take for granted its indeterminate nature?

kathyang: This is a good point - I'll add this as an open question. The way this is written is intended to align with ProgressBar.

active/progress/progress.md
kathyang: I think there is value in changing the base class to RangeBase given the addition of the determinate mode. I will determine if this is possible at this time.

active/progress/progress.md
```diff
@@ -0,0 +1,111 @@
...
+# API Details
+<!-- The exact API, in MIDL3 format (https://docs.microsoft.com/en-us/uwp/midl-3/) -->
+
+# Appendix
+<!-- Anything else that you want to write down for posterity, but 
+that isn't necessary to understand the purpose and usage of the API.
+For example, implementation details. -->
+
+# Open Questions
+Using IsIndeterminate as the property aligns with ProgressBar. However, this could break existing uses of ProgressRing that expect it to default to indeterminate mode.
```
chigy: I don't recall you proposing changing the default? I'm not sure if I understand this open issue clearly?

active/progress/progress.md
```diff
@@ -0,0 +1,111 @@
...
+  feature explanation,
+  example code
+  etc.-->
+  
+<!-- Code samples should be in C# and/or C++/WinRT -->
+
+<!-- As an example of this section, see the Examples section for the PasswordBox control 
+(https://docs.microsoft.com/windows/uwp/design/controls-and-patterns/password-box#examples). -->
+
+The following examples show how to use the isIndeterminate property to change the mode of the ProgressRing and the Value property to change the proportionate amount indicated.
```
jesbis: Would it be possible to add a visual mockup/screenshot that shows the difference?

mdtauk:
I would still say the Indeterminate mode should use a gradient rather than a solid shape, to help enforce the idea that progress is ongoing.

Also not sure if this becomes part of the control, but you could include the value percentage if the ProgressRing is large enough.  This is how Xbox has used this kind of control.

image

active/progress/progress.md
jesbis: I think it would also be reasonable to prioritize compatibility with existing ProgressRing behavior over alignment with ProgressBar, i.e. make the default IsIndeterminate="False"

active/progress/progress.md
```diff
@@ -0,0 +1,111 @@
...
+
+>See proposal documents for:
+>
+> [determinate mode for ProgressRing](https://github.com/microsoft/microsoft-ui-xaml/issues/688) 
+>
+> [ProgressRing style](https://github.com/microsoft/microsoft-ui-xaml/issues/837)
+>
+> [Guidance for Progress controls](https://github.com/microsoft/microsoft-ui-xaml/issues/880)
+
+Progress controls indicate to a user that an operation is occuring. This includes ProgressBar and ProgressRing controls. ProgressBar has both a determinate and indeterminate state. ProgressRing only has an indeterminate state.
```
jesbis: Once ProgressRing has this new property there would be closer functional parity between ProgressRing and ProgressBar. What guidance should we give to developers regarding when to use ProgressRing vs. ProgressBar?  Would it strictly be a stylistic difference now?  

If so, would we ideally just have a single control instead with two different views (control templates)?

mdtauk: If you combine ProgressRing and ProgressBar into a single Progress Control - you could have an enum to switch between a Ring or Bar layout.

You would also have to factor in how values like "StartingAngle" and Thickness would be used by the Bar layout.

ryandemopoulos: It might make sense to eventually do this (combine into one control); however, we also have a goal of making the adoption of WinUI 3.0 (especially when coming from UWP XAML that ships as a DLL inside Windows) as seamless as possible.  That means minimal API changes/churn, so I wonder if combination into a single control would be a better target for WinUI 4, rather than an earlier release.  Something to think about, anyway.

mdtauk: There is a new ScrollViewer control that currently coexists with the native UWP one.  Perhaps a combined control could come alongside the existing two before replacing it eventually with intellisense suggesting setting the IsIndeterminate property to true?

ryandemopoulos: Yeah, that's definitely an option we could consider too.  Not sure if I like that better than just having them separate, and then changing to combined in one fell swoop.  But worth some debate/thought at least.

active/progress/progress.md
kathyang: Just added some mockup images for the example!

active/progress/ProgressRing.md
```diff
@@ -0,0 +1,157 @@
...
+IsDeterminate is False
+
+| Value | ShowPaused | ShowError| Image | 
+|:--|:-:| :-:| :-:|
+|  | False | False | ![](images/ProgressRing-determinate-empty.PNG) |
+|  | True | False | ![](images/ProgressRing-determinate-empty.PNG) |
+|  | False | True | ![](images/ProgressRing-determinate-empty.PNG) |
+|  | True | True | ![](images/ProgressRing-determinate-empty.PNG) | 
+| 0 | False | False | ![](images/ProgressRing-determinate-empty.PNG) |
+| 75 | False | False | ![](images/ProgressRing-determinate.PNG) |
```
jesbis: Should this be 0.75 to align with [ProgressBar's default range of 0-1](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.controls.progressbar)?

It might be useful to add Min and Max values to the table as well (particularly if this will inherit from RangeBase).

active/progress/ProgressRing.md
```diff
@@ -0,0 +1,157 @@
...
+## Determinate ProgressRing
+
+```xml
+<ProgressRing IsActive="True" Height="100" Width="100" Value="75"/>
+```
+![](images/ProgressRing-determinate.PNG)
+
+Below are tables showing how ProgressRing is affected by the IsDeterminate, IsActive, Value, ShowPaused, and ShowError properties.
+
+## Not Active ProgressRing
```
jesbis:
```suggestion
## Inactive ProgressRing
```

active/progress/ProgressRing.md
kathyang: Actually after testing a solo ProgressBar, the default range is 0-100 (but it inherits properties from RangeBase to modify the min and max values) .

I've added Min and Max to the API table and limited examples (since designs are all preliminary).

active/progress/progress.md
kathyang: Our team has discussed this and for now we'll keep the controls separate

active/progress/progress.md
kathyang: From team discussion it was decided that ProgressRing won't inherit from RangeBase, as ProgressBar does not use many properties it inherits from RangeBase.

active/progress/ProgressRing.md
```diff
@@ -0,0 +1,170 @@
...
+
+# API Details
+<!-- The exact API, in MIDL3 format (https://docs.microsoft.com/en-us/uwp/midl-3/) -->
+
+# Appendix
+<!-- Anything else that you want to write down for posterity, but 
+that isn't necessary to understand the purpose and usage of the API.
+For example, implementation details. -->
+
+# Open Questions
```
YuliKl: I think there's an open question about the visual representation of a 98%, 99%, 100% full determinate ring given rounded corners, right?

kathyang: Good catch, I just added that open question!

mdtauk: > I think there's an open question about the visual representation of a 98%, 99%, 100% full determinate ring given rounded corners, right?

@YuliKl @kathyang  I imagine what is done with these cases, would be tied to the Border/Stroke Thickness.  So with a thickness of 4, then at the smallest it would look like a 4 x 4 px circle at the 1% end - and a space of 4 at the 99% scale.

PR #39, opened 2019-07-29
Button redlines

YuliKl:
The first phase of listing out button's color values. Does this meet your expectations?

### DIFF
--- a/active/ButtonStyles/ButtonStyles.md
+++ b/active/ButtonStyles/ButtonStyles.md
@@ -57,16 +57,16 @@ Button has the following UI elements that determines visuals:
 | :---------- | :------- | :------- | :------- | :------- |
 | Rest |1. #FF000000 (100% Black) *no change* <br/>2. #33000000 (20% Black) *no change* |1. #FFFFFFFF (100% White) *no change* <br/>2. 100% Accent color *no change* |1. #FF323130 <br/>2.#FFFFFFFF (100% White) <br/>3. #FF8A8886  |1. #FF000000 (100% Black) <br/>2. Transparent color |
 | Hover |1. #FF000000 (100% Black) *no change* <br/>2. #1A000000 (10% Black) |1. #FFFFFFFF (100% White) *no change* <br/>2. 100% AccentLight1 color |1. #FF201F1E <br/> 2. #FFF3F2F1 <br/>3. #FF8A8886 |1. #FF000000 (100% Black) <br/>2. #1A000000 (10% Black) |
-| Pressed |1. #FF000000 (100% Black) *no change* <br/>2. #66000000 (40% Black) *no change* |1. #FFFFFFFF (100% White) *no change* <br/>2. 100% AccentDark1 color  |1. #FF201F1E <br/>2. #FFEDEBE9 <br/>3. #FF8A8886 |1. #FF000000 (100% Black) <br/>2. #33000000 (20% Black) |
+| Pressed |1. #FF000000 (100% Black) *no change* <br/>2. #66000000 (40% Black) *no change* |1. #FFFFFFFF (100% White) *no change* <br/>2. 100% AccentDark1 color |1. #FF201F1E <br/>2. #FFEDEBE9 <br/>3. #FF8A8886 |1. #FF000000 (100% Black) <br/>2. #33000000 (20% Black) |
 | Disabled |1. #66000000 (40% Black) *no change* <br/>2. #33000000 (20% Black) *no change* |1. #66000000 (40% Black) *no change* <br/>2. #33000000 (20% Black) *no change*  |1. #FFA19F9D <br/>2. #FFF3F2F1 <br/> 3. Transparent color  |1. #66000000 (40% Black) <br/>2. Transparent color |
 
 ### Dark Theme Color values
 
 | State | Fill style | Accent style | Outline style | Ghost style |
 | :---------- | :------- | :------- | :------- | :------- |
-| Rest |  |  |  |  |
-| Hover |  |  |  |  |
-| Pressed |  |  |  |  |
-| Disabled |  |  |  |  |
+| Rest |1. #FFFFFFFF (100% White) *no change* <br/>2. #33FFFFFF (20% White) *no change* |1. #FFFFFFFF (100% White) *no change* <br/>2. 100% Accent color *no change* |  |1. #FFFFFFFF (100% White) <br/>2. Transparent color |
+| Hover |1. #FFFFFFFF (100% White) *no change* <br/>2. #1AFFFFFF (10% White) |1. #FFFFFFFF (100% White) *no change* <br/>2. 100% AccentDark1 color |  |1. #FFFFFFFF (100% White) <br/>2. #1AFFFFFF (10% White) |
+| Pressed |1. #FFFFFFFF (100% White) *no change* <br/> 2. #66FFFFFF (60% White) *no change* |1. #FFFFFFFF (100% White) *no change* <br/>2. 100% AccentLight1 color |  |1. #FFFFFFFF (100% White) <br/>2. #33FFFFFF (20% White) |
+| Disabled |1. #66FFFFFF (40% White) *no change* <br/>2. #33FFFFFF (20% White) *no change* |1. #66FFFFFF (40% White) *no change* <br/>2. #33FFFFFF (20% White) *no change* |  |1. #66FFFFFF (40% White) <br/>2. Transparent color |
 
 
--- a/active/ButtonStyles/ButtonStyles.md
+++ b/active/ButtonStyles/ButtonStyles.md
@@ -53,16 +53,16 @@ Button has the following UI elements that determines visuals:
 
 ### Light Theme Color values
 
-| State | 1. Fill | 2. Accent | 3. Outline | 4. Ghost |
+| State | Fill style | Accent style | Outline style | Ghost style |
 | :---------- | :------- | :------- | :------- | :------- |
-| Rest |  |  |  |  |
-| Hover |  |  |  |  |
-| Pressed |  |  |  |  |
-| Disabled |  |  |  |  |
+| Rest |1. #FF000000 (100% Black) *no change* <br/>2. #33000000 (20% Black) *no change* |1. #FFFFFFFF (100% White) *no change* <br/>2. 100% Accent color *no change* |3. #FF8A8886  |1. #FF000000 (100% Black) <br/>2. Transparent color |
+| Hover |1. #FF000000 (100% Black) *no change* <br/>2. #1A000000 (10% Black) |1. #FFFFFFFF (100% White) *no change* <br/>2. 100% AccentLight1 color |  |1. #FF000000 (100% Black) <br/>2. #1A000000 (10% Black) |
+| Pressed |1. #FF000000 (100% Black) *no change* <br/>2. #66000000 (40% Black) *no change* |1. #FFFFFFFF (100% White) *no change* <br/>2. 100% AccentDark1 color  |  |1. #FF000000 (100% Black) <br/>2. #33000000 (20% Black) |
+| Disabled |1. #66000000 (40% Black) *no change* <br/>2. #33000000 (20% Black) *no change* |1. #66000000 (40% Black) *no change* <br/>2. #33000000 (20% Black) *no change*  |  |1. #66000000 (40% Black) <br/>2. Transparent color |
 
 ### Dark Theme Color values
 
-| State | 1. Fill | 2. Accent | 3. Outline | 4. Ghost |
+| State | Fill style | Accent style | Outline style | Ghost style |
 | :---------- | :------- | :------- | :------- | :------- |
 | Rest |  |  |  |  |
 | Hover |  |  |  |  |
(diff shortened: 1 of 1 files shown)


mdtauk:
> The first phase of listing out button's color values. Does this meet your expectations?

Are these values to be hardcoded into the templates, or are they just the literal values for things like "SystemForegroundColor".

If so, I think it would be helpful if that value name could be included in the redlines - so consistency can be maintained with other control templates with the same border colours, or fill/foreground/background colours.

YuliKl:
> > The first phase of listing out button's color values. Does this meet your expectations?
> 
> Are these values to be hardcoded into the templates, or are they just the literal values for things like "SystemForegroundColor".
> 
> If so, I think it would be helpful if that value name could be included in the redlines - so consistency can be maintained with other control templates with the same border colours, or fill/foreground/background colours.

We definitely won't be hardcoding colors into the control templates because Button must continue to support Light, Dark, and High Contrast themes. I debated what best to list and would love @Chigy's input on this question. What makes this a little complicated are the levels of obfuscation between a color value and what actually gets referenced in the control template. 

For example, the Button control template uses the *ButtonBackground* theme resource, which is defined as `<StaticResource x:Key="ButtonBackground" ResourceKey="SystemControlBackgroundBaseLowBrush" />`. In tern, *SystemControlBackgroundBaseLowBrush* is defined as `<SolidColorBrush x:Key="SystemControlBackgroundBaseLowBrush" Color="{StaticResource SystemBaseLowColor}" />`, and finally *SystemBaseLowColor* bottoms out with actual hex values like `<Color x:Key="SystemBaseLowColor">#33FFFFFF</Color>` or `<Color x:Key="SystemBaseLowColor">#33000000</Color>` depending on theme.

So I'm wondering how useful it will be to call out *SystemBaseLowColor*, etc in this table because it doesn't tell the whole story.

mdtauk:
> We definitely won't be hardcoding colors into the control templates because Button must continue to support Light, Dark, and High Contrast themes. I debated what best to list and would love @chigy's input on this question. What makes this a little complicated are the levels of obfuscation between a color value and what actually gets referenced in the control template.
> 
> For example, the Button control template uses the _ButtonBackground_ theme resource, which is defined as `<StaticResource x:Key="ButtonBackground" ResourceKey="SystemControlBackgroundBaseLowBrush" />`. In tern, _SystemControlBackgroundBaseLowBrush_ is defined as `<SolidColorBrush x:Key="SystemControlBackgroundBaseLowBrush" Color="{StaticResource SystemBaseLowColor}" />`, and finally _SystemBaseLowColor_ bottoms out with actual hex values like `<Color x:Key="SystemBaseLowColor">#33FFFFFF</Color>` or `<Color x:Key="SystemBaseLowColor">#33000000</Color>` depending on theme.
> 
> So I'm wondering how useful it will be to call out _SystemBaseLowColor_, etc in this table because it doesn't tell the whole story.
>
This redline document, who is it for...  Those contributing and building the control/templates? Developers who may be styling and re-templating the control in the future?

Giving the final resolved colour is fine for those recreating the control designs - but looses the contextual information the assigned resources give.

Then if you take the Outlined Button Border Colour, would it use the same colour/resource as the CheckBox Border, or the ComboBox, TextBox border colours?  If they all share the same resource, that means there is a cohesive choice on the part of the design team.

There is a heirachy leading to the final resolved colour, so what levels do people _need_ to be made aware of?
> `<Button Border="Black"/>` (Local styling)
>> `OutlinedButtonBorderBrush` (Lightweight Control level styling)
>>> `SystemControlForegroundBaseMediumLowBrush` (Theme level styling)
>>>> `SystemBaseMediumLowColor` (System level styling)
>>>>> `#66FFFFFF` (Resolved colour)

chigy:
@mdtauk , the redline is for the engineer to make the style and make changes to our controls as well as for those who might want to copy the style where they cannot (like WinForms, WPF customers) since I heard those developers express interest for redlines when I talked with them at conferences.

>Then if you take the Outlined Button Border Colour, would it use the same colour/resource as the CheckBox Border, or the ComboBox, TextBox border colours? If they all share the same resource, that means there is a cohesive choice on the part of the design team.

That is the model our color system has already. We don't document this today because we have not heard request for this info, we don't have resources to document and maintain, etc.

mdtauk:
> @mdtauk , the redline is for the engineer to make the style and make changes to our controls as well as for those who might want to copy the style where they cannot (like WinForms, WPF customers) since I heard those developers express interest for redlines when I talked with them at conferences.
> 
If the engineers are the ones updating the controls, then they should surely not be using hex colours, but instead using the appropriate ThemeResources.

chigy:
@mdtauk , that's what we do but since we don't have those documented, engineers typically find the system theme color based on the hex provided and we make sure they are the ones we want them using.

mdtauk:
> @mdtauk , that's what we do but since we don't have those documented, engineers typically find the system theme color based on the hex provided and we make sure they are the ones we want them using.

So once the Resources have been assigned, would these redlines be updated to document both the resolved colour, and the System/Theme Resource?  Also some of those Theme Resources will use the same colour, but have a different contextual usage, and so an Outlined Button Border, shouldn't use SystemControlDisabledBorder for instance, as the border is used whilst the control is active.

chigy:
>So once the Resources have been assigned, would these redlines be updated to document both the resolved colour, and the System/Theme Resource?

@mdtauk , if we have unlimited resources that'll be how we would like to do, but we are unable to document and maintain such information so that'll have to be queried from the source code. I know that's not a good answer but that's the reality.

We do look at the colors systematically and assign the right color, change the related color if that's good for the system, etc. That's why seemingly small changes take a long time to design and plan.

mdtauk:
Ok, so accepting it is a decision based on speed, if not thoroughness right now @chigy, I see no reason to pursue it further.

As to the values that have been added so far.  Those Outline button styles...
image

The values look as if they have been lifted directly from the Fabric Web implementation, which does not match the colour values you would find in the Windows Theme Resources. (The colours are warmer greys, and not using transparency like the Default Button Fill.

image

**1. Button Text Colour**
**2. Button Fill Colour**
**3. Button Border Colour**

| State | Fill style | Accent style | Outline style | Ghost style |
| :---------- | :------- | :------- | :------- | :------- |
| Rest |1. #FF000000 (100% Black) *no change* <br/><br/>2. #33000000 (20% Black) *no change* |1. #FFFFFFFF (100% White) *no change* <br/><br/>2. 100% Accent color *no change* |1. **#FF000000 (100% Black)** <br/><br/>2. **Transparent** <br/><br/>3. **#66000000 (40% Black)**|1. #FF000000 (100% Black) <br/><br/>2. Transparent |
| Hover |1. #FF000000 (100% Black) *no change* <br/><br/>2. #1A000000 (10% Black) |1. #FFFFFFFF (100% White) *no change* <br/><br/>2. 100% AccentLight1 color |**1. #FF000000 (100% Black)** <br/><br/> **2. #19000000 (10% Black)** <br/><br/>**3. #66000000 (40% Black)**  |1. #FF000000 (100% Black) <br/><br/>2. #1A000000 (10% Black) |
| Pressed |1. #FF000000 (100% Black) *no change* <br/><br/>2. #66000000 (40% Black) *no change* |1. #FFFFFFFF (100% White) *no change* <br/><br/>2. 100% AccentDark1 color  |**1. #FF000000 (100% Black)** <br/><br/>2. **#33000000 (20% Black)** <br/><br/>3.  **#66000000 (40% Black)** |1. #FF000000 (100% Black) <br/><br/>2. #33000000 (20% Black) |
| Disabled |1. #66000000 (40% Black) *no change* <br/><br/>2. #33000000 (20% Black) *no change* |1. #66000000 (40% Black) *no change* <br/><br/>2. #33000000 (20% Black) *no change*  |1. **#66000000 (40% Black)** <br/><br/>2. **#00000000 (Transparent)** <br/><br/> 3. **#66000000 (40% Black)** |1. #66000000 (40% Black) <br/><br/>2. Transparent color |

chigy:
@mdtauk , you are correct. We are lifting the values from Fabric right now because there is no direct customers on Windows/shell that is using outline design today.

Would you be interested in proposing color model matching that to the Window design for Outline version of the style? Or the table you added your proposal?

mdtauk:
> @mdtauk , you are correct. We are lifting the values from Fabric right now because there is no direct customers on Windows/shell that is using outline design today.
> 
> Would you be interested in proposing color model matching that to the Window design for Outline version of the style? Or the table you added your proposal?

@Chigy  I have a proposal about altering the Light and Dark theme colours to make them less stark and more in line with other UI platforms - so the resolved colours would be part of that change.
[Proposal: Improve the default Light and Dark Theme colours #920](https://github.com/microsoft/microsoft-ui-xaml/issues/920)

The table above is specifically about bringing it into line with the existing Windows control styles, using existing Theme Resources - so if the Outline Style is added along with Ghost style - they will be consistent with the XAML controls.

If these redlines are there to guide those building the new control styles - the values used should be as close to the final values as possible - to save work changing them after the design team finalises the design comps.

The 40% Black colour is the same as those used by the Text Box, Combo Box, and other Picker controls - as buttons will sit alongside those controls in a Form very often.

chigy:
> The table above is specifically about bringing it into line with the existing Windows control styles, using existing Theme Resources - so if the Outline Style is added along with Ghost style - they will be consistent with the XAML controls.

@mdtauk , thanks! 
I have a few feedback on your design:

- I think # 3 should be # 4 in your table since # 3 is focus rect. :)
- Why are you making the hover color slightly different from the other styles?
- Shouldn't disabled outline be different color (lighter) than the rest of the states?

mdtauk:
I copied the existing table from the commit, so that may need updating too. But I have edited it now to correct it.

>Why are you making the hover color slightly different from the other styles?
>
The Hover is different in the Ghost style, and the Outline style from Fabric Web is more subtle than the Filled Styles in UWP - so I thought Outline and Ghost should match, and semantically I believe it works.

>Shouldn't disabled outline be different color (lighter) than the rest of the states?
>
I seem to remember with the placeholder text color, there being an issue with contrast ratios.  The disabled Text is a 40% Black value, and so I felt a different value for the border, may not be visible enough - but it could be made 20% Black.  It should be consistent with the Disabled borders for TextBoxes, CheckBoxes, ComboBoxes, and other Picker controls.

At the moment, those TextBoxes and Pickers are filled in when defaulted - but that would make the Outlined Button **more** visible than it is in it's Rest state - which could be a problem.

mdtauk:
Pushing forward with thinking through the Outlined button style.

### Toggle Button

image

Here are some initial thoughts.  I made some changes to the existing Filled and Accent button styles, as well as came up with an idea for the Outlined and Ghost Styles.

- With Accent style being made formal, I felt the Default ToggleButton should probably not use it for its toggled state, to avoid confusion.  So I use an "inverted" approach which gives emphasis.
- Like with the text and picker controls, I gave the Outlined buttons a 2px border, using the accent colour, for their toggled state. (I am not entirely happy with both Default and Accent styles sharing the same Toggled on state design, but apart from adding a border when toggled, I can't immediately think of a better approach)
- I took the idea of the selection indicator from the Pivot control, and used it for the Toggle State on the Ghost style buttons.

chigy:
@mdtauk , thanks for exploring. I personally don't feel Accent styled button makes sense to be toggled... Toggled ON state is basically accented to indicate it is ON. So in a way, ON state is the accent style. Does that make sense? 

There is a design direction team who makes the call around where to use accent color, etc. Since they are not telling us to change the use of accent color, I'd like to stay within the current design system (i.e. toggle on is the accent). This is unless the new proposed design makes it so much better for usability. In this case, it is interesting but I am not sure if it makes it better. It is just different.

mdtauk:
> @mdtauk , thanks for exploring. I personally don't feel Accent styled button makes sense to be toggled... Toggled ON state is basically accented to indicate it is ON. So in a way, ON state is the accent style. Does that make sense?
> 
> There is a design direction team who makes the call around where to use accent color, etc. Since they are not telling us to change the use of accent color, I'd like to stay within the current design system (i.e. toggle on is the accent). This is unless the new proposed design makes it so much better for usability. In this case, it is interesting but I am not sure if it makes it better. It is just different.

So I guess the best thing is to only offer Outlined, Filled, and Ghost style for the ToggleButton control, with Accent only available on the Button?

chigy:
> > @mdtauk , thanks for exploring. I personally don't feel Accent styled button makes sense to be toggled... Toggled ON state is basically accented to indicate it is ON. So in a way, ON state is the accent style. Does that make sense?
> > There is a design direction team who makes the call around where to use accent color, etc. Since they are not telling us to change the use of accent color, I'd like to stay within the current design system (i.e. toggle on is the accent). This is unless the new proposed design makes it so much better for usability. In this case, it is interesting but I am not sure if it makes it better. It is just different.
> 
> So I guess the best thing is to only offer Outlined, Filled, and Ghost style for the ToggleButton control, with Accent only available on the Button?

Yes. I personally feel very confused by the accent/dark button. It really does not communicate clearly which one is on or off... At least we have a mental model in Windows that accent color is associated with the on/selected state.

mdtauk:
Will Buttons use the Semibold font weight like Edgium and Fabric Web, as opposed to the regular weight like UWP XAML and FastDNA. as well as Hyperlinks and Body text?

image
_200% scale_

PR #41, opened 2019-07-30
CornerRadius related APIs

kaiguo:


active/CornerRadius/CornerRadiusFilterConverter.md
```diff
@@ -0,0 +1,37 @@
+# Background
+
+Some controls need to turn on/off part of their rounded corners depending on their states.
+
+e.g. When CommandBarFlyout opens up its secondary menu, primary and secondary menus are clipped together we need to cear rounded corners on bottom of primary menu and top of secondary menu to close the gaps.
```
YuliKl:
```suggestion
e.g. When CommandBarFlyout opens up its secondary menu, primary and secondary menus are clipped together we need to clear rounded corners on bottom of primary menu and top of secondary menu to close the gaps.
```

active/CornerRadius/CornerRadiusFilterConverter.md
```diff
@@ -0,0 +1,37 @@
...
+Some controls need to turn on/off part of their rounded corners depending on their states.
+
+e.g. When CommandBarFlyout opens up its secondary menu, primary and secondary menus are clipped together we need to cear rounded corners on bottom of primary menu and top of secondary menu to close the gaps.
+
+![CommandBarFlyout CornerRadius example](images/commandbarflyout-cornerradius-example.png)
+
+You can't set individual values of a CornerRadius object in markup according to our [doc](https://docs.microsoft.com/en-us/uwp/api/Windows.UI.Xaml.CornerRadius#notes-on-xaml-syntax). We need a way to get non-uniform corner radius values to implement the design above.
+# Description
+
+CornerRadiusFilterConverter helps us filter CornerRadius values on specific dirctions and set values to 0 for the rest.
```
YuliKl:
```suggestion
CornerRadiusFilterConverter helps us filter CornerRadius values on specific directions and set values to 0 for the rest.
```

active/CornerRadius/CornerRadiusFilterConverter.md
```diff
@@ -0,0 +1,37 @@
...
+    {
+        CornerRadiusFilterConverter();
+    };
+}
+```
+
+# Examples
+
+```XAML
+  <Border CornerRadius="{Binding Source={ThemeResource ControlCornerRadius}, Converter={StaticResource CornerRadiusFilterConverter}, ConverterParameter=Top}" />
```
MikeHillberg: Implementing this as a value converter like this is pretty low usability. E.g. no Intellisense for the possible values of ConverterParameter. Were any other solutions investigated? Are the values case sensitive?

active/CornerRadius/CornerRadiusFilterConverter.md
```diff
@@ -0,0 +1,37 @@
...
+```
+
+# Remarks
+
+Currently the ConverterParameter only accepts `Top`, `Bottom`, `Left` and `Right`. We can add more specific directions like `TopLeft` and `TopRight` in the future if there's a need.
```
MikeHillberg: So "Top" means you take both TopLeft and TopRight?

active/CornerRadius/CornerRadiusFilterConverter.md
```diff
@@ -0,0 +1,37 @@
+# Background
+
+Some controls need to turn on/off part of their rounded corners depending on their states.
+
+e.g. When CommandBarFlyout opens up its secondary menu, primary and secondary menus are clipped together we need to clear rounded corners on bottom of primary menu and top of secondary menu to close the gaps.
+
+![CommandBarFlyout CornerRadius example](images/commandbarflyout-cornerradius-example.png)
+
+You can't set individual values of a CornerRadius object in markup according to our [doc](https://docs.microsoft.com/en-us/uwp/api/Windows.UI.Xaml.CornerRadius#notes-on-xaml-syntax). We need a way to get non-uniform corner radius values to implement the design above.
```
MikeHillberg: You can set the components of a CornerRadius, for example:
```
<Border ...>
    <Border.CornerRadius>
        <CornerRadius>4,4,0,0</CornerRadius>
    </Border.CornerRadius>
</Border>

File: active/CornerRadius/CornerRadiusFilterConverter.md (hunk 3 above)
Comment (MikeHillberg) [inline active/CornerRadius/CornerRadiusFilterConverter.md]: Why not just create two new resources: ControlCornerRadiusTop/Bottom?

File: active/CornerRadius/CornerRadiusFilterConverter.md [hunk 6]
```diff
@@ -0,0 +1,37 @@
...
+Some controls need to turn on/off part of their rounded corners depending on their states.
+
+e.g. When CommandBarFlyout opens up its secondary menu, primary and secondary menus are clipped together we need to clear rounded corners on bottom of primary menu and top of secondary menu to close the gaps.
+
+![CommandBarFlyout CornerRadius example](images/commandbarflyout-cornerradius-example.png)
+
+You can't set individual values of a CornerRadius object in markup according to our [doc](https://docs.microsoft.com/en-us/uwp/api/Windows.UI.Xaml.CornerRadius#notes-on-xaml-syntax). We need a way to get non-uniform corner radius values to implement the design above.
+# Description
+
+CornerRadiusFilterConverter helps us filter CornerRadius values on specific directions and set values to 0 for the rest.
```
MikeHillberg: Precedent for this kind of problem is CommandBarTemplateSettings. Why aren't se using that here?

active/CornerRadius/CornerRadiusFilterConverter.md
kaiguo: Yeah that's correct.

active/CornerRadius/CornerRadiusFilterConverter.md
kaiguo: Yes, but what we need is a way to extract values from uniform CornerRadius. The problem we were facing was, if user passes in CornerRadius="x", how do we get non-uniform values like x,x,0,0 just using markup? 

Take FlipView as example, given `<FlipView CornerRadius="4"/>`, we need to set 0,4,4,0 for horizontal previous button and 4,0,0,4 for horizontal next button.
{81A9E3CE-45F5-43CC-903E-937AB93A4090}

If setting individual values is possible we can probably do 
```XAML
<CornerRadius x:Key="HorizontalPreviousCornerRadius"
    TopLeft="0"
    TopRight="{Binding RelativeSource={RelativeSource TemplatedParent}, 
    Path=CornerRadius.TopLeft}"
    BottomRight="{Binding RelativeSource={RelativeSource TemplatedParent}, 
    Path=CornerRadius.BottomRight}"
    TopRight="0"/>
```
But looks like that's not supported so we came up with this converter idea.

active/CornerRadius/CornerRadiusFilterConverter.md
kaiguo: We are trying to automatically extract correct non-uniform values from the value that user passes in, so we can't just hard code it.

If they set CornerRadius="x" on CommandBarFlyout, when secondary menu is not shown, primary menu CornerRadius is x,x,x,x, but when secondary menu opens up, primary menu CornerRadius becomes x,x,0,0.

active/CornerRadius/CornerRadiusFilterConverter.md
kaiguo: Yes, I think TemplateSettings works for CommandBarFlyout. There are other controls like AutoSuggestBox and ComboBox with similar problems but their code is not in MUX. We are trying to solve the problem without touching code in OS.

active/CornerRadius/CornerRadiusFilterConverter.md
MikeHillberg: I see, so the motivating use case is in control templates. Does that prevent usage of a method binding?
```
CornerRadius="{x:Bind FilterCorner(CornerRadius, FilterCornerLocation.Top)}"
```

mdtauk:
Quick question.  Does the value animate as it changes, or does it just switch?

So when a Flyout is shown below, does the bottom corner radii switch to 0, or does it transition between 2 and 0?

active/CornerRadius/CornerRadiusFilterConverter.md
kaiguo: The problem with that is the generated code has no place to live since generic.xaml has no code-behind. I tried x:bind first and it didn't compile :\

We had 3 options just like you mentioned in the review, x:bind to a function, TemplateSettings and converter. Converter was the last thing we wanted to do but with all the restrictions seems like that's the only option that works.

kaiguo:
> Quick question. Does the value animate as it changes, or does it just switch?
> 
> So when a Flyout is shown below, does the bottom corner radii switch to 0, or does it transition between 2 and 0?

No animations for now but I guess we can discuss with @chigy.  I personally think it actually looks better without animations since the secondary menu slides down when you open it (Same for AutoSuggestBox and ComboBox). The top part of secondary menu is almost immediately visible after clicking the button, if we animate the corner radii you'll see a gap at first and then it slowly closes (which i feel draws too much unnecessary attention?).

chigy:
> > Quick question. Does the value animate as it changes, or does it just switch?
> > So when a Flyout is shown below, does the bottom corner radii switch to 0, or does it transition between 2 and 0?
> 
> No animations for now but I guess we can discuss with @chigy. I personally think it actually looks better without animations since the secondary menu slides down when you open it (Same for AutoSuggestBox and ComboBox). The top part of secondary menu is almost immediately visible after clicking the button, if we animate the corner radii you'll see a gap at first and then it slowly closes (which i feel draws too much unnecessary attention?).

@kaiguo , @mdtauk ,
Kai is right. We want to be thoughtful and purposeful about adding motion. In this case, we decided not to add motion because we didn't see the value. That said, if that proves to be wrong, we can always add it later. For v1, we are not.

mdtauk:
> > > Quick question. Does the value animate as it changes, or does it just switch?
> > > So when a Flyout is shown below, does the bottom corner radii switch to 0, or does it transition between 2 and 0?
> > 
> > 
> > No animations for now but I guess we can discuss with @chigy. I personally think it actually looks better without animations since the secondary menu slides down when you open it (Same for AutoSuggestBox and ComboBox). The top part of secondary menu is almost immediately visible after clicking the button, if we animate the corner radii you'll see a gap at first and then it slowly closes (which i feel draws too much unnecessary attention?).
> 
> @kaiguo , @mdtauk ,
> Kai is right. We want to be thoughtful and purposeful about adding motion. In this case, we decided not to add motion because we didn't see the value. That said, if that proves to be wrong, we can always add it later. For v1, we are not.

@kaiguo @chigy  Thank you.  I only asked because there is a show transition for the drop down/flyout, and so the corner radius change would happen a tiny little bit before the flyout has finished appearing.

chigy:
> > > > Quick question. Does the value animate as it changes, or does it just switch?
> > > > So when a Flyout is shown below, does the bottom corner radii switch to 0, or does it transition between 2 and 0?
> > > 
> > > 
> > > No animations for now but I guess we can discuss with @chigy. I personally think it actually looks better without animations since the secondary menu slides down when you open it (Same for AutoSuggestBox and ComboBox). The top part of secondary menu is almost immediately visible after clicking the button, if we animate the corner radii you'll see a gap at first and then it slowly closes (which i feel draws too much unnecessary attention?).
> > 
> > 
> > @kaiguo , @mdtauk ,
> > Kai is right. We want to be thoughtful and purposeful about adding motion. In this case, we decided not to add motion because we didn't see the value. That said, if that proves to be wrong, we can always add it later. For v1, we are not.
> 
> @kaiguo @chigy Thank you. I only asked because there is a show transition for the drop down/flyout, and so the corner radius change would happen a tiny little bit before the flyout has finished appearing.

@kaiguo , let's make sure we take a look at the motion together once this is working on your machine.

active/CornerRadius/CornerRadiusFilterConverter.md
MikeHillberg: Untyped properties can be frustrating (no Intellisense). How about making this a property on the converter
```
<Grid.Resources>
    <CornerRadiusFilter Filter="Top" x:Name="CornerFilter" />
</Grid.Resources>
...
CornerRadius="{TemplateBinding CornerRadius, Converter={StaticResource CornerFilter}}"
...
enum CornerRadiusFilterKind
{
    Top,
    Right,
    Left,
    Bottom
}
```

kaiguo: That means we'll have to create a filter for each kind, like TopCornerFilter and RightCornerFilter?

active/CornerRadius/CornerRadiusFilterConverter.md
```diff
@@ -0,0 +1,63 @@
...
+
+# Examples
+A ControlTemplate for a Button that has square bottom corners, and top corners that can be set by the Button.CornerRadius property.
+```XAML
+<ControlTemplate TargetType="Button">
+    <Border Background="{TemplateBinding Background}"
+            CornerRadius="{TemplateBinding CornerRadius, Converter={StaticResource CornerRadiusFilterConverter}, ConverterParameter=Top}"
```
MikeHillberg: The bad thing about using ConverterParameter is that it's untyped, so you don't know what's valid (no Intellisense support or compile-time checking). Better would be to make the parameter a property on the converter class. So:

```
<Grid.Resources>
    <CornerRadiusFilterConverter Filter="Top" x:Name="TopCornersConverter" />
</Grid.Resources>
...
CornerRadius="{TemplateBinding CornerRadius, Converter={StaticResource TopCornersConverter}}"
...
enum CornerRadiusFilterKind
{
    Top,
    Right,
    Left,
    Bottom
}
```

jevansaks: @MikeHillberg I like this suggestion, but what would you recommend for the key names that we put in the resource dictionary? TopCornerRadiusFilterConverter, LeftCornerRadiusFilterConverter, etc? "TopCornersConverter" seems a little too vague for a global key.

active/CornerRadius/CornerRadiusFilterConverter.md
MikeHillberg: You're right. But I think that's worth it for the Intellisense guidance and compile-time validation.

active/CornerRadius/CornerRadiusFilterConverter.md
MikeHillberg: This is going to be a global resource (generic.xaml)?

jevansaks: Yes, since it would need to be used by default styles of controls.

active/CornerRadius/CornerRadiusFilterConverter.md
```diff
@@ -0,0 +1,82 @@
...
+    </Grid>
+</ControlTemplate>
+```
+
+# API Details
+
+```
+namespace Microsoft.UI.Xaml.Controls.Primitives
+{
+    [bindable]
```
MikeHillberg: Why do we need to set [bindable]?

kaiguo: Yeah we probably don't need it, will remove.

PR #43, opened 2019-07-31
TabView API Spec

stmoy:
Spec review for TabView. 

(See #28 for full history/comments)


active/tab/tabcontrol.md
```diff
@@ -0,0 +1,580 @@
...
+* Limited keyboard and accessibility support
+
+This spec therefore covers "document-style tabs" specifically as this platform gap isn't easily solved with an app-side workaround. 
+
+# Description
+<!-- Use this section to provide a brief description of the feature.
+For an example, see the introduction to the PasswordBox control 
+(http://docs.microsoft.com/windows/uwp/design/controls-and-patterns/password-box). -->
+
+TabView is a collection of tabs that each represents a new page or document in your app. Use TabView when your app has several pages of content and the user expects to be able to add, close, and rearrange the pages.
```
oldnewthing: > to be able to **switch between pages,** as well as add, close and rearrange the pages.

We sort of buried the lede: Switching between pages is what TabView spends most of its time doing!

active/tab/tabcontrol.md
```diff
@@ -0,0 +1,580 @@
...
+  feature explanation,
+  example code
+  feature explanation,
+  example code
+  etc.-->
+
+<!-- As an example of this section, see the Examples section for the PasswordBox control 
+(https://docs.microsoft.com/windows/uwp/design/controls-and-patterns/password-box#examples). -->
+
+## To replicate the behavior of Microsoft Edge
```
oldnewthing: Don't oversell this. It's not a perfect replica (doesn't support needy state, for example). Perhaps

> To get behavior typical of a Web browser

active/tab/tabcontrol.md
mdtauk: Do Tabs host pages?  If not, it may be an idea not to confuse these content panels with actual pages or frames?

active/tab/tabcontrol.md
```diff
@@ -0,0 +1,580 @@
...
+<Page>
+    <Grid>
+        <TabView>
+            <TabViewItem Icon="Home" Header="Home" IsCloseable="False" />
+            <TabViewItem Icon="Document" Header="Document 1" />
+            <TabViewItem Icon="Document" Header="Document 2" />
+            <TabViewItem Icon="Document" Header="Document 3" />
+        </TabView>
+
+        <Grid x:Name="CustomDragRegion" Width="200" Height="40" HorizontalAlignment="Right" VerticalAlignment="Top" />
```
oldnewthing: Presumably this drag region should be the `TabView.Footer`, so that tabs won't go into the drag region. Ideally, should use system metrics to obtain size of drag region, so we don't overlap system buttons.

How do we make things so that all the unused space in the TabView becomes drag region?

active/tab/tabcontrol.md
```diff
@@ -0,0 +1,580 @@
...
+See the [TabView tear out sample](https://github.com/windows-toolkit/Sample-TabView-TearOff/tree/master/TabViewTear) for a more complete sample.
+
+``` xml
+<TabView CanDragTabs="True"
+            CanReorderTabs="True"
+            TabDraggedOutside="TabView_TabDraggedOutside">
+```
+``` csharp
+// NOTE: The app is responsible for writing this code. We will provide a sample that may look something like:
+private async void TabView_TabDraggedOutside(object sender, TabDraggedOutsideEventArgs e)
```
oldnewthing: Name suggests (but hard to confirm in code) that this occurs as soon as the tab is "ripped out" of the TabView. The tab is still stuck to the pointer because the user hasn't dropped the tab anywhere yet. How does the app make sure that the correct point of the new window is under the active pointer?

Alternate name: TabDraggedOut?

active/tab/tabcontrol.md
```diff
@@ -0,0 +1,580 @@
...
+    event Windows.Foundation.TypedEventHandler<TabView, TabDraggedOutsideEventArgs> TabDraggedOutside;
+
+    // TODO: Update
+    // From ListView
+    [MUX_PROPERTY_CHANGED_CALLBACK(TRUE)]
+    Object TabItemsSource;
+
+    // TODO: Update
+    [MUX_PROPERTY_CHANGED_CALLBACK(TRUE)]
+    Windows.Foundation.Collections.IVector<Object> TabItems{ get; };
```
oldnewthing: Why `IVector<Object>` instead of `IVector<TabViewItem>`? What happens if I put something that isn't derived from `TabViewItem` into this vector?

active/tab/tabcontrol.md
```diff
@@ -0,0 +1,580 @@
...
+```
+
+## Databind to a set of tabs 
+
+``` xml
+<TabView TabItemsSource="{x:Bind TabViewItemCollection}" />
+```
+
+## Add a new tab using the Add (+) Tab Button
+The TabView control provides an Add (+) Tab button at the end of the tab strip. When the user clicks the Add Tab button, the control will fire an event which you can listen to in order to add a new tab.
```
oldnewthing: s/fire/raise/

active/tab/tabcontrol.md
```diff
@@ -0,0 +1,580 @@
...
+```
+
+# Appendix
+<!-- Anything else that you want to write down for posterity, but 
+that isn't necessary to understand the purpose and usage of the API.
+For example, implementation details. -->
+
+![Position of TabStripHeader and TabStripFooter](images/tab-control-parts-1.PNG)
+
+![Parts of a tab item](images/tab-control-parts-2.PNG)
```
oldnewthing: These diagrams are so important I think they need to be moved out of the Appendix to main body. In particular, the breakdown of TabViewItem.Header / TabViewItem.Content is essential for understanding the `CreateNewTabFromDocument` sample function.

active/tab/tabcontrol.md
```diff
@@ -0,0 +1,580 @@
...
+    doc.Title = "New Document";
+    doc.Content = GenerateBaconIpsum();
+    doc.Icon = new SymbolIcon(Symbol.Document);
+
+    // Create a new TabViewItem from the MyDocument object and add it to the TabItems collection
+    TabRoot.TabItems.Add(CreateNewTabFromDocument(doc));
+}
+
+private TabViewItem CreateNewTabFromDocument(MyDocument doc)
+{
```
oldnewthing: Somewhat unrealistic in that there is nothing that lets us get from the `MyDocument` to the `TabViewItem` or vice versa. Should we do a `newItem.Tag = doc;`?

active/tab/tabcontrol.md
```diff
@@ -0,0 +1,580 @@
...
+    Equal = 1,
+};
+
+[WUXC_VERSION_PREVIEW]
+[webhosthidden]
+runtimeclass TabViewTabClosingEventArgs
+{
+    Object Item { get; };
+
+    Boolean Cancel { get; set; };
```
oldnewthing: If I don't cancel, does the TabView automatically remove the tab and destroy it?

What if I need to prompt "Save changes?" Do we need a deferral?

Maybe we don't need a Cancel at all. This is just a `TabViewTabCloseRequestedEventArgs`, and it's up to the app to deal with the close request. Maybe it removes the tab immediately. Maybe it prompts before removing the tab. Does the system need to know when the close request is complete?

active/tab/tabcontrol.md
```diff
@@ -0,0 +1,580 @@
...
+    // TODO: Update
+    // From ListView
+    [MUX_PROPERTY_CHANGED_CALLBACK(TRUE)]
+    Object TabItemsSource;
+
+    // TODO: Update
+    [MUX_PROPERTY_CHANGED_CALLBACK(TRUE)]
+    Windows.Foundation.Collections.IVector<Object> TabItems{ get; };
+
+    Windows.UI.Xaml.DataTemplate ItemTemplate;
```
oldnewthing: What does this template apply to? Since a TabViewItem is a ContentControl, does the template apply to the Content? Or to the Header? Or both? (How do you even do both?) If this templates the header, then should it be called the ItemHeaderTemplate?

active/tab/tabcontrol.md
```diff
@@ -0,0 +1,580 @@
...
+
+[WUXC_VERSION_PREVIEW]
+[webhosthidden]
+unsealed runtimeclass TabView : Windows.UI.Xaml.Controls.Control
+{
+    TabView();
+
+    [MUX_DEFAULT_VALUE("winrt::TabViewItemWidthMode::Actual")]
+    [MUX_PROPERTY_CHANGED_CALLBACK(TRUE)]
+    TabViewItemWidthMode TabViewItemWidthMode{ get; set; };
```
oldnewthing: Is `ItemWidthMode` sufficient?

active/tab/tabcontrol.md
```diff
@@ -0,0 +1,580 @@
...
+
+    // TODO: Remove
```
oldnewthing: Are these TODOs still relevant?

active/tab/tabcontrol.md
```diff
@@ -0,0 +1,580 @@
...
+|---|---|
+| DependencyObject ContainerFromItem(object item) | Returns the generated container for the specified item. |
+| DependencyObject ContainerFromIndex(Int32 index) | Returns the generated container for the specified index. |
+
+### TabViewItem properties and events
+
+| Property | Description |
+|:-------- |:----------- |
+| Content | The main content that appears in the tab. |
+| Header | The content that appears inside the tab itself.  |
```
oldnewthing: The descriptions for `Content` and `Header` are nearly identical.

active/tab/tabcontrol.md
```diff
@@ -0,0 +1,580 @@
...
+```
+namespace MU_XC_NAMESPACE
+{
+
+[WUXC_VERSION_PREVIEW]
+[webhosthidden]
+enum TabViewItemWidthMode
+{
+    Actual = 0,
+    Equal = 1,
```
oldnewthing: What happens if tab strip overflows? My guess:

* Actual: You get horizontal scroll bars.  Or maybe they clip and autoscroll?
* Equal: The headers just get narrower and narrower. At the extreme, they could end up just 2 pixels wide each.

active/tab/tabcontrol.md
```diff
@@ -0,0 +1,580 @@
...
+};
+
+[WUXC_VERSION_PREVIEW]
+[webhosthidden]
+runtimeclass TabViewTabClosingEventArgs
```
oldnewthing: Event is called TabClosing but the property is called Item.

active/tab/tabcontrol.md
```diff
@@ -0,0 +1,580 @@
...
+    [MUX_DEFAULT_VALUE("winrt::TabViewItemWidthMode::Actual")]
+    [MUX_PROPERTY_CHANGED_CALLBACK(TRUE)]
+    TabViewItemWidthMode TabViewItemWidthMode{ get; set; };
+
+    // TODO: Remove
+    [MUX_DEFAULT_VALUE("true")]
+    Boolean CanCloseTabs{ get; set; };
+
+    [MUX_DEFAULT_VALUE("true")]
+    Boolean CanDragTabs{ get; set; };
```
oldnewthing: I think the matrix for the interaction between CanDragTabs and CanReorderTabs is like this:

|| CanDragTabs = true | CanDragTabs = false |
|-|-|-|
| CanReorderTabs = true | When user grabs a tab, it starts dragging.<br>If user drags left/right, then reordering happens.<br>If user drags up/down, then drag-out happens. | When user grabs a tab, it starts dragging.<br>If user drags left/right, then reordering happens.<br>Up/down motion is ignored.|
| CanReorderTabs = false | When user grabs a tab, it starts dragging.<br>Left/right motion is ignored when inside the tabstrip.<br>When user drags up/down, then drag-out happens. | When user grabs a tab, nothing happens. |

The name CanDragTabs is misleading in the case of the upper right corner, because you can drag them - you just can't drag them out.

Should we call it CanDragTabsOut?

active/tab/tabcontrol.md
```diff
@@ -0,0 +1,580 @@
...
+    Tabs.TabItems.Remove(e.Tab);
+
+    // Add tab to list of Tabs on new page
+    newPage.AddItemToTabs(e.Tab);
+
+    // Set the Window's content to the new page
+    ElementCompositionPreview.SetAppWindowContent(newWindow, newPage);
+
+    // Show the window
+    await newWindow.TryShowAsync();
```
oldnewthing: After the tab has been dragged out (but not yet dropped), say the user moves back into the original tabstrip. What happens? There is no TabDraggedIn message.

active/tab/tabcontrol.md
```diff
@@ -0,0 +1,580 @@
...
+    var newPage = new MainPage();
+
+    // Remove tab from existing list
+    Tabs.TabItems.Remove(e.Tab);
+
+    // Add tab to list of Tabs on new page
+    newPage.AddItemToTabs(e.Tab);
+
+    // Set the Window's content to the new page
+    ElementCompositionPreview.SetAppWindowContent(newWindow, newPage);
```
oldnewthing: After the tab has been dragged out (but not yet dropped), how can the app cancel the drag (and return the tab to its original location) when the user presses ESC?

active/tab/tabcontrol.md
```diff
@@ -0,0 +1,580 @@
...
+
+## Create a new window when tearing out tabs
+
+![An example picture showing that TabView can be torn out and moved to its own window.](images/tab-new-window.png)
+
+See the [TabView tear out sample](https://github.com/windows-toolkit/Sample-TabView-TearOff/tree/master/TabViewTear) for a more complete sample.
+
+``` xml
+<TabView CanDragTabs="True"
+            CanReorderTabs="True"
```
oldnewthing: There is no TabReordered event. How does the app know that the user reordered the tabs?

oldnewthing: How can the control reorder the tabs if the tabs are provided by an ItemsSource?

oldnewthing: If the app wants to reorder the tabs themselves (e.g. "Sort tabs by name"), what's the best way to do that? There is no IVector<T>.MoveTo method. Do they just have to delete and re-add? Will this result in confusion (e.g., a flurry of SelectionChanged messages)?

active/tab/tabcontrol.md
```diff
@@ -0,0 +1,580 @@
...
+  etc.-->
+
+<!-- As an example of this section, see the Examples section for the PasswordBox control 
+(https://docs.microsoft.com/windows/uwp/design/controls-and-patterns/password-box#examples). -->
+
+## To replicate the behavior of Microsoft Edge
+
+``` xml
+<TabView TabViewItemWidthMode="Equal"
+            CanCloseTabs="True"
```
oldnewthing: What happens if CanCloseTabs and TabViewItem.IsClosable conflict? Both are Booleans, neither is nullable.

active/tab/tabcontrol.md
```diff
@@ -0,0 +1,580 @@
...
+            CloseButtonOverlay="OnHover"
```
oldnewthing: This property is no longer in the IDL.

active/tab/tabcontrol.md
```diff
@@ -0,0 +1,580 @@
...
+## Create a new window when tearing out tabs
+
+![An example picture showing that TabView can be torn out and moved to its own window.](images/tab-new-window.png)
+
+See the [TabView tear out sample](https://github.com/windows-toolkit/Sample-TabView-TearOff/tree/master/TabViewTear) for a more complete sample.
+
+``` xml
+<TabView CanDragTabs="True"
+            CanReorderTabs="True"
+            TabDraggedOutside="TabView_TabDraggedOutside">
```
oldnewthing: What happens if there is only one tab? Can you drag out the sole remaining tab? (Why not?)

Presumably if an app doesn't want the last tab to be dragged out, they would set CanDragTabs = false when the tab count is 1, and also update their drag region so the tab is part of the window drag region.

active/tab/tabcontrol.md
```diff
@@ -0,0 +1,580 @@
...
+runtimeclass TabViewTabClosingEventArgs
+{
+    Object Item { get; };
+
+    Boolean Cancel { get; set; };
+}
+
+[WUXC_VERSION_PREVIEW]
+[webhosthidden]
+unsealed runtimeclass TabView : Windows.UI.Xaml.Controls.Control
```
MikeHillberg: Proposal: add drag/drop APIs for dragging the items. Match ListView:

    CanDragItems
    DragItemsCompleted
    DragItemsStarting
    IsDragSource

active/tab/tabcontrol.md
oldnewthing: I was using "pages" in a casual sense. Would "switch between content" be better?

mdtauk: Content sounds more accurate to me

active/tab/tabcontrol.md
oldnewthing: No definition for `TabDraggedOutsideEventArgs`. Assuming it's

```cs
runtimeclass TabDraggedOutsideEventArgs
{
    TabViewItem Tab { get; };
}
```

oldnewthing: This is a TypedEventHandler, so `sender` should be `TabView`.

active/tab/tabcontrol.md
```diff
@@ -0,0 +1,580 @@
...
+
+    // Add tab to list of Tabs on new page
+    newPage.AddItemToTabs(e.Tab);
+
+    // Set the Window's content to the new page
+    ElementCompositionPreview.SetAppWindowContent(newWindow, newPage);
+
+    // Show the window
+    await newWindow.TryShowAsync();
+}
```
oldnewthing: What about TabDroppedInside?

active/tab/tabcontrol.md
```diff
@@ -0,0 +1,580 @@
...
+<!-- Give a one or two line description of each API (type
+and member), or at least the ones that aren't obvious
+from their name.  These descriptions are what show up
+in IntelliSense. -->
+
+### TabView properties, events, methods
+
+| Property | Description |
+|:-------- |:----------- |
+| AddTabButtonCommand | Gets or sets the command to invoke when the Add button is tapped. |
```
oldnewthing: Why have both this and `AddTabButtonClick`?

active/tab/tabcontrol.md
stmoy: You're correct, this is a sample bug. The custom drag region should be contained within TabView.TabStripFooter with alignment = stretch to fill the whole region. 

> Ideally, should use system metrics to obtain size of drag region, so we don't overlap system buttons.

I haven't found a good way to do this beyond setting a hardcoded min-size to the custom drag region. Is there a better way?

active/tab/tabcontrol.md
stmoy: These TODOs generally mark deltas between what we have implemented and the "end goal". In this case, we have an API in code that we should remove since it's not desirable for the final control.

active/tab/tabcontrol.md
stmoy: Agreed, and I'm open to suggestions. I wasn't sure how to succinctly say "the stuff in the Header goes into the literal UI of the tab whereas the content goes into the body below the tab".

active/tab/tabcontrol.md
stmoy: Actual: almost correct. We horizontally scroll this area but hide the scrollbars and introduce bumpers (similar to UWP Edge)
Equal: almost correct. They get narrower and narrower to a min width point, and then you get horizontal scrolling/bumpers.

active/tab/tabcontrol.md
stmoy: The intent is to move CanCloseTabs, partially for this reason.

active/tab/tabcontrol.md
```diff
@@ -0,0 +1,580 @@
...
+            CanReorderTabs="True"
+            TabDraggedOutside="OpenTabInNewWindow"
+            AddTabButtonClick="NewTab_Click">
+    ...
+</TabView>
+```
+
+## Put TabView in the titlebar area
+
+![A sample app that shows how TabView can extend into the titlebar area](images/tab-extend-to-title.png)
```
oldnewthing: Note that web browsers put an address bar between the tabs and the content. Can TabView do this? It looks like the tab bar goes directly above the content. You don't get to pick where the tab bar goes relative to the content.

active/tab/tabcontrol.md
stmoy: I think this might be a misnomer. TabDraggedOutside is raised when the tab is _dropped_ outside of the window. 

> How does the app make sure that the correct point of the new window is under the active pointer?

Good question, and this is a hard problem that I don't believe we have a great solution for. I think we'll need a more complete sample that uses the positioner features of AppWindow?

active/tab/tabcontrol.md
oldnewthing: You can find an example in the [TitleBar sample](https://github.com/microsoft/Windows-universal-samples/tree/master/Samples/TitleBar).

* [XAML](https://github.com/microsoft/Windows-universal-samples/blob/master/Samples/TitleBar/shared/CustomTitleBar.xaml)
* [C# code-behind](https://github.com/microsoft/Windows-universal-samples/blob/master/Samples/TitleBar/cs/CustomTitleBar.xaml.cs)

active/tab/tabcontrol.md
stmoy: Replied to an earlier message. I think TabDraggedOutside should maybe be called TabDroppedInside, In this case (where it is dragged out, not dropped, and moved back in) the event wouldn't fire at all.

active/tab/tabcontrol.md
oldnewthing: How about something like

| Property | Description |
|:-|:-|
| Content | The content that appears in the TabView's content area when the TabViewItem is selected. |
| Header | The content that appears in the TabView's tab row to represent the tab. |

active/tab/tabcontrol.md
```diff
@@ -0,0 +1,580 @@
...
+
+    Windows.UI.Xaml.DataTemplate ItemTemplate;
+    Windows.UI.Xaml.Controls.DataTemplateSelector ItemTemplateSelector{ get; set; };
+
+    [MUX_DEFAULT_VALUE("-1")]
+    [MUX_PROPERTY_CHANGED_CALLBACK(TRUE)]
+    Int32 SelectedIndex;
+
+    [MUX_PROPERTY_CHANGED_CALLBACK(TRUE)]
+    Object SelectedItem;
```
oldnewthing: What happens if `SelectedItem == null` (aka `SelectedIndex == -1`)?

active/tab/tabcontrol.md
```diff
@@ -0,0 +1,580 @@
...
+### Selecting a tab
+When a TabViewItem has focus, pressing SPACE or ENTER will select that TabViewItem.
+
+_Use ARROW to move focus, ENTER to select tab_
+
+![Use ARROW to move focus, ENTER to select tab](images/tab-control-keyboarding-2.PNG)
+
+### Shortcuts for selecting adjacent tabs
+Ctrl+TAB will select the next TabViewItem.
+Ctrl+Shift+TAB will select the previous TabViewItem.
```
oldnewthing: When are the <kbd>Ctrl</kbd>+<kbd>TAB</kbd> and <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>TAB</kbd> shortcuts active?

* When focus is on the TabStrip.
* When focus is anywhere in the TabView.
* When focus is anywhere in the window. (What if there is more than one TabView in the window?)

If the app has controls outside the TabView (e.g., status bar at bottom), and it wants <kbd>Ctrl</kbd>+<kbd>TAB</kbd> to change tabs, then it presumably needs to install its own global shortcut. Will that interact adversely with the built-in shortcut?

active/tab/tabcontrol.md
```diff
@@ -0,0 +1,580 @@
...
+TAB selection cycles (meaning if the user has selected the last tab and presses Ctrl+TAB, the first tab becomes selected).
+
+### Closing a tab 
+Hitting CTRL+F4 will close the selected TabViewItem (assuming it is closable).
```
oldnewthing: Same question: When is this hotkey active?

active/tab/tabcontrol.md
MikeHillberg: Do we have guidelines the minimum size of the drag region?

active/tab/tabcontrol.md
MikeHillberg: Discussion: note that the dropped outside event requires the app to remove the tab item, but this event does not.

Recommend: Remove the Cancel property. Don't actually remove the tab item from the control; leave it for the app.

active/tab/tabcontrol.md
MikeHillberg: Recommend: Matching ListView is probably OK, but per the earlier discussion around reorder, get a more thorough explanation of ListView's behavior and consider matching it.

active/tab/tabcontrol.md
MikeHillberg: Discussion: This applies to the TabViewItem.Header + extras like the icon.

Recommend: Rename property to TabView.TabWidthMode.
Enum named TabViewTabWidthMode.

active/tab/tabcontrol.md
MikeHillberg: Recommend: remove TODOs from the spec

active/tab/tabcontrol.md
MikeHillberg: Recommend: Align these two properties with the same two properties on ListView

active/tab/tabcontrol.md
MikeHillberg: Recommend: Change type to `IVector<TabViewItem>`

active/tab/tabcontrol.md
MikeHillberg: Discussion: Maps TabViewItem.ContentTemplate

Recommend: Rename to 
TabView.TabContentTemplate
~TabView.TabItemContentTemplate~
~TabView.TabViewItemContentTemplate~

Also rename ItemTemplateSelector to TabContentTemplateSelector

Recommend: Don't we need a header template too?

active/tab/tabcontrol.md
MikeHillberg: Discussion: Behaves like ListView, it's OK to have nothing selected. (No tab is selected and the content area is empty)

active/tab/tabcontrol.md
MikeHillberg: Consider: Move this up

active/tab/tabcontrol.md
MikeHillberg: Discussion: When focus is anywhere inside the tab view (including the tab strip).

active/tab/tabcontrol.md
MikeHillberg: Discussion: Yes, when focus is inside the tab view.

active/tab/tabcontrol.md
stmoy: Thoughts on where this note should go? I was thinking in the TabWidthMode enum doc, but I'm not sure how discoverable that is.

active/tab/tabcontrol.md
stmoy: Should we rename TabClosing to something else (like CloseTabRequested)?

MikeHillberg: Yes, definitely.

active/tab/tabcontrol.md
MikeHillberg: I think it should be in the Remarks section. (There's only one Remarks section in this spec template, but ultimately it should go into the TabView class page's Remarks section.)

mdtauk: The minimum width should prioritise the display of the close button within the tab, if tabs are closable.

active/tab/tabcontrol.md
stmoy: For some reason, the overlay inset is always returning 0 🤔  Following up offline.

active/tab/tabcontrol.md
stmoy: This will be handled by the Drop events mentioned above.

active/tab/tabcontrol.md
stmoy: FYI: We added TabViewTabDroppedOutsideEventArgs (using the Tab instead of Item nomenclature) since its an event on TabView.

active/tab/tabcontrol.md
stmoy: @teaP - can you comment on this?

active/tab/tabcontrol.md
stmoy: For these, we matched ListView but put a TabView spin on the names. We have:
* CanDragTabs
* CanReorderTabs
* AllowDropTabs 
* TabDragStarting
* TabDragCompleted
* TabStripDragOver
* TabStripDrop

I have verified that these APIs allow us to build the basic drag-and-drop scenarios that ListView also supports so I believe this is everything we need.

active/tab/tabcontrol.md
stmoy: Added a note.

active/tab/tabcontrol.md
stmoy: Clarified the sample.

active/tab/tabcontrol.md
```diff
@@ -189,34 +189,16 @@ The TabView control provides an Add (+) Tab button at the end of the tab strip.
...
-
-    newItem.Header = doc.Title;
-    newItem.Icon = doc.Icon;
+	var newTab = new TabViewItem();
+	newTab.IconSource = new SymbolIconSource() { Symbol = Symbol.Document };
+	newTab.Header = "New Document";
 
-    // The Content of a TabViewItem is often a frame which hosts a page.
+	// The Content of a TabViewItem is often a frame which hosts a page.
     Frame frame = new Frame();
```
ExE-Boss:
```suggestion
	Frame frame = new Frame();
```

active/tab/tabcontrol.md
```diff
@@ -189,34 +189,16 @@ The TabView control provides an Add (+) Tab button at the end of the tab strip.
...
-    newItem.Content = frame;
-    frame.Navigate(typeof(DocumentPage), doc);
+    newTab.Content = frame;
```
ExE-Boss:
```suggestion
	newTab.Content = frame;
```

active/tab/tabcontrol.md
```diff
@@ -189,34 +189,16 @@ The TabView control provides an Add (+) Tab button at the end of the tab strip.
...
+    frame.Navigate(typeof(BaconIpsumPage));
```
ExE-Boss:
```suggestion
	frame.Navigate(typeof(BaconIpsumPage));
```

stmoy:
> Since you’ve mostly switched to tabs in [133604b](https://github.com/microsoft/microsoft-ui-xaml-specs/commit/133604b470d63363d49fbc6cf27424322f5bfa3a) (🎉), you should finish the conversion fully, and not leave examples half‑converted.

@ExE-Boss - this was actually not my intent as we generally use spaces 😅 - Agreed that we don't want to have a mix of both though, thanks for the call out!

active/tab/tabcontrol.md
stmoy: When the ItemsSource is set, use the TabItemTemplate to specify properties on the TabViewItem like Header and IconSource. We don’t have separate ContentTemplate and HeaderTemplate, as both can be satisfied by the following syntax:

```
<controls:TabView x:Name="DataBindingTabView" Grid.Row="1">
    <controls:TabView.TabItemTemplate >
        <DataTemplate x:DataType="local:MyData">
            <controls:TabViewItem Header="{x:Bind Header}" IconSource="{x:Bind IconSource}" Content="{x:Bind MyContent}">/>
        </DataTemplate>
    </controls:TabView.TabItemTemplate >
</controls:TabView>
```

PR #49, opened 2019-10-18
Updated and Shortened Grid Syntax

anawishnoff:
Proposal for this feature request: https://github.com/microsoft/microsoft-ui-xaml/issues/673.


anawishnoff:
Hi @marb2000, I actually opened this PR for reviewing and commenting from the dev team, but thanks so much for the quick response to my work! Going to leave this open for now :)

active/gridsyntax/GridSyntaxSpec.md
```diff
@@ -0,0 +1,324 @@
...
+area, just explanation enough to understand this new API, rather than telling
+the reader "go read 100 pages of background information posted at ...". -->
+
+
+# Description
+<!-- Use this section to provide a brief description of the feature.
+For an example, see the introduction to the PasswordBox control 
+(http://docs.microsoft.com/windows/uwp/design/controls-and-patterns/password-box). -->
+A new XAML language feature that allows read-only collection-type properties to be created by assigning the CreateFromString attribute. The CreateFromString attribute will allow these read-only collection-type properties to be defined/created from a comma-delimited string. In terms of Grid, this means that the ColumnDefinitions and RowDefinitions properties are able to be defined by providing a comma-delimited string of ColumnDefinitions (Width values) and a comma-delimited string of RowDefinitions (Height values), respectively. However, this feature is also applicable to other scenarios within the XAML language where collection-typed read-only properties are created (see Remarks for examples of such scenarios).
```
alwu-msft: Clarification: this new feature's purpose is to "allow the initialization of collection-type properties (even if they are read-only) using element attribute syntax". The CreateFromString attribute is one potential way in which a collection type might take advantage of this feature.

anawishnoff: Fixed.

active/gridsyntax/GridSyntaxSpec.md
```diff
@@ -0,0 +1,335 @@
...
+
+# Description
+<!-- Use this section to provide a brief description of the feature.
+For an example, see the introduction to the PasswordBox control 
+(http://docs.microsoft.com/windows/uwp/design/controls-and-patterns/password-box). -->
+## Overview
+A new XAML language feature that allows the initialization of collection-type properties (even if they are read-only) using element attribute syntax.
+
+## Implementation Details
+Read-only collection-type properties will be assigned a new CreateFromString attribute. The CreateFromString attribute will allow them to be defined/created from a comma-delimited string. In terms of Grid, this means that the ColumnDefinitions and RowDefinitions properties are able to be defined by providing a comma-delimited string of ColumnDefinitions (Width values) and a comma-delimited string of RowDefinitions (Height values), respectively. However, this feature is also applicable to other scenarios within the XAML language where collection-typed read-only properties are created (see Remarks for examples of such scenarios).
```
alwu-msft: At a high level, what we're implementing is the following:
1. When the parser encounters an element attribute that is assigning to a collection-type property, it recognizes that initialization of a collection from a string is being requested.
2. The parser tokenizes the attribute value (a comma-delimited string) into a series of substrings, each of which represents the initialization value of a single element of the collection
3. The parser will perform the usual "initialize an instance of a type from a string" operation. 
4. If the type has a ContentProperty, then a new instance of the type will be created (zero-argument constructor) and the string value will be assigned to the content property (this might in turn invoke a type converter, but the details are at a lower abstraction level and so are not germane to this discussion).
5. If the type has either a built-in type converter or a CreateFromString attribute specifying a factory method, then that will be invoked with the string value as an argument, returning a new instance of the type.
6. If neither step 4 nor 5 is performed, then this is an error.

For Grid-specific portion of this work, we'll be following step 4 from above, adding the ContentProperty attribute to RowDefinition and ColumnDefinition.

active/gridsyntax/GridSyntaxSpec.md
```diff
@@ -0,0 +1,335 @@
...
+
+### New succinct syntax for more complex types
+The code below shows an example of how the syntax would be used for collection-typed properties that have more complex value types.
+
+This shows how Point objects will be delimited by a comma, but indvidual values will be encased in single quotes.
+```xml
+<Foo PointList=" '1,2', '9,4', '2,6'"/>
+```
+
+This shows how String values will be supported, even if they include commas or single quotes inside of them. Each string will be encased in single quotes and delimited by commas, and single quotes inside of a string will be escaped by using an HTML entity. The collection shown below has 4 elements, and 'hello, world!' is one element.
```
alwu-msft: Following the XML grammar, if the attribute value is surrounded by `'`, then `"` is the grouping character that is used within the string. Basically, attribute values can be enclosed with either `'` or `"`, which then allows the other quotation mark to be used within the value itself without needing to be escaped with an XML character entity.

active/gridsyntax/GridSyntaxSpec.md
```diff
@@ -0,0 +1,335 @@
...
+### Syntax Details and Corner Cases
+The new succinct syntax is formed by assigning a read-only collection-typed property to a collection of values. This collection of values is encased between a set of double quotes. String values, as well as any other more complex type that may require commas within it (i.e. Point objects, sets), are encased between a set of single quotes. This ensures that string values may contain commas, and Points, sets, or any other kind of nested collection object may be supported. To use a single quote within a complex value that is encased in single quotes (i.e., to put a single quote or apostrophe inside of a string value), the HTML entity (&apos;) should be used to escape that character.
+
+### Other Use Cases for Succinct Syntax
+The succinct syntax was created with the goal of making Grid more intuitive. However, this new syntax will be implemented as a language feature and work properly for any read-only collection-typed property. The current syntaxes will still be functional for all other scenarios, as it will be with Grid. Examples of some of these scenarios include, but are not limited to:
```
alwu-msft: The language feature will work for any collection-typed property, not only those that are read-only.

active/gridsyntax/GridSyntaxSpec.md
```diff
@@ -0,0 +1,335 @@
...
+
+<!-- Option 2: Put these descriptions in the below API Details section,
+with a "///" comment above the member or type. -->
+
+# API Details
+<!-- The exact API, in MIDL3 format (https://docs.microsoft.com/en-us/uwp/midl-3/) -->
+
+**ColumnDefinition API**
+```csharp
+[contract(Windows.Foundation.UniversalApiContract, 1)]
```
alwu-msft: I don't think this is correct. We should follow up with @MikeHillberg off-line about our contract story.

active/gridsyntax/GridSyntaxSpec.md
```diff
@@ -0,0 +1,335 @@
...
+[webhosthidden]
+
+/// Sets the ColumnDefinition's content property to the Width property.
+[contentproperty("Width")]
```
alwu-msft: It'd be cool if we had some way to highlight this to indicate that it's what's being added.

active/gridsyntax/GridSyntaxSpec.md
```diff
@@ -0,0 +1,335 @@
...
+};
+```
+
+**ColumnDefinitions API**
+```csharp
+[contract(Windows.Foundation.UniversalApiContract, 1)]
+[webhosthidden]
+
+/// Allows ColumnDefinitions to be createable from string using the method Create.
+[CreateFromString(MethodName = nameof(Create))]
```
alwu-msft: We're not employing the CreateFromString attribute mechanism.

active/gridsyntax/GridSyntaxSpec.md
```diff
@@ -0,0 +1,335 @@
...
+};
+```
+
+**RowDefinitions API**
+```csharp
+[contract(Windows.Foundation.UniversalApiContract, 1)]
+[webhosthidden]
+
+/// Allows RowDefinitions to be createable from string using the method Create.
+[CreateFromString(MethodName = nameof(Create))]
```
alwu-msft: We're not employing the CreateFromString attribute mechanism.

active/gridsyntax/GridSyntaxSpec.md
```diff
@@ -0,0 +1,335 @@
...
+};
+```
+
+**RowDefinition API**
+```csharp
+[contract(Windows.Foundation.UniversalApiContract, 1)]
+[webhosthidden]
+
+/// Sets the RowDefinition's content property to the Height property.
+[contentproperty("Height")]
```
alwu-msft: It'd be cool if we had some way to highlight this to indicate that it's what's being added.

mdtauk:
Will there be an audit as to other controls and properties that could benefit from adopting this collection type and thus supporting a simplified syntax?

active/gridsyntax/GridSyntaxSpec.md
anawishnoff: Hope its okay that I used your words in the spec - added this in.

active/gridsyntax/GridSyntaxSpec.md
anawishnoff: I tried adding in html/even making it a block quote, but nothing worked because its inside of a code snippet

anawishnoff:
@mdtauk Not sure if there will be exactly an audit per se, but I definitely hope to at least gather some popular controls/their properties that will benefit from this new language feature. There are a couple mentioned in the spec, but I hope to include an expanded list with the release of this feature.

active/gridsyntax/GridSyntaxSpec.md
alwu-msft: That's unfortunate. :(

active/gridsyntax/GridSyntaxSpec.md
```diff
@@ -0,0 +1,313 @@
...
+
+# Description
+<!-- Use this section to provide a brief description of the feature.
+For an example, see the introduction to the PasswordBox control 
+(http://docs.microsoft.com/windows/uwp/design/controls-and-patterns/password-box). -->
+A new XAML language feature that allows the initialization of collection-type properties (including read-only properties) using element attribute syntax. We will also be assigning properties within Grid to make sure that this syntax works properly specifically for Grid's use cases.
+
+1. When the parser encounters an element attribute that is assigning to a collection-type property, it recognizes that initialization of a collection from a string is being requested.
+2. The parser tokenizes the attribute value (a comma-delimited string) into a series of substrings, each of which represents the initialization value of a single element of the collection
+3. The parser will perform the usual "initialize an instance of a type from a string" operation.
```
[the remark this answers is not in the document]
anawishnoff: I put in a reference to section 6.6.4 on Value Creation from Attribute Text - hope that's the correct section, but we can discuss more at the spec review.

active/gridsyntax/GridSyntaxSpec.md
```diff
@@ -0,0 +1,313 @@
...
+      RowDefinitions="1*, Auto, 25, 14, 20">
+</Grid>
+```
+
+### New succinct syntax for more complex types
+The code below shows an example of how the syntax would be used for collection-typed properties that have more complex value types.
+
+This shows how Point objects will be delimited by a comma, but indvidual values will be encased in single quotes or double quotes, whichever is not used to enclose the attribute value.
+```xml
+<Foo PointList=" '1,2', '9,4', '2,6'"/>
```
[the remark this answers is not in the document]
anawishnoff: @MikeHillberg we previously looked at ColorWheel for Silverlight and its Colors property, is that what you're referring to? (Sorry for the later responses! Just getting to this now.)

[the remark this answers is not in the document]
MikeHillberg: Yes, that one's more real-world than <Foo> :)
Although you have other examples below too.

active/gridsyntax/GridSyntaxSpec.md
MikeHillberg: This is on the Width property, which is indeed UAC1.

active/gridsyntax/GridSyntaxSpec.md
MikeHillberg: This is a frustrating issue. Maybe we should adopt a pattern of highlighting lines with comments. E.g.

```c#
    Double SomeProperty;

    // vvvvv
    int TheNewProperty;
    // ^^^^^
    string SomeOtherProperty
```

active/gridsyntax/GridSyntaxSpec.md
```diff
@@ -0,0 +1,310 @@
...
+```
+
+This shows how String values will be supported, even if they include commas or quotes inside of them. Attribute values (i.e `Words` in this example) can be enclosed with either ' or ", so whichever quotation mark is not used can be employed within the value itself, and does not need to be escaped as an XML character entity. The collection shown below has 4 elements, and 'hello, world!' is one element.
```
MikeHillberg:
```suggestion
This shows how String values will be supported, even if they include commas or quotes inside of them. Following standard XML syntax,  you can enclose portions within attribute values (e.g. `hello` in this example) with opposite quotes. So for example if the attribute value uses double quotes, you can use single quotes within the attribute value. The collection shown below has 4 elements, and 'hello, world!' is one element.
```

active/gridsyntax/GridSyntaxSpec.md
```diff
@@ -0,0 +1,310 @@
...
+
+# Remarks
+<!-- Explanation and guidance that doesn't fit into the Examples section. -->
+
+<!-- APIs should only throw exceptions in exceptional conditions; basically,
+only when there's a bug in the caller, such as argument exception.  But if for some
+reason it's necessary for a caller to catch an exception from an API, call that
+out with an explanation either here or in the Examples -->
+### Syntax Details and Corner Cases
+The new succinct syntax is formed by assigning a collection-type property to a collection of values. This collection of values is encased between a set of double quotes. String values, as well as any other more complex type that may require commas within it (i.e. Point objects, sets), are encased between a set of single quotes. This ensures that string values may contain commas, and Points, sets, or any other kind of nested collection object may be supported. To use a single quote within a complex value that is encased in single quotes (i.e., to put a single quote or apostrophe inside of a string value), the HTML entity (&apos;) should be used to escape that character.
```
MikeHillberg:
```suggestion
The new succinct syntax is formed by assigning a collection of values to a read-only collection-type property. This collection of values is encased between a set of quotes. String values, as well as any other more complex types that may require commas within it (i.e. Point objects, sets), are encased between a set of opposite quotes. This ensures that string values may contain commas, and Points, sets, or any other kind of nested collection object may be supported.
```

active/gridsyntax/GridSyntaxSpec.md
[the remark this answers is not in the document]
anawishnoff: Just changed it 👍

active/gridsyntax/GridSyntaxSpec.md
anawishnoff: I like that idea, I'll incorporate into this spec for now!

active/gridsyntax/GridSyntaxSpec.md
anawishnoff: In WinUI3, we won't have the UAC contract

PR #51, opened 2019-11-15
Spec suggestions and feedback

MikeHillberg:


active/NumberBox/NumberBox.md
```diff
@@ -57,17 +58,30 @@ example code with each description. The general format is:
...
+<NumberBox x:Name="EquationInputNumberBox" Header="Enter equation:" />
 ```
-
 ![](images/numberbox-placeholder-text.png)
 
 ### Enable calculation support
 
+To let the user enter equations, use the `AcceptsCalculation` property. This default to false.
+
+<!-- todo: Need to briefly tell the story of calculations here, I'm not sure how it works. When the user hits enter, does it do the calculation? Does the user have to enter the '='? Does it keep showing the equation after it's done the calculation or does setting this property just allow me to type non-numerics into the box?
```
teaP: Currently, when the user hits enter or the control loses focus, it calculates the value of the expression. = is not required (or allowed). The text input is replaced with the answer at that time.

active/NumberBox/NumberBox.md
```diff
@@ -80,6 +94,10 @@ XAML
 
 ### Add increment and decrement stepping
 
+Use the `SpinButtonPlacementMode` property to enable buttons in the NumberBox control that can be clicked to increment or decrement the value in the NumberBox. This defaults to `Hidden`, set to `Inline` to enable the buttons. The amount of increment/decrement is specified with the `StepFrequency` property, which defaults to 1.
+
+<!-- todo: What happens if I set this as well as AcceptsCalculation? Is this ignored? -->
```
teaP: They work together -- you could enter "5 + 3", hit enter to get "8", and then press up or down to get 9 or 7. Although I don't know what happens if you type in an expression and press one of the arrows before evaluation occurs... I added a note for myself to check that in #1505.

active/NumberBox/NumberBox.md
```diff
@@ -159,24 +185,31 @@ unsealed runtimeclass NumberBox : Windows.UI.Xaml.Controls.Control
     String Text; 
     String PlaceholderText;
```
teaP: I'm not sure that PlaceholderText (although I bind it to TextBox.PlaceholderText) will ever be visible. A community member has an open question about whether the NumberBox can ever be empty -- right now it can't.

active/NumberBox/NumberBox.md
```diff
@@ -159,24 +185,31 @@ unsealed runtimeclass NumberBox : Windows.UI.Xaml.Controls.Control
...
 
+<!-- todo: Why does this say "basic"? Is there also an advanced validation mode? -->   
     NumberBoxBasicValidationMode BasicValidationMode;
 
     Boolean AcceptsCalculation;
```
teaP: Had a comment on my last PR about whether this should be "AcceptCalculations", plural. I also would like to through "AcceptsExpressions" in the ring because I think it's more accurate.

active/NumberBox/NumberBox.md
```diff
@@ -159,24 +185,31 @@ unsealed runtimeclass NumberBox : Windows.UI.Xaml.Controls.Control
...
+<!-- NumberFormatter is worth an example -->
     Windows.Globalization.NumberFormatting.INumberFormatter2 NumberFormatter;
 
+<!-- todo: Suggest DecrementButtonIcon naming, to match DecrementButtonStyle -->
+
     IconSource DecrementIcon;
     IconSource IncrementIcon;
     Windows.UI.Xaml.Style DecrementButtonStyle;
     Windows.UI.Xaml.Style IncrementButtonStyle;
     NumberBoxTemplateSettings TemplateSettings{ get; };
```
teaP: Current control does not have properties for icons or styles (there is no reason, if we wanted them to be customizable we should use lightweight styling and define them in resources). I also do not have a TemplateSettings.

active/NumberBox/NumberBox.md
```diff
@@ -159,24 +185,31 @@ unsealed runtimeclass NumberBox : Windows.UI.Xaml.Controls.Control
...
 
+<!-- todo: Do we need ValueChanging? Assume we do, the Remarks or API Notes section should explain it's behavior. Does this respond to TextBox's TextChanging or BeforeTextChanging events? -->
```
teaP: I also would like to know that -- we don't currently have it. But no, since we don't evaluate when the textbox text changes we don't listen to or handle TextChanging.

active/NumberBox/NumberBox.md
teaP: I also have Header and HeaderTemplate as properties.

active/NumberBox/NumberBox.md
```diff
@@ -159,24 +185,31 @@ unsealed runtimeclass NumberBox : Windows.UI.Xaml.Controls.Control
...
+<!-- todo: Why does this say "basic"? Is there also an advanced validation mode? -->   
     NumberBoxBasicValidationMode BasicValidationMode;
 
     Boolean AcceptsCalculation;
 
     NumberBoxSpinButtonPlacementMode SpinButtonPlacementMode{ get; set; };
 
+<!-- todo: This should have the "Is" prefix, to differentiate them from events. -->
     Boolean HyperScrollEnabled;
     Boolean HyperDragEnabled;
```
teaP: This doesn't exist yet and sounds like it probably won't for v1.

active/NumberBox/NumberBox.md
```diff
@@ -159,24 +185,31 @@ unsealed runtimeclass NumberBox : Windows.UI.Xaml.Controls.Control
     String Text; 
     String PlaceholderText;
 
+<!-- todo: Why does this say "basic"? Is there also an advanced validation mode? -->
```
teaP: There isn't :)

active/NumberBox/NumberBox.md
MikeHillberg: So neither OneNote nor Excel behavior. I think that's OK though; I think Excel behavior makes sense, but it could be compatibly be added later.

MikeHillberg: What if there's a syntax error in the equation?

If you Alt+Tab away, does that count as lost focus, and so it evaluates the equation?

active/NumberBox/NumberBox.md
MikeHillberg: Clicking the button will trigger a focus change, maybe that will evaluate it.

active/NumberBox/NumberBox.md
MikeHillberg: Not a deal breaker IMO, but since this is an API, an unqualified "expression" might be confused with an [ExpressionAnimation](https://docs.microsoft.com/uwp/api/Windows.UI.Composition.ExpressionAnimation).

active/NumberBox/NumberBox.md
MikeHillberg: Actually, instead of "respond" I meant to say "correspond". 

On the one hand, I'd expect this event to correspond to [TextBox.TextChanging](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.Controls.TextBox.TextChanging), based simply on the name. If so, we should call that out. On the other hand, TextBox.TextChanging has some complicated semantics, so it's reasonable to consider breaking from that. But that should be called out as well.

active/NumberBox/NumberBox.md
MikeHillberg: Let's save 5 letters :)

SavoySchuler: Edited in the spec.

active/NumberBox/NumberBox.md
SavoySchuler: I believe you're right. I have removed these properties.

active/NumberBox/NumberBox.md
SavoySchuler: Changing focus should trigger the evaluation. The Spinner button press should then effect the newly evaluated value. For example, when the NumberBox contains "2 + 2" as input, and the "Up" Spinner button is pressed, the final value should read "5" as "2+2=4" is evaluated before the increment occurs.

This behavior was evaluated and approved at an earlier spec review.

active/NumberBox/NumberBox.md
SavoySchuler: Agreed. This was temporarily left empty while we were determining how evaluation is triggered. I will add some crisp documentation now that this has been defined.

@teaP, Mike raises an excellent point that NumberBox should accept a leading "=" as not throw Excel-minded users. Given time constraints, I can file this as a bug if you prefer?

SavoySchuler: @MikeHillberg, errors will block evaluation and trigger a validation warning (unless the developer has opted out of our prescribed validation warning options).

active/NumberBox/NumberBox.md
SavoySchuler: I have added the "Is" prefix to all hyper and wrap properties. 

Hyper properties can be re-prioritized for the 3.0.

active/NumberBox/NumberBox.md
SavoySchuler: Added!

SavoySchuler: Is that an artifact of needing a valid digit occupying the Text field?

active/NumberBox/NumberBox.md
SavoySchuler: Agreed that this should at least be a plural "Calculations" if that is the route taken.

@MikeHillberg, would "AcceptsNumericalExpressions" be more clear?

active/NumberBox/NumberBox.md
teaP: Sure, let's file a bug on the = thing. Currently the only validation we have is not changing the value to anything invalid, so if the user types in an invalid expression, the value just doesn't change and the text reverts back to displaying the current value.

active/NumberBox/NumberBox.md
teaP: The spinner buttons don't get focus. I verified that currently, it ignores everything you typed, which is not so good. :) I will handle it as a bug.

active/NumberBox/NumberBox.md
teaP: Yes, the text should always display whatever is in the Value property. @adrientetar would like the text to be empty when the value is NaN, shall we do that?

active/NumberBox/NumberBox.md
SavoySchuler: Will do,

And that is correct - right now we only have the "override invalid input" and "I'll take care of validation myself" options. Come 3.0, we will be able to add options to display validation warnings for the end user.

active/NumberBox/NumberBox.md
adrientetar: A compound calculation [is a calculation](https://en.wikipedia.org/wiki/Calculation): *For example, multiplying 7 by 6 is a simple algorithmic calculation. Estimating the fair price for financial instruments using the Black–Scholes model is a more complex algorithmic calculation.*

And the current naming has the advantage of matching TextBox.AcceptsReturn.

active/NumberBox/NumberBox.md
```diff
@@ -57,17 +58,30 @@ example code with each description. The general format is:
...
 ```XAML
-<NumberBox x:Name="EquationInputNumberBox" Header="Enter equation:" PlaceholderText="A + B * C" />
+<NumberBox x:Name="EquationInputNumberBox" Header="Enter equation:" />
 ```
-
 ![](images/numberbox-placeholder-text.png)
 
 ### Enable calculation support
 
+To let the user enter equations, use the `AcceptsCalculation` property. This default to false.
```
YuliKl:
```suggestion
To let the user enter equations, use the `AcceptsCalculation` property. This defaults to false.
```

active/NumberBox/NumberBox.md
MikeHillberg: I think it's better to leave out the '=' prefix now, and save it for later when it "works". I.e. when I do "=3+2" and tab away I get '5', but when I tab back I get "=3+2".

active/NumberBox/NumberBox.md
MikeHillberg: We've moved away from NaN as a sentinel value and using nullable instead.

active/NumberBox/NumberBox.md
MikeHillberg: AcceptsNumericalExpressions is long, and good point about AcceptsReturn. I like AcceptsCalculation.

active/NumberBox/NumberBox.md
teaP: Should I make Value an IReference<double> then?

active/NumberBox/NumberBox.md
SavoySchuler: @MikeHillberg, it is not an ask for the control to preserve history/for expressions to be accessible post evaluation. Does a leading "=" imply that functionality? If so, it should not be supported at this time.

active/NumberBox/NumberBox.md
SavoySchuler: @adrientetar could you help me understand the value of being able to have an empty NumberBox? 

In order to reduce end user confusion, the guidance is that a NumberBox should have _at least_ either a numeric value or placeholder text indicating how a NumberBox should be used (e.g, "A + B"). 

The only reason I see for allowing nullable is so that PlaceholderText can be visible - but this adds complexity to the validation features.  

I'm interested to hear @adrientetar's perspective - otherwise nullable + PlaceholderText could be left for V2 pending justifiable interest.

active/NumberBox/NumberBox.md
SavoySchuler: I will leave this as is and mark "Resolved."

active/NumberBox/NumberBox.md
adrientetar: @SavoySchuler, I have number boxes that display the position and size of selected objects in my app (and you can edit to move, similar to Adobe XD sidebar fields). When nothing is selected, these get disabled and display nothing. So yes, being able to display nothing is important.

active/NumberBox/NumberBox.md
MikeHillberg: Yes, I think an interesting future feature would be to preserve the expression if it's prefixed with an '='. In the meantime, it would be better to not support the prefix at all, so that we're reserving the ability to add it in the future.

active/NumberBox/NumberBox.md
MikeHillberg: Could that be accomplished by having no placeholder text?

teaP: No, when the control doesn't have focus it always displays the current Value, which is either a number or NaN.

active/NumberBox/NumberBox.md
SavoySchuler: ValueChanging was requested before we planned built-in input validation for NumberBox. I am finding its value limited now. Correct me if I am wrong, but I believe @mrlacey originally called out the absence of ValueChanging. @mrlacey - I know your time is valuable, so I hope this is a quick question, but are there any scenarios in which you are looking to use NumberBox that would be disabled if ValueChanging were absent? 

Otherwise, I also believe it is acceptable to break from TextBox event parity considering the addition and ease of our coming input validation for NumberBox.

active/NumberBox/NumberBox.md
SavoySchuler: Noted. I'll add it to the list of features to consider for V2 and resolve this conversation.

active/NumberBox/NumberBox.md
SavoySchuler: To synthesize, @adrientetar is using the nullable state as an indicator to disable the corresponding attribute. I certainly see this need scaling. 

@teaP Is a valid nullable state an easy add?

teaP: Shouldn't be too bad. The downside from the app side is I don't believe you can x:Bind to a normal double, and prior to RS5 you couldn't set the value in xaml at all.

SavoySchuler: @jevansaks do you have insight on this? 

Those cons seem unacceptably large. A workaround for scenarios like @adrientetar's could be to disable the corresponding property outside of the control with a CheckBox. While this isn't as seamless as what @adrientetar has implemented, the the pro does not seem worth the cost of losing x:Bind and requiring RS5 for full compatibility. 

 scenario  to handle disabling the property

active/NumberBox/NumberBox.md
```diff
@@ -156,55 +129,49 @@ unsealed runtimeclass NumberBox : Windows.UI.Xaml.Controls.Control
...
     String Text; 
     String PlaceholderText;
-
-    NumberBoxBasicValidationMode BasicValidationMode;
+  
+    NumberBoxValidationMode ValidationMode;
 
     Boolean AcceptsCalculation;
 
     NumberBoxSpinButtonPlacementMode SpinButtonPlacementMode{ get; set; };
```
SavoySchuler: Now that the SpinButton is a Flyout that will appear above the control, I believe this should be renamed to "IsSpinButtonEnabled".  

@MikeHillberg, @teaP, does that sound ok to you?

SavoySchuler: I take that back - I am paging in our extensibility conversation. An enum makes it easy to add more placement modes later.

teaP: Oh, uh, I was going to leave them both in. So there would be three modes -- none, inline, and flyout. Does no one want inline as an option?

active/NumberBox/NumberBox.md
MikeHillberg: Also, what is the purpose of the args `IsContentChanging` property? It's not obvious from the name, and not explained anywhere.

active/NumberBox/NumberBox.md
SavoySchuler: I forget you had both already prototyped. This was just to save cost on secondary configurations. If the implementation work is done, let's keep them both! 

I'll fix the doc and API to acknowledge this.

active/NumberBox/NumberBox.md
SavoySchuler: This was an artifact of originally intending event parity with TextBox's ValueChanged and ValueChanging events and their [args](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.controls.textboxtextchangingeventargs). 

If we're in agreement to cut ValueChanging from NumberBox, IsContentChanging would be dropped as well.

active/NumberBox/NumberBox.md
adrientetar: I mean, if you add a way to display NaN as empty string (maybe optionally), that's all I really need. Or could I somehow set the Text property in this case to empty string?

MikeHillberg: With the nullable, you could set the Value to null. That would cause it to display the PlaceholderText (which could be empty).

active/NumberBox/NumberBox.md
```diff
@@ -118,28 +93,22 @@ with a "///" comment above the member or type. -->
 | StepFrequency | Gets or sets the value part of a value range that steps should be created for. |
 
 ## API Details
-<!-- The exact API, in MIDL3 format (https://docs.microsoft.com/en-us/uwp/midl-3/) -->
+<!-- todo: Missing some APIs in here, like TemplatSettings class and event args classes -->
```
SavoySchuler: I added this for a moment, but was unsure of what could or should be templated here. @MikeHillberg or @teaP, did you have something particular in mind?

active/NumberBox/NumberBox.md
adrientetar: > The downside from the app side is I don't believe you can x:Bind to a normal double,

Yes but you make a double? property in your view model then that property can be assigned double values seamlessly, IIUC.

As an alternative… Qt's QSpinBox has protected functions to override the conversion between value and text

image

If there is something like that, I could use to change conversion of NaN to empty string. I mean that would suffice for my use, nullable is still good if you want to be able to display null state and NaN separately.

active/NumberBox/NumberBox.md
```diff
@@ -118,28 +95,22 @@ with a "///" comment above the member or type. -->
...
 };
 
-runtimeclass NumberBoxValueChangingEventArgs
-{
-    Boolean IsContentChanging;
-};
-
 runtimeclass NumberBoxValueChangedEventArgs
 {
     System.Windows.Controls.UndoAction Action;
```
teaP: The current event args look like this:
```
runtimeclass NumberBoxValueChangedEventArgs
{
    Double OldValue{ get; };
    Double NewValue{ get; };
};
```

SavoySchuler: The [original event args](https://docs.microsoft.com/en-us/dotnet/api/system.windows.controls.textchangedeventargs.-ctor?view=netframework-4.8#System_Windows_Controls_TextChangedEventArgs__ctor_System_Windows_RoutedEvent_System_Windows_Controls_UndoAction_) were chosen to maintain alignment with TextBox's events. But seeing as how we are already diverging, the args you implemented are certainly more clear + easy to use and makes the most sense for the scenarios this control is used in. 

I have the spec updated to match. 😊

active/NumberBox/NumberBox.md
SavoySchuler: Handled offline. Events updated and TemplateSettings unneeded. Resolving.

active/NumberBox/NumberBox.md
SavoySchuler: @adrientetar - we have enabled this scenario. When a NumberBox is empty, the value will read NaN and the placeholder text will display (if set).

PR #53, opened 2019-11-20
ColorPicker Orientation API spec

chingucoding:
This is the API spec for [this proposal](https://github.com/microsoft/microsoft-ui-xaml/issues/1637), which proposes additional layouts for the ColorPicker.


mdtauk:
Will **DisplayMode** only encompass layouts.  Or could it be made to handle HSL, RGB, Greyscale, Alpha, and additional colour formulas as well as simplified and possible swatch views?

YuliKl:
> Will **DisplayMode** only encompass layouts. Or could it be made to handle HSL, RGB, Greyscale, Alpha, and additional colour formulas and libraries - as well as simplified and possible swatch views?

I think this interpretation of DisplayMode is both reasonable, and not what @chingucoding originally proposed. I think the original proposal was trying to get at `Orientation` in the same sense as used by `StackPanel`, so perhaps that would be a better API name with a more limited scope?

api_review/ColorPickerDisplayMode/ColorPickerDisplayModeSpec.md
```diff
@@ -0,0 +1,118 @@
...
+This API is an update to the existing ColorPicker control. 
+The aim of this new API is, to make the ColorPicker more versatile to use and provide different layouting ways to maximize the space available.
+
+# Description
+<!-- Use this section to provide a brief description of the feature.
+For an example, see the introduction to the PasswordBox control 
+(http://docs.microsoft.com/windows/uwp/design/controls-and-patterns/password-box). -->
+
+With the DisplayMode API for the ColorPicker, the ColorPicker is able to either layout itsself horizontally or vertically. 
+In addition to this the ColorPicker can also choose its layout based on the space available.
```
YuliKl: This auto behavior doesn't feel strictly necessary to me, and omitting it would, I suspect, greatly simplify the implementation. 

I read the original proposal as giving app developers a choice of whether the control should lay itself out horizontally or vertically. So it feels ok to enable that choice without trying to do anything clever automatically.

api_review/ColorPickerDisplayMode/ColorPickerDisplayModeSpec.md
```diff
@@ -0,0 +1,118 @@
...
+DisplayMode: The DisplayMode determining wether to display horizontally or vertically.
+# API Details
+<!-- The exact API, in MIDL3 format (https://docs.microsoft.com/en-us/uwp/midl-3/) -->
+```c#
+
+[webhosthidden]
+unsealed runtimeclass ColorPicker
+{
+    /// Gets or sets the DisplayMode of the ColorPicker
+    ColorPickerDisplayMode DisplayMode = ColorPickerDisplayMode.Vertical;
```
YuliKl: As I suggested in the comments, perhaps

```suggestion
    ColorPickerOrientation Orientation = ColorPickerOrientation.Vertical;
```

api_review/ColorPickerDisplayMode/ColorPickerDisplayModeSpec.md
```diff
@@ -0,0 +1,118 @@
...
+}
+
+enum ColorPickerDisplayMode
```
YuliKl: There's already an [Orientation enum](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.controls.orientation), perhaps we can leverage that, either directly or as inspiration?

chingucoding: Great idea! I updated the spec to use the existing orientation enum.

api_review/ColorPickerDisplayMode/ColorPickerDisplayModeSpec.md
chingucoding: As you suggested, I updated the proposal to use the existing Orientation enum. I also updated the property name to "Orientation".

api_review/ColorPickerDisplayMode/ColorPickerDisplayModeSpec.md
chingucoding: Seems like a reasonable idea. If an "Auto" mode is really necessary, developers could achieve it using VisualStateTriggers or we can add it later. I removed the "Auto" mode from the spec proposal.

chingucoding:
> Will **DisplayMode** only encompass layouts. Or could it be made to handle HSL, RGB, Greyscale, Alpha, and additional colour formulas and libraries - as well as simplified and possible swatch views?

I updated my spec proposal, to make it more clear what this will achieve. Regarding your idea, maybe this could be an additional API/property? However I am not sure which options this would provide. Would this be something along the "Inline" idea?

mdtauk:
I was putting the thought into the conversation, because **DisplayMode** is a loose term which could be used for future proposals and ideas for making the control richer.

If the scope is purely around the arrangement or alignment of the parts of the control - then perhaps the name should be similarly focused.

Or keep DisplayMode and allow future additions over time.

chingucoding:
> I was putting the thought into the conversation, because **DisplayMode** is a loose term which could be used for future proposals and ideas for making the control richer.
> 
> If the scope is purely around the arrangement or alignment of the parts of the control - then perhaps the name should be similarly focused.
> 

Yes, I agree "DisplayMode" was a very loose term. I think having the term "Orientation" for this is better to convey a clear purpose.

> Or keep DisplayMode and allow future additions over time.

This would also work. However one problem I see is how new additions would work with that. For example if we add a new labeling mode, and we already have DisplayMode=Vertical and DisplayMode=Horizontal. will there be Vertical,VerticalInline,... potentially creating a long list of different display modes that are essentially a combination of multiple properties fused into one?

Maybe if there is a need for DisplayMode we can add it later :)

mdtauk:
@chingucoding  If you are going with **Orientation**, then it should follow convention for other controls' orientation property.

I think that may rule out any "inline" variations.

Maybe I will make some mockups, and submit a proposal for alternate/additional "DisplayModes"

mdtauk:
image

Here is my thinking.

chingucoding:
> image
> 
> Here is my thinking.

Those look great! However I think the old/new color box needs to be on the side since we also have a textbox for opacity. Placing the color block on the site will create that needed place. I will update my screenshots.

mdtauk:
> > image
> > Here is my thinking.
> 
> Those look great! However I think the old/new color box needs to be on the side since we also have a textbox for opacity. Placing the color block on the site will create that needed place. ~I will update my screenshot.~ I updated the screenshot.
> 
> Regarding the ColorSwatches, I think we should create a proposal for this in WinUI first. 😅
>

Does this image help?
image

chingucoding:
> Do these images help?

Thank you for your effort! I have already changed my horizontal screenshot (though it is not pixel perfect like your images, since I used Paint 😅).

I love the idea with the color swatches! Do you want to create a proposal for that, or would you like me to create the proposal? 

Also what is your opinion on how to handle this? We could use an enum DisplayMode (Picker, Swatches, Both) or we could use a bool EnableSwatches (false only shows the color picker, true shows the navigation and enables the swatches page).

YuliKl:
> I love the idea with the color swatches! Do you want to create a proposal for that, or would you like me to create the proposal?
> 
> Also what is your opinion on how to handle this? We could use an enum DisplayMode (Picker, Swatches, Both) or we could use a bool EnableSwatches (false only shows the color picker, true shows the navigation and enables the swatches page).

I agree that the color swatches idea should be its own proposal. Seems best to avoid discussing implementation details of a feature in the API spec comments of a different feature 😉

mdtauk:
> > Do these images help?
> 
> Thank you for your effort! I have already changed my horizontal screenshot (though it is not pixel perfect like your images, since I used Paint 😅).
> 
> I love the idea with the color swatches! Do you want to create a proposal for that, or would you like me to create the proposal?
> 
> Also what is your opinion on how to handle this? We could use an enum DisplayMode (Picker, Swatches, Both) or we could use a bool EnableSwatches (false only shows the color picker, true shows the navigation and enables the swatches page).

You seem like you have a good handle on this control, and other than including it because the Win32 common dialog has swatches, it could have its uses.

- It would need to suppress the "Windows colours" section if the app developer wishes it;
- Possibly set the header text for the "App colours" which could be a list of colour values in the control declaration;
- Allow/Disallow the use of Recent colours;

But it would also be a mechanism for additions to the Color Picker in the future, if use cases emerge.  Maybe it could host custom XAML content?

----

Part of WinUI is about allowing developers to move existing old UI over to XAML - and so this would fill a gap in the WinUI control, which is present in the Win32 dialog.

chingucoding:
> > > Do these images help?
> > 
> > 
> > Thank you for your effort! I have already changed my horizontal screenshot (though it is not pixel perfect like your images, since I used Paint 😅).
> > I love the idea with the color swatches! Do you want to create a proposal for that, or would you like me to create the proposal?
> > Also what is your opinion on how to handle this? We could use an enum DisplayMode (Picker, Swatches, Both) or we could use a bool EnableSwatches (false only shows the color picker, true shows the navigation and enables the swatches page).
> 
> You seem like you have a good handle on this control, and other than including it because the Win32 common dialog has swatches, it could have further uses.
> 
> * Possibly expose the "Windows colours" as shown in the Settings app, in WinUI or the System; (colours that have had legibility and usability tested by Microsoft;
> * It would need to suppress the "Windows colours" section if the app developer wishes it;
> * Possibly set the header text for the "App colours" which could be a list of colour values in the control declaration;
> * Allow/Disallow the use of Recent colours;
> 
> But it would also be a control API mechanism for additional sections to the Color Picker in the future, if use cases emerge. Maybe it could host custom XAML content?
> 
> Part of WinUI is about allowing developers to move existing old UI over to XAML - and so this would fill a gap in the WinUI control, which is present in the Win32 dialog.

I have created a proposal with your awesome mock up over at WinUI: https://github.com/microsoft/microsoft-ui-xaml/issues/1659 😊

chingucoding:
@kikisaints Is there anything preventing this API proposal from being merged? 😅

chingucoding:
Is "api_review/ColorPickerDisplaymode" the correct folder for the API spec or should the spec files should be moved into a different folder e.g. "active/ColorPickerDisplayMode"?

robloo:
What about the RGB textbox orientaion themselves? I really would like to see that update. Using DisplayMode would also allow this detail.

**DisplayMode.Vertical**



**DisplayMode.VerticalCompact**



**DisplayMode.Horizontal**

Capture5

**DisplayMode.HorizontalCompact**

Needs a mockup

chingucoding:
I don't think that compact would make much sense as the abbreviation of some colors may have the same letter and then it can be confusing having the same letter twice. I am not sure though if that could be the case anywhere. One thing that you also need to think of are languages where abbreviations can become difficulet e.g. Arabic or Japanese.

So I'm not entirely sure if a compact mode the way you design it could work well with localization.

robloo:
@chingucoding Photoshop(r) makes color channel abbreviations work. I think we could make abbreviations work as well. Keep in mind an abbreviation doesn't mean only 1 letter. In other languages it maybe be a bit longer to avoid ambiguity. Design-wise compact mode is much better than the current design.

chingucoding:
But how many letters or how much space would you want to accept for the abbreviations? Because in your screenshot, there is not much until it would look crowded and spacing would become awkward.

robloo:
I would consider 3 letters the max, but we may not need to translate the abbreviations if it's considered a technical term.

chingucoding:
Hmm I think with three letters, we might already have problem when textscaling is increasing.

Also currently we do localize those names (e.g.: locale de-DE):
image

robloo:
Yes, I understand the full names are localized: 'R' 'G' 'B' and 'A' may not need to be though and other apps don't translate these color abbreviations. Same for HSL, etc.

mdtauk:
By creating this new Orientation property an **enum** it would allow the addition of compact or other variations in the future.  If it makes it easier for V1, then keeping the Vertical as is, and adding the Horizontal should be the first step.

The reason for considering compact at all, is to offer another solution to the problem of how much space the ColorPicker currently takes up on the screen.  There is also a potential for confusion between Compact Orientations, and Compact Density - which would be two separate things as the proposal stands.  Maybe Compact is the wrong word to use, so alternatives should be considered.

Also there is the option to reconsider the default arrangement of those text boxes, and offer an _expanded_ option?

chingucoding:
> Yes, I understand the full names are localized: 'R' 'G' 'B' and 'A' may not need to be though and other apps don't translate these color abbreviations. Same for HSL, etc.

Yeah it makes somewhat sense since we also don't localize the colormode dropdown now that I think about this.

> By creating this new Orientation property an enum it would allow the addition of compact or other variations in the future.

Current spec is to use the winrt Orientation enum, so if we do indeed want to add new values later, we would have to switch to a new enum. So we should decide this now, switching the enum later would be a breaking change.

> The reason for considering compact at all, is to offer another solution to the problem of how much space the ColorPicker currently takes up on the screen. There is also a potential for confusion between Compact Orientations, and Compact Density - which would be two separate things as the proposal stands. Maybe Compact is the wrong word to use, so alternatives should be considered.

Yes that is a really good point. While it does indeed take up a lot of space currently, one thing to note is that there is an option to provide a "more" button which collapses those textboxes if they are not needed.

> Also there is the option to reconsider the default arrangement of those text boxes, and offer an expanded option?

The question would be, how should that expanded mode look like? Currently those buttons are already spaced quite well.

----

Also, I am not sure if a HorizontalCompact orientation would make sense at all. There wouldn't be much space we could save when using abbreviations I think.

mdtauk:
image

robloo:
>> By creating this new Orientation property an enum it would allow the addition of compact or other variations in the future.

> Current spec is to use the winrt Orientation enum, so if we do indeed want to add new values later, we would have to switch to a new enum. So we should decide this now, switching the enum later would be a breaking change.

This is actually why I'm bringing this up now. I don't want to get locked into the Orientation enum as it oversimplifies potential future layouts. Whether we stick to the term 'Compact' doesn't matter much to me although I can't think of a better term at the moment.

> Also, I am not sure if a HorizontalCompact orientation would make sense at all. There wouldn't be much space we could save when using abbreviations I think.

I think you'll find the opposite might be true. The horizontal mock up as pictured doesn't allow hardly any spacing for longer text strings from localization. In practice I bet the horizontal mode has to be quite a bit wider to accommodate longer, localized text. 'Compact' could be very useful here as it ensure uniform width and consistent layout among all languages.

mdtauk:
> > > By creating this new Orientation property an enum it would allow the addition of compact or other variations in the future.
> 
> > Current spec is to use the winrt Orientation enum, so if we do indeed want to add new values later, we would have to switch to a new enum. So we should decide this now, switching the enum later would be a breaking change.
> 
> This is actually why I'm bringing this up now. I don't want to get locked into the Orientation enum as it oversimplifies potential future layouts. Whether we stick to the term 'Compact' doesn't matter much to me although I can't think of a better term at the moment.
> 
> > Also, I am not sure if a HorizontalCompact orientation would make sense at all. There wouldn't be much space we could save when using abbreviations I think.
> 
> I think you'll find the opposite might be true. The horizontal mock up as pictured doesn't allow hardly any spacing for longer text strings from localization. In practice I bet the horizontal mode has to be quite a bit wider to accommodate longer, localized text. 'Compact' could be very useful here as it ensure uniform width and consistent layout among all languages.

In the image I posted above, the width would adjust as the strings increase in length - but using a single or 3 letter acronym makes sense to me - but I am not a localisation expert, so I would defer to their decisions.

chingucoding:
> https://github.com/microsoft/microsoft-ui-xaml-specs/pull/53#issuecomment-605377367

I think that without WinUI 3.0 using a NumberBox might be difficult as the logic of verifying that input is handled inside the OS right now, and I am not sure if they would be able to deal with a new control there now (e.g. different typecast would be needed).

> This is actually why I'm bringing this up now. I don't want to get locked into the Orientation enum as it oversimplifies potential future layouts. Whether we stick to the term 'Compact' doesn't matter much to me although I can't think of a better term at the moment.

Something we could also do is have a seperate flag "IsCompact" or something like that. Though that does not scale really well. A small problem I have right now is that Compact sizing does not really fit well in the scope of orientation.

> I think you'll find the opposite might be true. The horizontal mock up as pictured doesn't allow hardly any spacing for longer text strings from localization. In practice I bet the horizontal mode has to be quite a bit wider to accommodate longer, localized text. 'Compact' could be very useful here as it ensure uniform width and consistent layout among all languages.

I can see that one would want to have consistent sizing among languages, but that is something that one could achieve by allocating the minimum space for the largest translations. Also, I am not sure if the difference in translation widths can justify this additional mode. I am not sure if ensuring same sizing among multiple languages is also a goal developers need to ensure and more importantly are actually able to do. 

For example the moment you have a checkbox or RadioButtons, you would either need to specify the width manually or you wont get same sizing among multiple languages.

But I am not a localization expert either, so that is something the team needs to decide.

robloo:
> Something we could also do is have a seperate flag "IsCompact" or something like that. Though that does not scale really well. A small problem I have right now is that Compact sizing does not really fit well in the scope of orientation.

This is all related together, as @mdtauk stated, "to the problem of how much space the ColorPicker currently takes up on the screen." This is not so much an orientation problem as much as it's a control layout/density problem which is more generic than simply orientation. I think the scope needs to be expanded before anything is implemented here so we don't back ourselves into a corner for the future.

I'm also not asking for the 'compact' layout to be implemented now. It would be nice to do it all at once but at least we should consider it and plan for it in the future. I can't stress enough, from a design standpoint, 'compact' is much better than what we have now and it also takes up less space.

Another reason to avoid using the existing Orientation enum is so that we can add an 'Auto' value which would be the default. That could be useful depending on screen real estate and placement in the future.

mdtauk:
@robloo  I like the idea of the control adapting to fit the space it is in, but having the ability to "lock" that orientation is good too.

How the control is used in an app is numerous.  It could be in a ContentDialog, or in a Sidebar, or even on the page itself.  So it needs to be flexible.

Compact Density is where the individual controls that make up the control, get smaller with smaller padding.  If that is applied to the Color Picker, that is different to altering the layout in the way we have discussed - which is why I said it could cause some confusion with that naming.

I think there is value in providing a space saving layout, and there are many different versions of a color picker, where there is no obvious right or wrong way to do it.  So if it becomes an option for this control, it needs to provide a real benefit.

For Compact/Condensed layout Translations - there are the acronyms which could be used as the labels, which could have the full string as a Narrator name and displayed in a Tooltip.

chingucoding:
The initial spec suggested using a "DisplayMode" property, however this was changed for the following reasons:
- Displaymode is to loose as to what it should contain
- @mdtauk Suggested something like `Inline` mode, which I think would have been something like your compact mode. The problem is that Compact and Orientation are two things that are not coupled, and would unnecessarily bloat the enum.
- Auto was not seen/ is not seen as something that would be useful to have, as it creates complexity which should more or less be handled by the developers (how should we choose which is best in a given situation)

@robloo Earlier in this thread, most of those points where already discussed I think, with reasons why we ended up on orientation.

mdtauk:
> * Displaymode is to loose as to what it should contain
>
DisplayMode was mentioned when the Swatches were part of the discussion, Orientation is ok to use, as it contains the two options we are focusing on.

> * @mdtauk Suggested something like `Inline` mode, which I think would have been something like your compact mode. The problem is that Compact and Orientation are two things that are not coupled, and would unnecessarily bloat the enum.
>
If the _Inline_ mode is approved and added, it would still apply whether the control is in the Vertical or Horizontal orientation - so having it be separate is sensible.

> * Auto was not seen/ is not seen as something that would be useful to have, as it creates complexity which should more or less be handled by the developers (how should we choose which is best in a given situation)
> 
> @robloo Earlier in this thread, most of those points where already discussed I think, with reasons why we ended up on orientation.
>
With orientation, it will at least enable the developer to use VisualStateTriggers to switch the control's orientation property as the space it occupies changes.  Not as seamless as Auto, but easier to implement.

chingucoding:
> > Displaymode is to loose as to what it should contain
> 
> DisplayMode was mentioned when the Swatches were part of the discussion, Orientation is ok to use, as it contains the two options we are focusing on.

The question is if compact would make sense with orientation or not.


>> @mdtauk Suggested something like Inline mode, which I think would have been something like your compact mode. The problem is that Compact and Orientation are two things that are not coupled, and would unnecessarily bloat the enum.
> 
> If the Inline mode is approved and added, it would still apply whether the control is in the Vertical or Horizontal orientation - so having it be separate is sensible.

So same would go for compact right? Compact is not something that is dependent on orientation is it?

> > * Auto was not seen/ is not seen as something that would be useful to have, as it creates complexity which should more or less be handled by the developers (how should we choose which is best in a given situation)
> > 
> > @robloo Earlier in this thread, most of those points where already discussed I think, with reasons why we ended up on orientation.
> 
> With orientation, it will at least enable the developer to use VisualStateTriggers to switch the control's orientation property as the space it occupies changes. Not as seamless as Auto, but easier to implement.

That was the reasoning on why we should not provide an Auto. Mainly because there are to many parameters we would have to respect (precedence, what if neither space is enough etc).

mdtauk:
@chingucoding  I would say Orientation would switch between the two layouts, without any changes to the controls making up the whole control - just their position.

Then with using a Compact Density ResourceDictionary, that would use the smaller padding between the text box controls (and possibly a smaller Color Spectrum.)

Finally there would be a property which would change the TextBox headers to be inline, using acronyms and tooltips, moving them onto one line, reducing the overall space the control will take up.

All three of these are independent of each other.

chingucoding:
That seems like the most sensible idea to tackle this. Great summary @mdtauk !
So this spec is pretty much final in that regard then, if we do not want to include any additional properties right?

robloo:
Hm, I still think having one 'DisplayMode' property to control all the various layouts is useful and better than adding different properties. NavigationView is an example of this type of thinking https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.controls.navigationviewdisplaymode. Notice it has 'Compact' as well.

Honestly, I would like someone from Microsoft to weigh in on this. It makes more sense to me to have one 'DisplayMode' property than two (or more in the future) to convey control layout settings. That said, of course it would be valid to have Orientation and 'ColorChannelLayout' enums separately.

chingucoding:
Yes NavigationView has LeftCompact, LeftMinimal, Left and Top. However in that case, there is no Compact without "Left", there simply isn't a TopCompact mode.
In our case compact is independent of the orientation, compact can be/is valid in either Vertical or Horizontal.

chingucoding:
Yes NavigationView has LeftCompact, LeftMinimal, Left and Top. However in that case, there is no Compact without "Left", there simply isn't a TopCompact mode.
In our case compact is independent of the orientation, compact can be/is valid in either Vertical or Horizontal.

robloo:
@chingucoding Yes, I understood that point. It is valid to separate the information of 'compact/normal' from 'vertical/horizontal'. That type of thinking does make sense and is a form of simplification. 

My point is still that this is all describing the layout of the control and perhaps it would be better to be in the same enum. That provides future flexibility for cases we haven't already considered as well. ColorPicker, unlike most other controls, has a lot of child controls within it. It's not primitive by any means and layouts are bound to change as the control evolves over the coming years.

chingucoding:
The problem I see is that that kind of enum could potentially explode. Currently we would have 2 values: Vertical and Horizontal, if we have compact/not compact, we have 4 values. With the next value (maybe inline/not inline or something else) that could potentially double that, and now we have 8 values. And that tends to not be user friendly or easy to use. Especially the naming would come in the way. 

However, potentially adding new values in the future is a good point, and I am not sure if that could happen or not. I think it probably won't in the way orientation would be affected, but that's just a guess here, nobody knows how the future could look like. Something to keep in mind though is the fact that colorpicker really hasn't changed a lot so far, so we need to think if it might change that much at all in the next years or not.

robloo:
> The problem I see is that that kind of enum could potentially explode. Currently we would have 2 values: Vertical and Horizontal, if we have compact/not compact, we have 4 values. With the next value (maybe inline/not inline or something else) that could potentially double that, and now we have 8 values. 

This is also a good point and probably the strongest reasoning against an all-encompassing DisplayMode. I still don't think we should implement Orientation without considering other layouts though.

Could we have a rough design of how to implement the following? All of these are potentially layout related as you discussed above (it's similar to @mdtauk's breakdown). That would help avoid spec'ing ourselves into a corner. 

1. Horizontal/Vertical layouts
2. Compact or Inline color channel layouts
3. Swatches

enum Orientation
 - Orientation.Vertical : As discussed above
 - Orientation.Horizontal : As discussed above

enum ColorChannelDisplayMode
 - ColorChannelDisplayMode.Default : As currently designed
 - ColorChannelDisplayMode.Inline : The inline/compact layout proposal including 'R' 'G' 'B' abbreviations

enum ColorSwatchesDisplayMode
 - ColorSwatchesDisplayMode.Disabled : No swatches visible
 - ColorSwatchesDisplayMode.Default : Default swatches with system recents visible
 - ColorSwatchesDisplayMode.StandardSixteen : The standard sixteen colors that have been around since monitors were first color. This is quick for most people to select from.
 - ColorSwatchesDisplayMode.Custom : Note that swatches probably need to be customized which is going to create a much larger change than simply an enum. A separate list of colors will need to be specified.
IsAlphaSwatchEnabled is probably needed as well

chingucoding:
@robloo I think the enum separation you proposed seems like a reasonable way to do this. 

I think colorswatches could instead be the implemented in the following way:

```c#
enum DisplayMode
{
  ColorsOnly, // Would be current mode
  Swatches, // Only show swatches
  Both // Shows both color swatches and color picker
}

enum ColorSwatchesMode
{
  AccentColors, // Windows 10 accent colors
  StandardSixteen, // Standard sixteen colors
  Custom // Binds to a custom collection of colors
}
```

mdtauk:
I too am in favour of Orientation being kept separate as Horizontal and Vertical.

DisplayMode or DisplayLayout would be good for appendages like a swatch/picker pivot.

Depending on what other changes may come in the future, it may be better to think of the Inline labels as a discreet thing, rather than as additional modes.  Something like **LabelDisplay.Full** and **LabelDisplay.Inline**?

chingucoding:
> Depending on what other changes may come in the future, it may be better to think of the Inline labels as a discreet thing, rather than as additional modes. Something like LabelDisplay.Full and LabelDisplay.Inline?

Yes that could also make sense. Good idea!

mdtauk:
image
Orientation: Vertical and Horizontal

mdtauk:
image
Labels Inline

mdtauk:
image
Swatches

chingucoding:
Those are awesome mockups! I think the last mockup should also be posted on here: https://github.com/microsoft/microsoft-ui-xaml/issues/1659

mdtauk:
@chingucoding  I have done so, but that proposal has not been accepted, so I was focusing on comparing the three separate options and how they affect the control in different ways.

chingucoding:
> @chingucoding I have done so, but that proposal has not been accepted, so I was focusing on comparing the three separate options and how they affect the control in different ways.

Hopefully, it will be one day. Color swatches are something that can really help in certain areas. Also, this comparison is really helpful in determining how the API's should look like!

robloo:
I'm on the same page now and think this all makes sense. I do apologize for throwing a wrench into the discussion and then circling back around to what you already decided. I do feel better knowing we are not overconstraining ourselves for the future. 

As always @mdtauk, really good illustrations/mock-ups. @chingucoding taking all of these additions up is really amazing of you too.

chingucoding:
@robloo Thank you for engaging and raising your concerns! After all we could have missed something and that could have definitely been detrimental!

mdtauk:
So what are the next steps and will this need to wait for WinUI 3.0 or could it make it to 2.5 or 2.6?

chingucoding:
> So what are the next steps and will this need to wait for WinUI 3.0 or could it make it to 2.5 or 2.6?

When the API review is ready to go, and the team has accepted it. We (or I) can start working on this as the code for the ColorPicker is already open source.

chingucoding:
Closed as this PR is being superseded by https://github.com/microsoft/microsoft-ui-xaml/pull/545.

PR #54, opened 2019-11-25
TreeViewDragItemsCompletedEventArgs.NewParent spec

kaiguo:


active/TreeView/TreeViewDragItemsCompletedNewParent.md
```diff
@@ -0,0 +1,43 @@
...
+
+TreeView has built-in support for drag&drop reorder but currently there is no easy way to figure out where the item is being dropped. 
+
+# Description
+
+TreeView's data usually has hierarchical structure by nature and "index" doesn't work very well in this scenario. Parent + SubIndex is a more accurate way to locate items in TreeView. 
+
+A `NewParent` property in TreeViewDragItemsCompletedEventArgs should be sufficient to represent the drop location. To get SubIndex, developers can do NewParent.Children.IndexOf().
+
+# Examples
```
MikeHillberg: This example isn't very informative, especially with NewParent being un-typed. Could you add some detail? Suggestion: assume the TreeView is showing a Dictionary<string,List<string>>, where the first string is team names and the second string is player names. Then in the event handler set a TextBlock.Text to "'{playerName}' has changed to the {teamName}".

active/TreeView/TreeViewDragItemsCompletedNewParent.md
```diff
@@ -0,0 +1,43 @@
...
+
+# API Details
+
+```
+[webhosthidden]
+runtimeclass TreeViewDragItemsCompletedEventArgs
+{
+    ...
+
+    Object NewParent{ get; };
```
MikeHillberg: OldParent isn't necessary?

MikeHillberg: Not clear if "parent" is an item or a container (or a node). Suggest renaming to NewParentItem

active/TreeView/TreeViewDragItemsCompletedNewParent.md
kaiguo: Sure I'll update : )

active/TreeView/TreeViewDragItemsCompletedNewParent.md
kaiguo: Jevan brought this up as well when I talked to him. Sounds like a good thing to have. Will add it.

kaiguo: Hmm, I tried adding it but then I realized you can select multiple items and drag them together, they don't necessarily share the same parent. OldParentItem is probably not applicable here?

kaiguo: Or should we make OldParent a dictionary of select items and their parents?

PR #58, opened 2020-01-17
Fixing formatting problem that occurred

adambarlow:
There was a formatting glitch that needs to be fixed.


adambarlow:
@DmitriyKomin a formatting error crept in. Does this look right to you now?

PR #67, opened 2020-02-13
RadialGradientBrush API review

jesbis:


api_review/RadialGradientBrush/RadialGradientBrush.md
```diff
@@ -0,0 +1,244 @@
...
+reason it's necessary for a caller to catch an exception from an API, call that
+out with an explanation either here or in the Examples -->
+
+## Gradient layout
+
+The gradient is drawn within an ellipse that is defined by the `EllipseCenter` and `EllipseRadius` properties. Colors for the gradient start at the center of the ellipse and end at the radius.
+
+The colors for the radial gradient are defined by color stops added to the `GradientStops` collection property. Each gradient stop specifies a color and an offset along the gradient.
+
+The gradient origin defaults to center and can be offset using the `GradientOriginOffset` property.
```
jesbis: This "OriginOffset" approach of offsetting from center is currently consistent with [CompositionRadialGradientBrush](https://docs.microsoft.com/uwp/api/windows.ui.composition.compositionradialgradientbrush), vs. WPF which instead had an "[Origin](https://docs.microsoft.com/dotnet/api/system.windows.media.radialgradientbrush.gradientorigin?view=netframework-4.8#System_Windows_Media_RadialGradientBrush_GradientOrigin)" property. 

Is it better to have closer WPF copy-pasteability or to be consistent with newer WinUI/WinRT APIs?

Felix-Dev: I prefer being consistent within WinUI/WinRT itself. 

In the beginning, may new developers to WinUI (Desktop) will probably have a rather strong background in WPF. As the WinUI developer story will mature over the coming years though, I believe more and more developers will join who don't necessarily have that strong background. They will expect and should be provided with an in itself consistent platform for the best developer experience.

MikeHillberg: Discussion:
WPF and default value is 0.5,0.5
Composition Origin is the middle and then offsets from that.
Both have 0,0 at the upper left

Seems odd for 0,0 not to be the "origin"

Summary:
In Comp you can't set the origin but you can offset the origin. (So it's an Origin _Offset_ property)
In WPF you set the origin but you can't offset it. (Origin property defaults to 0.5,0.5)

So what should the origin be, upper left like WPF or center like Comp?

Let's follow up on the reasoning behind the Comp choice of origin concept. Pending new information, let's align with WPF for consistency within the two Xaml APIs.

api_review/RadialGradientBrush/RadialGradientBrush.md
```diff
@@ -0,0 +1,184 @@
...
+
+    Windows.Foundation.Point GradientOriginOffset { get; set; };
+
+    Windows.UI.Xaml.Media.BrushMappingMode MappingMode { get; set; };
+
+    Windows.UI.Composition.CompositionColorSpace InterpolationSpace { get; set; };
+
+    Windows.UI.Xaml.Media.GradientSpreadMethod SpreadMethod { get; set; };
+
+    Windows.Foundation.Collections.IObservableVector<Windows.UI.Xaml.Media.GradientStop> GradientStops { get; };
```
MikeHillberg: Discussion:
Note that this is a get-only IVector property, whereas the existing [GradientBrush.GradientStops](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.Media.GradientBrush.GradientStops) is a read/write concrete type. The advantage of a concrete type is that it can be created in markup and can be data bound and shared.

But with a read-only property you can still set the stops in markup, you just can't for example define a stop collection in a ResourceDictionary and then use it in multiple places on the page.

The existing GradientStopCollection doesn't raise change notifications (no public API), so implementing this would be difficult unless we created a new collection type.

The read-only property is less work and looks good enough in terms of scenarios we can think of. We can always add a setter later.

api_review/RadialGradientBrush/RadialGradientBrush.md
```diff
@@ -0,0 +1,184 @@
...
+
+## Downlevel fallback behavior
+
+When this brush is used on pre-1903 versions of Windows, where [CompositionRadialGradientBrush](https://docs.microsoft.com/uwp/api/Windows.UI.Composition.CompositionRadialGradientBrush) isn't available, the options for fallback behavior are:
+
+1. Draw a gradient using another renderer, e.g. Direct2D
+   * This would require loading additional expensive dependencies which WinUI doesn't currently rely on.
+2. Draw a custom gradient surface, e.g. using a CompositionSurfaceBrush or WriteableBitmap
+   * This would require largely reimplementing the complex CompositionRadialGradientBrush.
+3. **Rely on FallbackColor**
```
MikeHillberg: Recommend: Make it more clear that this is the choice that was made.

api_review/RadialGradientBrush/RadialGradientBrush.md
```diff
@@ -0,0 +1,184 @@
...
+
+## Composition brush usage
+
+The Xaml RadialGradientBrush will inherit from [XamlCompositionBrushBase](https://docs.microsoft.com/uwp/api/windows.ui.xaml.media.xamlcompositionbrushbase) and use a [CompositionRadialGradientBrush](https://docs.microsoft.com/uwp/api/windows.ui.composition.compositionradialgradientbrush), which is supported on Windows 10 1903 (v10.0.18362+). This seems like a useful model for new brushes going forward for both WinUI 2 and 3, and we aren't too concerned about duplicating some properties from other candidate base classes [GradientBrush](https://docs.microsoft.com/uwp/api/windows.ui.xaml.media.gradientbrush).
+
+The RadialGradientBrush API will more closely mirror the behavior of the Composition API than the WPF API: for example, the gradient origin defaults to center and provides a `GradientOrigin` property to specify an offset rather than providing a `GradientOrigin` property.
```
ranjeshj, line 18: This sentence is confusing. Don't we match WPF behavior now ? WPF's origin defaults to center and the same property exists going from 0-1 with 0,0 as the top left.

api_review/RadialGradientBrush/RadialGradientBrush.md
```diff
@@ -0,0 +1,184 @@
...
+
+```XAML
+<Page
+  xmlns:media="using:Microsoft.UI.Xaml.Media">
+
+  <Rectangle Width="200" Height="200">
+      <Rectangle.Fill>
+          <media:RadialGradientBrush
+            MappingMode="Absolute"
+            EllipseCenter="0.5,0.5"
```
ranjeshj, line 96: This is not accurate. When in absolute mapping mode, In order to get results that match the picture above, the values will need to be 
EllipseCenter="50,50"
EllipseRadius="100,100"
GradientOrigin="50,0"

api_review/RadialGradientBrush/RadialGradientBrush.md
```diff
@@ -0,0 +1,184 @@
...
+
+# API Notes
+
+## Fields
+
+| | |
+| - | - |
+| EllipseCenter | The center of the ellipse that contains the gradient. The default is `(0.5, 0.5)`. |
+| EllipseRadius | The radius of the ellipse that contains the gradient. The default is `(0.5, 0.5)`. |
+| GradientOrigin | The gradient origin's offset from the center of the element. |
```
chingucoding, line 143: The behavior now is starting from top left and the default is updated to be the center.

```suggestion
| GradientOrigin | The gradient origin's offset from the top left the element. The default is `(0.5, 0.5)` |
```

api_review/RadialGradientBrush/RadialGradientBrush.md
MikeHillberg: I missed that in the conversion, I updated it now.

api_review/RadialGradientBrush/RadialGradientBrush.md
MikeHillberg: I got confused about mapping modes during the conversion, I've updated it. But shouldn't the GradientOrigin for this picture be 100,50?

api_review/RadialGradientBrush/RadialGradientBrush.md
```diff
@@ -110,15 +111,15 @@ This example creates a radial gradient that uses Absolute mapping mode with cust
...
+The gradient is drawn within an ellipse that is defined by the `Center`, `RadiusX`, and `RadiusY` properties. Colors for the gradient start at the center of the ellipse and end at the radius.
 
 The colors for the radial gradient are defined by color stops added to the `GradientStops` collection property. Each gradient stop specifies a color and an offset along the gradient.
 
 The gradient origin defaults to center and can be offset using the `GradientOrigin` property.
 
-`MappingMode` defines whether `Center`, `EllipseRadius` and `GradientOrigin` represent relative or absolute coordinates.
+`MappingMode` defines whether `Center`, `RadiusX`, `RadiusY`, and `GradientOrigin` represent relative or absolute coordinates.
 
-When `MappingMode` is set to `RelativeToBoundingBox`, the X and Y values of the three properties are treated as relative to the brush's rendered bounds, where `(0,0)` represents the top left and `(1,1)` represents the bottom right of the brush's rendered bounds for the `Center` and `EllipseRadius` properties and `(0,0)` represents the center for the `GradientOrigin` property.
```
ranjeshj: We should say the element bounds instead of the brush's rendered bounds. That is the current behavior (same as WPF). We say it here and a couple more times in the document.

SavoySchuler: Thanks, fixed!

api_review/RadialGradientBrush/RadialGradientBrush.md
```diff
@@ -0,0 +1,189 @@
...
+[contentproperty("GradientStops")]
+unsealed runtimeclass RadialGradientBrush : Windows.UI.Xaml.Media.XamlCompositionBrushBase
+{
+    RadialGradientBrush();
+
+    Windows.Foundation.Point Center { get; set; };
+
+    Windows.Foundation.Point RadiusX { get; set; };
+    
+    Windows.Foundation.Point RadiusY { get; set; };
```
MikeHillberg, line 164: Radius properties should be type Double

PR #69, opened 2020-03-02
Hierarchical NavigationView API review

YuliKl:


active/NavigationView/HierarchicalNav/HierarchicalNavigationView.md
```diff
@@ -0,0 +1,265 @@
...
+
+## Add items in markup
+Declare app navigation hierarchy in markup.
+
+```Xaml
+<muxc:NavigationView>
+    <muxc:NavigationView.MenuItems>
+        <muxc:NavigationViewItem Content="Home" Icon="Home" ToolTipService.ToolTip="Home"/>
+        <muxc:NavigationViewItem Content="Collections" Icon="Keyboard" ToolTipService.ToolTip="Collections">
+            <muxc:NavigationViewItem.MenuItems>
```
ranjeshj: >MenuItems [](start = 37, length = 9)

Can we make MenuItems the content property of NavigationViewItem ? That way you do not require this line in  markup.

active/NavigationView/HierarchicalNav/HierarchicalNavigationView.md
```diff
@@ -0,0 +1,265 @@
...
+    </muxc:NavigationView.MenuItems>
+</muxc:NavigationView>
+```
+
+## Add items using data binding
+
+```xaml
+<DataTemplate x:Key="NavigationViewMenuItem" x:DataType="local:Category">
+    <muxc:NavigationViewItem Content="{x:Bind Name}" MenuItemsSource="{x:Bind Children}" 
+        SelectsOnInvoked="{x:Bind IsLeaf}"/>
```
ranjeshj: >IsLeaf [](start = 34, length = 6)

Do we require IsLeaf on the data object in order to support this ?

active/NavigationView/HierarchicalNav/HierarchicalNavigationView.md
```diff
@@ -0,0 +1,265 @@
...
+</DataTemplate>
+
+<muxc:NavigationView x:Name="navview" MenuItemsSource="{x:Bind categories, Mode=OneWay}" 
+    MenuItemTemplate="{StaticResource NavigationViewMenuItem}" ItemInvoked="{x:Bind ClickedItem}"
```
ranjeshj: nit: Can we name the handlers as OnItemInvoked, OnItemExpanding and OnItemCollapsed

active/NavigationView/HierarchicalNav/HierarchicalNavigationView.md
```diff
@@ -0,0 +1,265 @@
...
+        public ObservableCollection<Category> Children { get; set; }
+        public bool IsLeaf { get; set; }
+
+        public Category(String name, String icon, ObservableCollection<Category> children, bool isLeaf)
+        {
+            this.Name = name;
+            this.Icon = icon;
+            this.Children = children;
+            this.IsLeaf = isLeaf;
+        }
```
ranjeshj: can probably remove this and just use initialization syntax. especially because we end up passing nulls in several calls below.

ranjeshj: TreeView has similar events with Item and Node in it. Should we do Item and Container ?

active/NavigationView/HierarchicalNav/HierarchicalNavigationView.md
```diff
@@ -0,0 +1,265 @@
...
+[WUXC_VERSION_RS3]
+[webhosthidden]
+[WUXC_INTERFACE_NAME("INavigationViewItemBase", edf04eb1-37d1-471f-8570-3829ee5b2bc6)]
+[WUXC_CONSTRUCTOR_NAME("INavigationViewItemBaseFactory", eb014cef-7890-4ebb-8245-02e8510f321d)]
+[default_interface]
+unsealed runtimeclass NavigationViewItemBase : Windows.UI.Xaml.Controls.ContentControl
+{
+    [WUXC_VERSION_PREVIEW]
+    {
+        Boolean IsSelected{ get; set; };
```
ranjeshj: Header and Separator derive from NavigationViewItemBase, but IsSelected might not make sense for those derivations.

active/NavigationView/HierarchicalNav/HierarchicalNavigationView.md
```diff
@@ -0,0 +1,265 @@
...
+{
+...
+    [WUXC_VERSION_PREVIEW]
+    {
+        [MUX_DEFAULT_VALUE("false")]
+        Boolean IsExpanded{ get; set; };
+        [MUX_DEFAULT_VALUE("false")]
+        Boolean HasUnrealizedChildren{ get; set; };
+        [MUX_DEFAULT_VALUE("false")]
+        Boolean IsChildSelected{ get; };
```
ranjeshj: do we mean child or descendent ?

active/NavigationView/HierarchicalNav/HierarchicalNavigationView.md
```diff
@@ -0,0 +1,265 @@
...
+
+        [MUX_PROPERTY_CHANGED_CALLBACK(TRUE)]
+        Windows.Foundation.Collections.IVector<Object> MenuItems{ get; };
```
ranjeshj: we could make this content property for NavigationViewItem

active/NavigationView/HierarchicalNav/HierarchicalNavigationView.md
```diff
@@ -0,0 +1,265 @@
...
+When providing users with a hierarchical tree of navigation options, you may choose to make parent items non-selectable, for example when your app doesn't have a destination page associated with that parent item.
+To prevent an item from showing the selection indicator when invoked, set its [SelectsOnInvoked](https://docs.microsoft.com/en-us/uwp/api/microsoft.ui.xaml.controls.navigationviewitem.selectsoninvoked?view=winui-2.3) property to False.
+
+Selected items will draw their selection indicators along their left edge when in left mode or their bottom edge when in top mode. 
+The selected item may not always remain visible.
+For example, the selected item may be a child node inside a non-expanded subtree.
+In this situation, the first visible ancestor of the selected item will show as selected, and the selection indicator will move as users expand the subtree. 
+The entire navigation view will show no more than one selection indicator.
+
+## Keyboarding
```
chigy: Is this going to use TreeView as a source or is this implementing the TreeView's behavior but is a copy? I ask because we did have a tons of discussion about how to support gamepad with TreeView and I'm wondering if this is something you should also keep in mind and support here?

YuliKl: It's a copy. I admit I had not given gamepad any thought...

active/NavigationView/HierarchicalNav/HierarchicalNavigationView.md
YuliKl: oops...this is left over from early days, I need to update

active/NavigationView/HierarchicalNav/HierarchicalNavigationView.md
ranjeshj: It is not using TreeView and it is not a copy either. It is a completely different implementation using ItemsRepeater compared to TreeView which uses ListView with a flattened tree.

active/NavigationView/HierarchicalNav/HierarchicalNavigationView.md
ojhad: This might be a bit confusing since a NavigationViewItem has a content presenter as well.

active/NavigationView/HierarchicalNav/HierarchicalNavigationView.md
ojhad: I think this was taken from the test page of my PR. This is not required.

YuliKl: Dilip is correct, that's exactly where I copied the code from. I'll remove it from here and instead add a sample showing how to use SelectsOnInvoked.

active/NavigationView/HierarchicalNav/HierarchicalNavigationView.md
YuliKl: Renamed.

active/NavigationView/HierarchicalNav/HierarchicalNavigationView.md
YuliKl: fixed

active/NavigationView/HierarchicalNav/HierarchicalNavigationView.md
YuliKl: As Dilip mentioned earlier, NavigationViewItem already has a Content property because it derives from ContentControl. Today, this property is already used to set the text label, or other content, to the right of the icon.

active/NavigationView/HierarchicalNav/HierarchicalNavigationView.md
YuliKl: Sorry for being vague, I meant the behavior is being copied. 

The only new concept we're introducing is the use of left/right arrow keys (which do nothing in NavigationView today). Those are present on the controller as well. What kind of differences between controller and keyboard do I really need to think about in this case?

chigy: There's quite a bit of work that happened. Unfortunately Premalini has the details and I don't want to be found quoting on anything wrong, but for example, we use A/B buttons and not directional keys to open/collapse the items (I think... it's been such a long time and went back and forth...)

active/NavigationView/HierarchicalNav/HierarchicalNavigationView.md
YuliKl: @ojhad is this still true after the rewrite with ItemsRepeater?
> it derives from ContentControl

<br/>
(Are NavigationViewItemBase objects still ListViewItems?)

YuliKl: Found the answer in the idl. NavigationViewItemBase now derives directly from ContentControl and not ListViewItem.

active/NavigationView/HierarchicalNav/HierarchicalNavigationView.md
YuliKl: @predavid can you help fill in the details?
I would expect the A button to invoke the item rather than expand.

active/NavigationView/HierarchicalNav/HierarchicalNavigationView.md
```diff
@@ -0,0 +1,272 @@
...
+[WUXC_INTERFACE_NAME("INavigationViewItem", 8614be0f-b7b6-4851-960a-f5e3f69f624a)]
+[WUXC_STATIC_NAME("INavigationViewItemStatics", 803c0081-fda5-4b90-aace-3f2306dbe5c4)]
+[WUXC_CONSTRUCTOR_NAME("INavigationViewItemFactory", 973bdb4a-7e08-4f76-923c-f12bd685e86e)]
+unsealed runtimeclass NavigationViewItem : NavigationViewItemBase
+{
+...
+    [WUXC_VERSION_PREVIEW]
+    {
+        [MUX_DEFAULT_VALUE("false")]
+        Boolean IsExpanded{ get; set; };
```
MikeHillberg: This is silently ignored if this has no children?

active/NavigationView/HierarchicalNav/HierarchicalNavigationView.md
MikeHillberg: So that's still a problem for [ContentProperty].

And if this is a breaking change, don't we have to call this 3.0? (And then DCPP becomes 4.0?)

active/NavigationView/HierarchicalNav/HierarchicalNavigationView.md
YuliKl: I think it should return false in all circumstances other than when it's a NavigationViewItem with MenuItems and those items are visible because it's expanded. (this property will also be available on NavigationViewItemSeparator and NavigationViewItemHeader, which inherit from NVIB.)

active/NavigationView/HierarchicalNav/HierarchicalNavigationView.md
YuliKl: I don't think this is a breaking change, I think we want to add a new MenuItems property. My only question is whether MenuItems should be added to NavigationViewItemBase or to NavigationViewItem?

active/NavigationView/HierarchicalNav/HierarchicalNavigationView.md
MikeHillberg: Descendant is definitely a useful feature

active/NavigationView/HierarchicalNav/HierarchicalNavigationView.md
MikeHillberg: It's not unheard of, but it's unusual to set something true and have it read back false. Having this property on the separator/header is odd, and doesn't seem necessary since this property isn't on the Base?

active/NavigationView/HierarchicalNav/HierarchicalNavigationView.md
ranjeshj: We could add this when needed.

active/NavigationView/HierarchicalNav/HierarchicalNavigationView.md
YuliKl: Resolving this conversation because I believe the sample code is correct and we are not able to follow the suggestion.

active/NavigationView/HierarchicalNav/HierarchicalNavigationView.md
YuliKl: True. But not adding this property would constitute a breaking change. In the past, NVIB had derived from ListViewItem, which derives from SelectorItem and has the IsSelected property.

active/NavigationView/HierarchicalNav/HierarchicalNavigationView.md
YuliKl: Oh, I misread the parent class, headers and separators won't have IsExpanded.

So yes, I expect setting IsExpanded to true on a leaf node to get ignored.

active/NavigationView/HierarchicalNav/HierarchicalNavigationView.md
YuliKl: I think we mean descendent but are using the term "child" for brevity. @ojhad can you confirm?

active/NavigationView/HierarchicalNav/HierarchicalNavigationView.md
ranjeshj: I see. This is not a breaking change. The MenuItems is a new property.

active/NavigationView/HierarchicalNav/HierarchicalNavigationView.md
ranjeshj: @ojhad Can you provide details on the gamepad behavior that you currently have ? @predavid can you comment if the current behavior matches your expectation (relative to TreeView) ?

active/NavigationView/HierarchicalNav/HierarchicalNavigationView.md
ranjeshj: I would expect setting IsExpanded to true/false will expand or collapse the item if it contains any children, do nothing otherwise. The last written value of the property should be what is read back.

active/NavigationView/HierarchicalNav/HierarchicalNavigationView.md
MikeHillberg: It's already there: [NavigationView.MenuItems](https://docs.microsoft.com/uwp/api/Microsoft.UI.Xaml.Controls.NavigationView.MenuItems)

ranjeshj: Are we talking about different propertes ? I was talking about the new MenuItems property we are adding on NavigationViewItem (not NavigationView, which already exisits) to get the nesting support.

active/NavigationView/HierarchicalNav/HierarchicalNavigationView.md
ojhad: Yep, you are correct, we mean descendant in this case.

active/NavigationView/HierarchicalNav/HierarchicalNavigationView.md
ranjeshj: Ok to keep it since moving it is a breaking change.

active/NavigationView/HierarchicalNav/HierarchicalNavigationView.md
ranjeshj: ok. That makes sense. I'm not sure about precedents - but perhaps we can update the naming to reflect that.

active/NavigationView/HierarchicalNav/HierarchicalNavigationView.md
```diff
@@ -0,0 +1,345 @@
...
+                <muxc:NavigationViewItem Content="Bookshelf" Icon="Library" 
+                    ToolTipService.ToolTip="Bookshelf"/>
+                <muxc:NavigationViewItem Content="Mail" Icon="Mail" ToolTipService.ToolTip="Mail"/>
+            </muxc:NavigationViewItem.MenuItems>
+        </muxc:NavigationViewItem>
+    </muxc:NavigationView.MenuItems>
+</muxc:NavigationView>
+```
+
+## Add items using data binding
```
MikeHillberg:
```suggestion
## Add a hierarchy of items using data binding

Add a hierarchy of menu items to the NavigationView by 
* binding the MenuItemsSource property to the hierarchical data
* defining the item template to be a NavigationViewMenuItem, with its Content set to be the label of the menu item, and its its MenuItemsSource property bound to the next level of the hierarchy

This example also demonstrates the Expanding and Collapsing events. These events are raised for a menu item with children.
```

active/NavigationView/HierarchicalNav/HierarchicalNavigationView.md
ojhad: Currently, with the existing gamepad behavior, in order to open the flyout you need to invoke the item using the A button. Directional keys are just used for x/y navigation.

active/NavigationView/HierarchicalNav/HierarchicalNavigationView.md
```diff
@@ -0,0 +1,351 @@
...
+section.  For example, see the Remarks for the MediaPlayerElement 
+(https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.Controls.MediaPlayerElement#remarks). -->
+
+
+# API Notes
+<!-- Give a one or two line description of each API (type
+and member), or at least the ones that aren't obvious
+from their name.  These descriptions are what show up
+in IntelliSense. -->
+
```
MikeHillberg:
```Suggestion
| API | Description | Notes |
| - | - | - | 
| MenuItems property | Gets the collection of menu items displayed as children of the the NavigationMenuItem. | |
| MenuItemsSource property | Gets or sets an object source used to generate the children of the NavigationViewItem | |
| IsChildSelected property | A descendant item is selected | |
| IsExpanded property | Gets or sets a value that indicates whether a tree node is expanded. Ignored if there are no menu items. | Analogous to [TreeViewItem.IsExpanded](https://docs.microsoft.com/uwp/api/Microsoft.UI.Xaml.Controls.TreeViewItem.IsExpanded) |
| HasUnrealizedChildren property | Gets or sets a value that indicates whether the current item has child items that haven't been shown. | Analogous to [TreeViewItem.HasUnrealizedChildren](https://docs.microsoft.com/uwp/api/Microsoft.UI.Xaml.Controls.TreeViewItem.HasUnrealizedChildren) |
| Expand method |  Expands the specified node in the tree. | Analogous to [TreeView.Expand](https://docs.microsoft.com/uwp/api/Microsoft.UI.Xaml.Controls.TreeView.Expand) |
| Collapse method |  Collapses the specified node in the tree. | Analogous to [TreeView.Collapse](https://docs.microsoft.com/uwp/api/Microsoft.UI.Xaml.Controls.TreeView.Collapse) |
| Expanding event | Occurs when a node in the tree starts to expand. | Analogous to [TreeView.Expanding](https://docs.microsoft.com/uwp/api/Microsoft.UI.Xaml.Controls.TreeView.Expanding) |
| Collapsed event | Occurs when a node in the tree is collapsed. |  Analogous to [TreeView.Collapsed](https://docs.microsoft.com/uwp/api/Microsoft.UI.Xaml.Controls.TreeView.Collapsed) |
```

active/NavigationView/HierarchicalNav/HierarchicalNavigationView.md
```diff
@@ -0,0 +1,351 @@
...
+# Background
+The Xaml [NavigationView](https://docs.microsoft.com/uwp/api/Microsoft.UI.Xaml.Controls.NavigationView) control enables navigation of app content. It has a header, a view for the main content, and a menu pane for navigation commands. There is additional discussion about how to use this control in the [Navigation view](https://docs.microsoft.com/en-us/windows/uwp/design/controls-and-patterns/navigationview) design document.
+
+Currently, NavigationView's MenuItems list allows for displaying a flat list of items in the pane. 
+It's common for apps to want to present users with a hierarchical navigation tree. 
+This feature adds the capability to nest items within the pane.
+
+# Hierarchy
+To show a hierarchical list of nested navigation items in the pane, use NavigationViewItem's `MenuItems` property.
```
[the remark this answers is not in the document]
ranjeshj: NavigationView.MenuItems shows the first level of menus. **NavigationViewItem**.MenuItems allows the nested case. I think the initial text was talking about the nested case which is lost with the update suggestion.

PR #70, opened 2020-03-06
[WIP] TabView v2

stmoy:
Spec corresponding to: https://github.com/microsoft/microsoft-ui-xaml/issues/2007#


active/TabView/TabView_v2.md
```diff
@@ -0,0 +1,163 @@
...
+# Description
+<!-- Use this section to provide a brief description of the feature.
+For an example, see the introduction to the PasswordBox control 
+(http://docs.microsoft.com/windows/uwp/design/controls-and-patterns/password-box). -->
+
+## TabViewWidthMode: Compact
+
+This change introduces a new value to the existing `TabViewWidthMode` enum: `Compact`.
+
+Setting `Compact` will cause unselected tab headers to display only the tab's icon. If an icon is not set, the tab will display nothing.
```
chingucoding: I am not sure if the wording here is clear enough. We do not display an icon however we (should) still allocate the space and act like one was specified. Otherwise the TabViewItem is quite small and may be hard to notice or click.

active/TabView/TabView_v2.md
```diff
@@ -0,0 +1,163 @@
...
+<!-- For example, this section is a place to explain why you're adding this API rather than
+modifying an existing API. -->
+
+<!-- For example, this is a place to provide a brief explanation of some dependent
+area, just explanation enough to understand this new API, rather than telling
+the reader "go read 100 pages of background information posted at ...". -->
+
+> This spec corresponds to [issue 2007](https://github.com/microsoft/microsoft-ui-xaml/issues/2007) on the WinUI repo.
+
+The WinUI TabView is missing two main features relative to the Windows Community Toolkit (WCT) TabView:
```
MikeHillberg:
```suggestion
The WinUI [TabView](https://docs.microsoft.com/uwp/api/Microsoft.UI.Xaml.Controls.TabView) control is a way to display a set of tabs and their respective content. There's also a TabView control in the [Windows Community Toolkit](https://docs.microsoft.com/en-us/windows/communitytoolkit/controls/tabview), which has two main features not found in the WinUI version.:
```

active/TabView/TabView_v2.md
```diff
@@ -0,0 +1,163 @@
...
+* **TabWidthMode: Compact** ([WCT API Link](https://docs.microsoft.com/en-us/dotnet/api/microsoft.toolkit.uwp.ui.controls.tabwidthmode?view=win-comm-toolkit-dotnet-stable)) 
+* **Overlay close button** ([WCT API Link](https://docs.microsoft.com/en-us/dotnet/api/microsoft.toolkit.uwp.ui.controls.tabview.isclosebuttonoverlay?view=win-comm-toolkit-dotnet-stable#Microsoft_Toolkit_Uwp_UI_Controls_TabView_IsCloseButtonOverlay))
+
+Once these feature gaps have been closed and the WinUI TabView is at parity with the Windows Community Toolkit version, we will be able to deprecate the WCT TabView control.
+
+# Description
```
MikeHillberg: The content here should be in API Notes (this is the information that should show up on the API pages). I don't think we need a Description section in this particular spec.

active/TabView/TabView_v2.md
```diff
@@ -0,0 +1,163 @@
...
+
+![Compact](./TabView_Width_Compact.png)
+
+## CloseButtonOverlayMode
+
+Describes the behavior of the x-to-close button found on each TabViewItem. 
+
+The `CloseButtonOverlayMode` enum has three values: {`Auto`, `OnHover`, `Persistent`}
+
+> The `CloseButtonOverlayMode` will only effect tabs that are closeable - ie. the value of the `TabViewItem`'s `IsClosable` property is `TRUE`.
```
MikeHillberg:
```suggestion
> The `CloseButtonOverlayMode` will only affect tabs that are closeable - ie. the value of the `TabViewItem`'s `IsClosable` property is `TRUE`.
```

active/TabView/TabView_v2.md
```diff
@@ -0,0 +1,163 @@
...
+
+`OnHover`: The x-to-close button only appears on the selected tab and any tabs that are being hovered.
```
MikeHillberg: The name is "hover", but the behavior shown in the animation is "over". (Hover means that you have to hold the mouse still for a short time before anything happens.) I think the intended behavior is to match that of [ButtonBase.IsPointerOver](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.Controls.Primitives.ButtonBase.IsPointerOver), so maybe call this OnPointerOver?

MikeHillberg: How is touch supported?

active/TabView/TabView_v2.md
```diff
@@ -0,0 +1,163 @@
...
+
+![Hover](./TabView_Close_Hover.gif)
+
+`Persistent`: The x-to-close button always appears on every tab.
```
MikeHillberg: Very subjective opinion, but "persistent" is an unusual word in the rest of the API set. I think this is more of an "Always", like [NavigationViewShoulderNavigationEnabled.Always](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.Controls.NavigationViewShoulderNavigationEnabled#fields).

active/TabView/TabView_v2.md
```diff
@@ -0,0 +1,163 @@
...
+# Appendix
+<!-- Anything else that you want to write down for posterity, but 
+that isn't necessary to understand the purpose and usage of the API.
+For example, implementation details. -->
+
+**Why is CloseButtonOverlayMode an enum and not a boolean?**
+
+1. The corresponding visual effect might change over time
+
+The CloseButtonOverlayMode enum describes a visual effect that is likely to change over time. For example, Spartan Edge used the hover model, whereas chromium Ege uses a persistent model. An enum allows that platform to have an opinion (via `Auto`) and an option to change the behavior over time in a way that a boolean does not.
```
MikeHillberg:
```suggestion
The CloseButtonOverlayMode enum describes a visual effect that is likely to change over time. For example, Spartan Edge used the hover model, whereas chromium Edge uses a persistent model. An enum allows that platform to have an opinion (via `Auto`) and an option to change the behavior over time in a way that a boolean does not. 
```

active/TabView/TabView_v2.md
```diff
@@ -0,0 +1,163 @@
...
+
+`OnHover`: The x-to-close button only appears on the selected tab and any tabs that are being hovered.
+
+![Hover](./TabView_Close_Hover.gif)
+
+`Persistent`: The x-to-close button always appears on every tab. 
+
+![Persistent](./TabView_Close_Persistent.png)
+
+`Auto`: Maps to `Persistent`.
```
MikeHillberg:
```suggestion
`Auto`: Currently maps to `Persistent`.
```

active/TabView/TabView_v2.md
```diff
@@ -0,0 +1,163 @@
...
+<!-- The exact API, in MIDL3 format (https://docs.microsoft.com/en-us/uwp/midl-3/) -->
+
+```
+[WUXC_VERSION_MUXONLY]
+[webhosthidden]
+enum TabViewWidthMode
+{
+    Equal = 0,
+    SizeToContent = 1,
+    Compact = 2,
```
MikeHillberg: The new field (Compact) should be marked preview. And for the purposes of this documentation, we should do something to indicate that the first two fields exist already.

active/TabView/TabView_v2.md
stmoy: In this mode, the selected tab can be closed since the x-to-close button always shows on the selected tab. To close an unselected tab, the user must either select the tab, or open the context menu on the unselected tab and choose the close option (assuming the developer followed the guidance).

active/TabView/TabView_v2.md
stmoy: CC: @chingucoding

active/TabView/TabView_v2.md
```diff
@@ -0,0 +1,163 @@
...
+};
+
+[WUXC_VERSION_PREVIEW]
+[webhosthidden]
+enum TabViewCloseButtonOverlayMode
+{
+    Auto = 0,
+    OnHover = 1,
```
stmoy: Since Auto currently maps to Persistent, should we swap the order of OnHover vs. Persistent?

MikeHillberg: I think it's OK either way. The only API guideline we have is that the default case be the zero value.

active/TabView/TabView_v2.md
chingucoding: Thanks for CCing!  I am not quite sure if there is a way to mark only one enum value as preview, I haven't found a comparable example in the project. 🤔

MikeHillberg: Actually maybe that's not necessary, I don't think we've been versioning the changes in MUXC. @ranjeshj?

ranjeshj: No. There is no versioning from what I understand, just preview (pre-release) or stable. I don't think the version is necessary since we go down-level. @jevansaks to correct me if I am wrong here.

jevansaks: Versioning would work here and you might want to if this enum value might stay under preview even through the next major release. Otherwise I agree it probably doesn't matter.

chingucoding: So should we just add it to the normal enum without marking that entry as preview?

active/TabView/TabView_v2.md
```diff
@@ -0,0 +1,163 @@
...
+    Compact = 2,
+};
+
+[WUXC_VERSION_PREVIEW]
+[webhosthidden]
+enum TabViewCloseButtonOverlayMode
+{
+    Auto = 0,
+    OnPointerOver = 1,
+    Always = 2,
```
chingucoding: Should the last value called "Always" or "Persistent"? In the WinUI proposal, it's still called "Persistent".

MikeHillberg: For naming consistency I think Always is right; there's a lot of Always in the APIs today but no Persistent.

chingucoding: Thanks for the clarification @MikeHillberg

PR #71, opened 2020-03-16
Create Window_and_Application_API_Spec.md

marb2000:
Adding draft of Window and Application API Specs to support Win32

### DIFF
--- a/active/Win32/Window_and_Application_API_Spec.md
+++ b/active/Win32/Window_and_Application_API_Spec.md
@@ -482,9 +482,9 @@ Contains the argument returned by a window size change event.
     public bool Handled { get; set; };
     ```
 
-- WindowSizeChangedEventArgs.Visible: Gets the new size of the window in units of effective (view) pixels.
+- WindowSizeChangedEventArgs.Size: Gets the new size of the window in units of effective (view) pixels.
     ```CS
-    public Size Visible { get; } 
+    public Size Size { get; } 
     ```
 
 
--- a/active/Win32/Window_and_Application_API_Spec.md
+++ b/active/Win32/Window_and_Application_API_Spec.md
@@ -1,63 +1,70 @@
-# Window and Application API Spec
+# Window and Application API updates 
 
-This spec describes updates to the XAML Window and Application APIs to accommodate this move to support Desktop (win32) apps.
+This spec describes updates to the XAML Window and Application APIs to enable them to support Desktop apps in addition to UWP.
 
-## Spec Status
-In review. No scheduled yet for reviewing by the API Board. 
+# Background
+
+Xaml in UWP has a [Window](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.Window) class which wraps a [CoreWindow](https://docs.microsoft.com/uwp/api/Windows.UI.Core.CoreWindow), and an [Application](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.Application) class which wraps a [CoreApplication](https://docs.microsoft.com/uwp/api/Windows.ApplicationModel.Core.CoreApplication). For WinUI 3 this is being expanded to not require a CoreWindow or CoreApplication; Window can use an hwnd and Application can run a message pump. This spec has the API additions to Application and Window to support this.
+
+Note that some existing APIs will also behave differently when running as a Desktop app. For example, the static Window.Current property today returns the Window for the current (calling) thread, but in a non-UWP app it will return null. Similarly the Window.CoreWindow property will be null when not running as UWP.
+
+There is also one API not (yet) present in WinUI: [Window.SetTitleBar](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.Window.SetTitleBar). That API relies on a Composition feature that only exists for system Visuals and is not yet supported for WinUI Visuals.
 
 
-## Execution Notes
+# Examples and API Notes
 
-This API spec contains the the APIs under the following criteria:
-1. Already exists in Window.UI.Xaml.Window. 
-2. Are part of the Mininum Viable Product
+## Window class
 
-There are other proposed APIs proposed that will be part of the future phases. 
-> A Phase doesn't correspond with a product version, it's just a set of Sprints with allocated resources to execute this work.
+Window represents a WinUI application window. It can be used in a [UWP](https://docs.microsoft.com/en-us/windows/uwp/get-started/universal-application-platform-guide) app or a Desktop app. When run in a UWP app there can only be one instance on a thread.
 
-##   Design Agreement
-- The new Window implementation reuses the current Windows.UI.XAML.Window class and it will be placed under Microsoft.UI.XAML.Window. 
-- Current methods, events, and properties should still work for UWP and add the new Win32 support. 
-- These APIs use different code paths depending on whether WinUI 3 is running on Win32 or UWP. 
-- There are some APIs that don’t work on Win32 or UWP yet in this phase. Those APIs could return **null** or **throw** an exception depending on the case. 
+### Examples
 
-This agreement applies to the Application class too.
+> Spec note: The only new API in this set of examples is the Window constructor. Shown here is the general description for the class, so that we can see it in full context. But none of the behavior for UWP apps is changed.
 
-## Window Class
-Namespace: Microsoft.UI.Xaml
+Set the content of a given window and give it keyboard focus.
 
 ```CS
-public class Window : IClosable
+void InitializeAndActivateWindow(Window window)
+{
+    window.Content = new TextBlock() { Text = "Hello" };
+    window.Activate();
+}
 ```
 
-### **Window management, appearance, and behaviors**
+In a UWP app the read-only [Window.Current](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.Window.Current) property returns the Window for the current thread. (A thread has at most one Window and it's created automatically on each UI thread.) This property returns null on a non UI thread or in a Desktop app.
 
-### Window() constructor **[NEW]** 
-The constructor initializes the Window object. Windows are created in the same thread where they are instantiated. Frequently, this thread is the UI Thread.
+The following sets the content of the calling thread's Window in a UWP app and gives it keyboard focus.
 
 ```CS
-public Window();
+Window.Current.Content = new TextBlock() { Text = "Hello" };
+Window.Current.Activate();
 ```
-#### Samples
-##### Code Sample
+
+In a Desktop app you create each Window, and you can create more than one Window on a thread.
+
 ```CS
 Window window = new Window();
 window.Content = new TextBlock() { Text = "Hello" };
 window.Activate(); 
 ```
-##### Markup Sample
+
+You can also define a new Window in markup:
+
+| Spec note: no <Window.Content> tag is required because Window.Content is updating to become the [ContentPropertyAttribute](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.Markup.ContentPropertyAttribute).
 
 ```XML
 <Window 
     x:Class="MainWindow"
     xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
-    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
->
-    <Grid><Button Content=”Click” Click=”OnClick”/></Grid>
+    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml">
+    <TextBlock>Hello</TextBlock>
 </Window>
 ```
+
+with corresponding code-behind:
+
 ```CS
-public partial class MainWindow:Window
+public partial class MainWindow : Window
 {
   public MainWindow()
   {
@@ -65,183 +72,105 @@ public partial class MainWindow:Window
   }
 }
 ```
-##### Multiple Widndows Sample
-Developers can create multi-window experiences with WinUI 3 in the same UI thread. 
+
+and then you can create/activate that custom window from code:
+
 ```CS
-Window w = new Window ();
-Button b = new Button() { Content = "Click Me" };
-b.Click += OnClick;
-w.Content = b;
-w.Activate();
+MainWindow window = new MainWindow();
+window.Activate();
+
 ```
+
+In a UWP app the main thread already has a Window on it, which you can retrieve using the static [Window.Current](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.Window.Current) property. You can create additional windows by creating additional [CoreApplicationViews](https://docs.microsoft.com/uwp/api/Windows.ApplicationModel.Core.CoreApplicationView), which are always created on a new thread, and which automatically create the following for the new thread: [ApplicationView](https://docs.microsoft.com/uwp/api/Windows.UI.ViewManagement.ApplicationView), [CoreWindow](https://docs.microsoft.com/uwp/api/Windows.UI.Core.CoreWindow), and [Window](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.Window).
+
+Create a new Window in a UWP app (on a new thread):
+
 ```CS
-private void OnClick(object sender, RoutedEventArgs e)
+_ = CoreApplication.CreateNewView().DispatcherQueue.TryEnqueue(() =>
 {
-   Window w2 = new Window ();
-   w2.Content = new TextBlock(){Text=”Window 2”}
-   w2.Activate();
-}
+    // This code runs on the new thread
+    Window.Current.Content = new TextBlock() { Text = "Hello" };
+    Window.Current.Activate();
+});
 ```
-##### Create a Window in a separated thread
-To create a Window in a separated thread, you should create the thread first and create a Window in this new Thread. 
+
+To create a new Window on a new thread in a Desktop app, create the thread first:
+
 ```CS
-private void Button_Click(object sender, RoutedEventArgs e)
+Thread thread = new Thread(() =>
 {
-    Thread thread = new Thread(() =>
-    {
-        Window w = new Window();
-        w.Content = new TextBlock() { Text = "Separated Thread Window" };
-        w.Activate();
-        w.Closed += (sender1, e1) => w.Dispatcher.InvokeShutdown();
+    Window window = new Window();
+    window.Content = new TextBlock() { Text = "Hello" };
+    window.Activate();
+    window.Closed += (sender1, e1) => w.Dispatcher.InvokeShutdown();
 
-        System.Windows.Threading.Dispatcher.Run();
+    System.Windows.Threading.Dispatcher.Run(); // Issue: There should be a Window.Run()?
 
-    });
-    thread.SetApartmentState(ApartmentState.STA);
-    thread.IsBackground = true; 
-    thread.Start();
-} 
+});
+thread.SetApartmentState(ApartmentState.STA);
+thread.Start(); 
 ```
 
- #### Remarks
+### Remarks
 
-- **Win32**: The constructor creates a top-level Win32 Window (HWnd). A top-level Window is a window that is not a child window or has no parent window. 
+Attempting to activate (`new`) a new Window in a UWP app will fail and log a debug message.
 
-- **UWP**: There is already one CoreWindow created and UWP only allows one CoreWindow per thread, as a consequence, the contructor and the XAML Markup throw an exception when running in UWP.
-    - Exception type: TBD
- 
-    - Exception Message: TBD. Something like "Only one Window per thread is allowed in UWP app model".
+> Implementation note: This will use [RoOriginateError](https://docs.microsoft.com/en-us/windows/win32/api/roerrorapi/nf-roerrorapi-rooriginateerror) to show an explanatory message in the debugger.
 
-### Window.Bounds property
-Gets a Rect value containing the height and width of the application window in logical pixels.
+Creating a new Window in a Desktop app creates a new top level hwnd.
 
-```CS
-public Rect Bounds { get; }
-```
-#### Property value
+## IWindowNative Interface **[NEW]**
 
-Rect: A value that reports the height and width of the window.
+This interface is implemented by Window, and in a Desktop app can be used to get the Window's underlying hwnd.
 
-#### Remarks
+> Issue: need sample here
 
-- **Win32**: The Window is created with the default size of the Window specify by the Window Shell.
-- **UWP**: Keep the current behavior. To change the size of the CoreWindow in UWP, it’s required to use ApplicationView’s PreferredLaunchViewSize and PreferredLaunchWindowingMode.
-
-### Window.Icon property **[NEW]** 
+## Window.Icon property **[NEW]**
 Gets or sets a window's icon. For each window, this icon is used in its title bar, its task bar button, and in its ALT+TAB application selection list entry.
+
+> Note: this property is ignored in a UWP app
 
 ```CS
 public ImageSource Icon { get; set; }
 ```
 
-#### Property value
-Icon: An ImageSource object that represents the icon.
-
-#### Sample
 ```XML
 <Window
   xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
   xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
   x:Class="Sample.MainWindow"
-    Icon="WinUI3Icon.ico">
+  Icon="WinUI3Icon.ico">
 </Window>
 ```
 
-#### Remarks:
-- **Win32**: The image is set as an Icon of the Window. 
-- **UWP**: Do nothing. The property value is ignored.
-
-### Window.Content property
-Gets or sets the UIElement visual root of the window. 
-
-```CS
-[ContentProperty(“Content”)]
-public UIElement Content { get; set; }
-```
-#### Property Value
-UIElement: The visual root of a window.
-
-#### Remarks
-This property has the ContentProperty attribute so the XAML designer can display the content on the designer.
-
-### Window.Visible property
-Gets a value that reports whether the window is visible. Visibility does not consider the window being behind another window and not visible. It only detects if it is on the screen. This property matches the XamlRoot’s IsHostVisible property.
-
-```CS
-public bool Visible { get; } 
-```
-
-#### Property Value
-bool: **true** if the window is visible; **false** if the window is not visible.
-
-#### Remarks
-- **Win32**: False when the Window is minimized or never displayed. 
-- **UWP**: Keep the same current UWP behavior.
 
 
-### Window.VisibilityChanged Event
-
-Occurs when the value of the Visible property changes. Also, it’s fired when the XamlRoot’s IsHostVisible changes as a consequence of cascade change of the Window’s Visibility.
-
-```CS 
-Window w = new Window();
-window.Content = new TextBlock() { Text = "Hello" };
-window.VisibilityChanged += OnVisibilityChanged;
-```	
-
-```CS
-private void OnVisibilityChanged(object sender, VisibilityChangedEventArgs e)
-{
-   if (e.Visible)
-   {
-      //
-   }
-}
-```
-#### Remarks
-VisibilityChangedEventArgs includes a Boolean Visible property that you should check to determine the visibility state when the event was fired.
-
-### Window.SetTitleBar(UIElement) Method
-Set the custom XAML content that should act as the title bar.
-```CS 
-public void SetTitleBar(UIElement); 
-```
-
-#### Parameters
-UIElement: Custom XAML content that should act as the title bar. To use multiple objects, wrap them in a container element such as one derived from Panel.
-
-#### Remarks
-
-- **Win32**: This method do nothing when running in Win32.
-- **UWP**: Continue using the OS behavior. More info [here](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.window.settitlebar) 
-
-### Window.Title property **[NEW]**
-
-Set or get a string to display on the Title. 
+## Window.Title property **[NEW]**
+Get or set a string to display on the Title. 
 
 ```CS
 public string Title { get; set; }
 ```
-#### Sample
-```XML
+
+### Example
+
 ```XML
 <Window 
     x:Class="MainWindow"
     xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
     xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
-    Title="Main Window" 
->
+    Title="Main Window" >
     
 </Window>
 ```
 
-#### Remarks
-This is not a DependencyProperty.
+### Remarks
 
-- **UWP**: This string is not displayed in the Window's Title bar when SetTitleBar method is used or set a not null content. The Title is still used in Alt+Tab scenarios,
+In a UWP app, this property is a wrapper for [ApplicationView.Title](https://docs.microsoft.com/uwp/api/Windows.UI.ViewManagement.ApplicationView.Title), which is ignored if [AppWindowTitleBar.ExtendsContentIntoTitleBar](https://docs.microsoft.com/uwp/api/Windows.UI.WindowManagement.AppWindowTitleBar.ExtendsContentIntoTitleBar) is set to true.
 
+In a Desktop app this is a wrapper for [SetWindowText](https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-setwindowtexta).
 
-### Window.Current property
+## Window.Current property
 
 Gets the window of the current thread.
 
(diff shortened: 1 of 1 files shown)


active/Win32/Window_and_Application_API_Spec.md
```diff
@@ -0,0 +1,618 @@
...
+<Window
+  xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+  xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+  x:Class="Sample.MainWindow"
+    Icon="WinUI3Icon.ico">
+</Window>
+```
+
+#### Remarks:
+- **Win32**: The image is set as an Icon of the Window.
```
Felix-Dev: What happens when 'null' is specified for the icon? Do Win32 apps fall back to the WPF behavior which is displaying a default application icon or does it use the WinForms & UWP behavior and displays no icon?

active/Win32/Window_and_Application_API_Spec.md
```diff
@@ -0,0 +1,618 @@
...
+```CS 
+public void SetTitleBar(UIElement); 
+```
+
+#### Parameters
+UIElement: Custom XAML content that should act as the title bar. To use multiple objects, wrap them in a container element such as one derived from Panel.
+
+#### Remarks
+
+- **Win32**: This method do nothing when running in Win32.
```
Felix-Dev: There has been some thinking in WinUI to add UWP's titlebar customization functionality (i.e. CoreApplicationViewTitleBar.ExtendViewIntoTitleBar) to WinUI Desktop apps too. Seeing as Window.SetTitleBar is not supported on Win32 apps, does that mean that WinUI Desktop apps will still have to re-draw the entire titlebar and re-implement titlebar (drag, minimize, maximize, close,...) logic if they want to provide a custom titlebar?

active/Win32/Window_and_Application_API_Spec.md
```diff
@@ -0,0 +1,618 @@
...
+private void OnVisibilityChanged(object sender, VisibilityChangedEventArgs e)
+{
+   if (e.Visible)
+   {
+      //
+   }
+}
+```
+#### Remarks
+VisibilityChangedEventArgs includes a Boolean Visible property that you should check to determine the visibility state when the event was fired.
```
dotMorten: I'd like to see this entire visibility api moved to UIElement instead. See issue logged here with justifications:
https://github.com/microsoft/microsoft-ui-xaml/issues/674

MikeHillberg: Obvious overlap, but these look like two separate issues. I agree we should have a UIElement.IsVisible property like WPF's, but the Window.Visible property already exists, and Window isn't a UIElement. So we can have both Window.Visible and UIElement.IsVisible.

dotMorten:
You `Activate` a Window to view it, but you `Close` it to remove it. That seems rather inconsistent. Should `Close` be `Deactiveate` or should `Activate` be `Open` ?

Also it's not clear to me whether `IsVisible` is true or false when the window is minimized. Should it be `IsOpen` is `IsActivated` ?

active/Win32/Window_and_Application_API_Spec.md
```diff
@@ -0,0 +1,618 @@
...
+
+- **Win32**: The constructor creates a top-level Win32 Window (HWnd). A top-level Window is a window that is not a child window or has no parent window. 
+
+- **UWP**: There is already one CoreWindow created and UWP only allows one CoreWindow per thread, as a consequence, the contructor and the XAML Markup throw an exception when running in UWP.
+    - Exception type: TBD
+ 
+    - Exception Message: TBD. Something like "Only one Window per thread is allowed in UWP app model".
+
+### Window.Bounds property
+Gets a Rect value containing the height and width of the application window in logical pixels.
```
dotMorten: A `Rect` also provides location.  Is this the location on the current screen? Across all screens? Which screen is it on?
If it truly only provides height and width, this should be a `Size` type

active/Win32/Window_and_Application_API_Spec.md
```diff
@@ -0,0 +1,618 @@
...
+Namespace: Microsoft.UI.Xaml
+
+Contains the set of arguments returned to an app after a window input or behavior event.
+
+### Properties
+
+- **WindowEventArgs.Handled**: Gets or sets whether the event was handled. **true** if the event has been handled by the appropriate delegate; **false** if it has not.
+    ```CS
+    public bool Handled { get; set; }
+    ```
```
dotMorten: What's the point of handling this? Will that prevent it from resizing?

active/Win32/Window_and_Application_API_Spec.md
```diff
@@ -0,0 +1,618 @@
...
+
+```CS
+public class WindowActivatedEventArgs
+```
+
+### Properties
+- **WindowActivatedEventArgs.Handled**: Gets or sets whether the window activation changed event was handled. **true** if the event has been handled by the appropriate delegate; **false** if it has not.
+    ```CS
+    public bool Handled { get; set; }
+    ```
```
dotMorten: What's the point of handling this? Will that prevent it from getting activated? If I don't handle it, what will it bubble up to?

active/Win32/Window_and_Application_API_Spec.md
```diff
@@ -0,0 +1,618 @@
...
+Contains the arguments returned by the event fired when a Window instance's visibility changes.
+```CS
+public class VisibilityChangedEventArgs
+```
+
+### Properties:
+- VisibilityChangedEventArgs.Handled: Gets or sets whether the window activation changed event was handled. **true** if the event has been handled by the appropriate delegate; **false** if it has not.
+    ```CS
+    public bool Handled { get; set; };
+    ```
```
dotMorten: What's the point of handling this? Will that prevent it from becoming visible or hidden?

active/Win32/Window_and_Application_API_Spec.md
```diff
@@ -0,0 +1,618 @@
...
+## WindowSizeChangedEventArgs class
+Namespace: Microsoft.UI.Xaml
+
+Contains the argument returned by a window size change event.
+
+### Properties
+
+- WindowSizeChangedEventArgs.Handled: Gets or sets whether the  event was handled. **true** if the event has been handled by the appropriate delegate; **false** if it has not.
+    ```CS
+    public bool Handled { get; set; };
```
dotMorten: What's the point of handling this? Will that prevent it from resizing?

active/Win32/Window_and_Application_API_Spec.md
```diff
@@ -0,0 +1,618 @@
...
+
+
+```CS
+public class Application:IClosable
+```
+```XML
+<Application />
+ ```
+
+> **NOTE**: It is not a goal of this doc to review all the members of this class. Just that ones that requires changes in Win32. You can find more inforation about Application class [here](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.application).
```
dotMorten: I wish this was on top, because I just did that :)

active/Win32/Window_and_Application_API_Spec.md
```diff
@@ -0,0 +1,618 @@
...
+```CS
+public partial class MainWindow:Window
+{
+  public MainWindow()
+  {
+	InitializeComponent();
+  }
+}
+```
+##### Multiple Widndows Sample
```
jeffstall, line 68: >Widndows [](start = 15, length = 8)

Spelling

active/Win32/Window_and_Application_API_Spec.md
```diff
@@ -0,0 +1,618 @@
...
+private void Button_Click(object sender, RoutedEventArgs e)
+{
+    Thread thread = new Thread(() =>
+    {
+        Window w = new Window();
+        w.Content = new TextBlock() { Text = "Separated Thread Window" };
+        w.Activate();
+        w.Closed += (sender1, e1) => w.Dispatcher.InvokeShutdown();
+
+        System.Windows.Threading.Dispatcher.Run();
```
jeffstall, line 97: >System [](start = 8, length = 6)

Which Dispatcher did you intend here?

active/Win32/Window_and_Application_API_Spec.md
```diff
@@ -0,0 +1,618 @@
...
+   w2.Content = new TextBlock(){Text=”Window 2”}
+   w2.Activate();
+}
+```
+##### Create a Window in a separated thread
+To create a Window in a separated thread, you should create the thread first and create a Window in this new Thread. 
+```CS
+private void Button_Click(object sender, RoutedEventArgs e)
+{
+    Thread thread = new Thread(() =>
```
jeffstall, line 90: >Thread [](start = 4, length = 6)

Does it make sense to have a Xaml worker function / class that makes this more maintainable, abstracts Win32 and UWP, and enables us to service? May we follow-up?

active/Win32/Window_and_Application_API_Spec.md
```diff
@@ -0,0 +1,618 @@
...
+- **UWP**: There is already one CoreWindow created and UWP only allows one CoreWindow per thread, as a consequence, the contructor and the XAML Markup throw an exception when running in UWP.
+    - Exception type: TBD
+ 
+    - Exception Message: TBD. Something like "Only one Window per thread is allowed in UWP app model".
+
+### Window.Bounds property
+Gets a Rect value containing the height and width of the application window in logical pixels.
+
+```CS
+public Rect Bounds { get; }
```
jeffstall, line 119: >Bounds [](start = 12, length = 6)

A "Rect" is particularly problematic. Can we change this to ClientSize or something? We'll want to distinguish from FrameSize, which is in a different coordinate space and affected by borders, etc.

active/Win32/Window_and_Application_API_Spec.md
```diff
@@ -0,0 +1,618 @@
...
+This agreement applies to the Application class too.
+
+## Window Class
+Namespace: Microsoft.UI.Xaml
+
+```CS
+public class Window : IClosable
+```
+
+### **Window management, appearance, and behaviors**
```
jeffstall, line 33: >Window management, appearance, and behaviors [](start = 6, length = 44)

Please take a look at AppWindow and Forms for examples of basic knobs to customize the Window (frame style, show title bar, etc.):
[https://docs.microsoft.com/en-us/uwp/api/windows.ui.windowmanagement.appwindow](https://docs.microsoft.com/en-us/uwp/api/windows.ui.windowmanagement.appwindow)
[https://docs.microsoft.com/en-us/dotnet/api/system.windows.forms.form?view=netframework-4.8](https://docs.microsoft.com/en-us/dotnet/api/system.windows.forms.form?view=netframework-4.8)

active/Win32/Window_and_Application_API_Spec.md
```diff
@@ -0,0 +1,618 @@
...
+<Window 
+    x:Class="MainWindow"
+    xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+>
+    <Grid><Button Content=”Click” Click=”OnClick”/></Grid>
+</Window>
+```
+```CS
+public partial class MainWindow:Window
```
jeffstall, line 60: >MainWindow:Window [](start = 21, length = 17)

Please write "MainWindow : Window", with the spaces.

active/Win32/Window_and_Application_API_Spec.md
```diff
@@ -0,0 +1,618 @@
...
+  xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+  xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+  x:Class="Sample.MainWindow"
+    Icon="WinUI3Icon.ico">
+</Window>
+```
+
+#### Remarks:
+- **Win32**: The image is set as an Icon of the Window. 
+- **UWP**: Do nothing. The property value is ignored.
```
jeffstall, line 152: Does it make sense to have a more unified way to specify the icon between Win32 and UWP? I think it comes from the manifest for UWP.

active/Win32/Window_and_Application_API_Spec.md
```diff
@@ -0,0 +1,618 @@
...
+   {
+      //
+   }
+}
+```
+#### Remarks
+VisibilityChangedEventArgs includes a Boolean Visible property that you should check to determine the visibility state when the event was fired.
+
+### Window.SetTitleBar(UIElement) Method
+Set the custom XAML content that should act as the title bar.
```
jeffstall, line 205: This is out of scope for Nov 2020 for Win32 apps.

active/Win32/Window_and_Application_API_Spec.md
```diff
@@ -0,0 +1,618 @@
...
+- **UWP**: This string is not displayed in the Window's Title bar when SetTitleBar method is used or set a not null content. The Title is still used in Alt+Tab scenarios,
+
+
+### Window.Current property
+
+Gets the window of the current thread.
+
+```CS
+public static Window Current { get; }
+```
```
jeffstall, line 250: IMPORTANT: Absolutely do not do this! There is no "current" window when we support multiple windows on a single thread. Trying to store this and update it "depending on the operation" is a very problematic issue that does not have any good answers. It's been well researched for the last decade as part of the ApplicationView / CoreWindow problems with GetForCurrentView.

active/Win32/Window_and_Application_API_Spec.md
```diff
@@ -0,0 +1,618 @@
...
+
+Closes and destroy the window. The Closed event occurs when a Window closes. 
+```CS
+public void Close()
+```
+
+#### Remarks
+- **UWP**: UWP app typically use a single Window instance, and do not open and close additional Window instances. Additionally, you will not typically provide custom UI to enable users to close the app window.
+
+### Window.Closed event
```
jeffstall, line 307: >Closed [](start = 11, length = 6)

Do you want to have typical functionality here:
1. Allow app to get a deferral and start saving state.
2. Potentially override the "closing" behavior.

I'd recommend talking with Roberth and Francis on real-world scenarios here.

active/Win32/Window_and_Application_API_Spec.md
```diff
@@ -0,0 +1,618 @@
...
+```
+#### Remarks
+- **UWP**: More info [here](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.window.sizechanged).
+
+### **Get the underneath window implementation**
+UWP apps use CoreWindow as a window implementation and Win32 apps use HWnds instead. The Microsoft.UI.Xaml.Window hides these detail implementations from the application developers and exposes a set of APIs that allow to create and manipulate Windows.
+These properties expose these underneath objects to enable advanced (or basic but not supported yet) scenarios, so developers can get the CoreWindow or HWnd objects to apply setters, listening events, etc.
+
+### Window.CoreWindow property: 
+
```
jeffstall, line 362: Xaml's "Window" is going to become huge if it becomes a dumping ground for properties that will rarely be used. Can we move these to Hosting as a static member instead?

static class UwpWindowHosting
{
    static ApplicationView GetView(Window window);
    static CoreWindow GetCoreWindow(Window window);
}

Or, just have them call CoreWindow.GetForCurrentThread() and ApplicationView.GetForCurrentView() like everyone has done since Win 8.x.

active/Win32/Window_and_Application_API_Spec.md
```diff
@@ -0,0 +1,618 @@
...
+IntPtr hwnd = w.GetInterop<IWindowInterop>().Handle;
+```
+
+### **Others**
+### Window.Dispatcher property
+Gets the Dispatcher object for the Window.
+
+```CS
+public CoreDispatcher Dispatcher { get; }
+```
```
jeffstall, line 387: Change to the DispatcherQueue, since CoreDispatcher isn't that interesting.

active/Win32/Window_and_Application_API_Spec.md
mdtauk: I would prefer there to be no Icon by default, which would mimic UWP behaviour

active/Win32/Window_and_Application_API_Spec.md
mdtauk: With Live Tiles displaying Icons, it probably makes sense to ensure a single ico file can be used to cover all the uses, without having to duplicate.

active/Win32/Window_and_Application_API_Spec.md
marb2000: This is a great suggestion. No showing up Icon unless you specify one. The TaskBar and Alt+Tab icon could be the application icon.

active/Win32/Window_and_Application_API_Spec.md
```diff
@@ -405,12 +405,79 @@ Provides event information when an app is launched.
...
+{
+    String Arguments{ get; };
+    Windows.ApplicationModel.Activation.LaunchActivatedEventArgs UWPLaunchActivatedEventArgs{ get; };
+};
+```
+
+# Appendix
+
+Changes in this spec from Preview2:
+* Rename LaunchActivatedEventArgs to XamlLaunchActivatedEventArgs
```
dotMorten, line 481: Isn't that already implied from the namespace? IMHO I find it a bit weird to prefix this with `Xaml` for an application that launched.

active/Win32/Window_and_Application_API_Spec.md
dotMorten: @jeffstall didn't you just say in the above comment
> There is no "current" window when we support multiple windows on a single thread

Is so, how would CoreWindow.GetForCurrentThread() work?

active/Win32/Window_and_Application_API_Spec.md
MikeHillberg: That's to try and avoid confusion with the existing type named [LaunchActivatedEventArgs](https://docs.microsoft.com/uwp/api/Windows.ApplicationModel.Activation.LaunchActivatedEventArgs) (in the Windows.ApplicationModel.Activation namespace). This Xaml type is a wrapper around that.

Another alternative could be to simply give this a different name; "launch activated" is a bit confusing in this context. For example `ApplicationActivatedEventArgs`.

active/Win32/Window_and_Application_API_Spec.md
MikeHillberg: This is an existing API and this is the [existing text]([Window.Bounds](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.Window.Bounds)), but agreed it's wrong. 
* In UWP this should be simply returning the [CoreWindow.Bounds](https://docs.microsoft.com/uwp/api/Windows.UI.Core.CoreWindow.Bounds).
* On Desktop this returns the GetClientRect adjusted for rasterization scale.

active/Win32/Window_and_Application_API_Spec.md
MikeHillberg: This is an existing API. But see the [other comment](https://github.com/microsoft/microsoft-ui-xaml-specs/pull/71/files/9c72d8dde073924516b125e1ddaab1de36c61bc4#r393373967) about this property.

active/Win32/Window_and_Application_API_Spec.md
MikeHillberg: This is an existing API. The purpose of calling it out in this spec is to add the below documentation which explains that it returns null in Desktop apps for that reason.

active/Win32/Window_and_Application_API_Spec.md
MikeHillberg: It's an existing property. But DispatcherQueue is being added

active/Win32/Window_and_Application_API_Spec.md
MikeHillberg: I think his point is that rather than having instance members on the Window class, we should have static members on helper classes that take a Window instance as their first parameter.

We've done this for dumping grounds before (like [Typography](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.Documents.Typography)), but in general it serves to make APIs undiscoverable. And I don't see Window becoming a dumping ground (WPF's [Window](https://docs.microsoft.com/dotnet/api/System.Windows.Window) is an existing precedent to look at).

marb2000:
Closing this PR given we already moved and updated the Spec [here](
https://github.com/microsoft/microsoft-ui-xaml-specs/blob/Win32/active/Win32/Window_and_Application_API_Spec.md)

marb2000:
Sorry, I mistake.

active/Win32/Window_and_Application_API_Spec.md
```diff
@@ -118,10 +115,41 @@ thread.Start();
...
+//Asumming MainWindow has a CustomTitleBar public property which is a UserControl, and this UserControl also has some public properties for customization
+private Window m_window;
+
+protected override void OnLaunched(Microsoft.UI.Xaml.LaunchActivatedEventArgs args)
+{
+    m_window = new MainWindow();
+
+    //Assuming TitleBar never is null
+    m_window.ExtendContentIntoTitleBar = true;
+    m_window.SetTitleBar(m_window.CustomTitleBar);
```
dotMorten, line 143: Why not just `m_window.TitleBar = m_window.CustomTitleBar;` ?

active/Win32/Window_and_Application_API_Spec.md
```diff
@@ -433,6 +458,7 @@ Microsoft.UI.Xaml namespace
         String Title;
         ImageSource Icon;
         Microsoft.System.DispatcherQueue DispatcherQueue{ get; };
+	Boolean ExtendContentIntoTitleBar { get; set;}
```
dotMorten, line 461: nit: indent

active/Win32/Window_and_Application_API_Spec.md
marb2000: This is good question. I just kept it for compatibility with UWP because it's an existing API, but I don't know the reasons of why we did in this way long time ago. @MikeHillberg any idea?

active/Win32/Window_and_Application_API_Spec.md
marb2000: Closing this conversation given Icon property is no longer a plan for 2020. We will discuss about it in the next update.

active/Win32/Window_and_Application_API_Spec.md
marb2000: We are introducing an API in the XAML layer to extend the content into the Title bar. Resolving this conversation.

active/Win32/Window_and_Application_API_Spec.md
marb2000: We kept it for backwards compatibility with the existing APIs. Honestly, I didn't find a scenario that justify this, but perhaps my scope (from XAML) is narrower than the scope of UWP windowing creators.

active/Win32/Window_and_Application_API_Spec.md
marb2000: Some comment as above

active/Win32/Window_and_Application_API_Spec.md
marb2000: We change the scope of this spec, and only cover the updates, so this comment is outdated. Resolving it.

active/Win32/Window_and_Application_API_Spec.md
marb2000: This is a good point, but out of scope for 2020. Adding a note to capture this in the backlog.

active/Win32/Window_and_Application_API_Spec.md
marb2000: Icon property was moved out of 2020. We will discuss this when we add this feature in the future.

active/Win32/Window_and_Application_API_Spec.md
marb2000: Same comment as above

active/Win32/Window_and_Application_API_Spec.md
marb2000: It seems that the preferred was XamlLaunchActivatedEventArgs. Resolving the conversation

active/Win32/Window_and_Application_API_Spec.md
marb2000: Fixed.

active/Win32/Window_and_Application_API_Spec.md
marb2000: XAML's Window will live with Current for backwards compatibility. Perhaps we can deprecate it in the future, but not now.

active/Win32/Window_and_Application_API_Spec.md
marb2000: Thanks for the piece of advice!  We will talk to them. Resolving this conversation,

active/Win32/Window_and_Application_API_Spec.md
MikeHillberg: I don't recall either. I do remember we talked about adding a Get method. We in general have a guideline that property setters and getters should be trivial, and if they do any real work should be methods (though we've probably violated that in a few places to be friendlier to markup). Maybe that kicked in here.

MikeHillberg: Leaving this as a method since it's an existing API.

active/Win32/Window_and_Application_API_Spec.md
```diff
@@ -523,6 +547,28 @@ In a UWP app, this is equivalent to LaunchActivatedEventArgs.Arguments.
 
 # API Details
 
+?? Need to cover resources:
```
mdtauk, line 550: You could use ` ```XAML ` to add context colour to this code.

Also when you say Need to cover - does that mean you will still be able to set these values in Xaml for both UWP and Win32 **Window**s that are created in WinUI 3?

active/Win32/Window_and_Application_API_Spec.md
```diff
@@ -0,0 +1,654 @@
...
+Only one UIElement can act as Title bar, so the last set wins. To use multiple objects, 
+you need to wrap them in a container element (for example a Grid).
+
+The rectangular area occupied by the element acts as the title bar for pointer purposes,
+even if the element is blocked by another element,
+even if the element is transparent.
+
+This method is typically used with the Window's `ExtendContentIntoTitleBar` property 
+set to true in order to hide the default system title bar.
+However, even when the default system title bar is not hidden, SetTitleBar can be used to make
```
stevenbrix, line 196: what is the user scenario for making any part of the app behave like the title bar? Since we want a consistent look and feel across all Windows apps, it would seem like that in the 97% use case, we don't want apps to differentiate from the default behavior. I'm sure there is a valid scenario where this is desired, and it would be helpful for the uninformed to understand when that is.

mdtauk: > what is the user scenario for making any part of the app behave like the title bar? Since we want a consistent look and feel across all Windows apps, it would seem like that in the 97% use case, we don't want apps to differentiate from the default behavior. I'm sure there is a valid scenario where this is desired, and it would be helpful for the uninformed to understand when that is.
>
Pretty much any modern UWP app will extend into the Titlebar area, and will place a XAML element there so you can still grab the top of the window to move the window around.

image
_Settings_

image
_Your Phone_

image
_Calculator_

image
_Heutro_

image
_Windows Store_

image
_Alarms and Clock_

image
_Maps_

image
_Messenger (React Native)_

stevenbrix: Sorry, I must not have explained myself well. I'm not suggesting that extending into the title bar doesn't have value. 

But the spec, and implementation, allows me to do something like this:

image

Where the "title bar" is now in the center of my app. And, unlike the spec, I can no longer interact with the top of the Window and drag it. It doesn't matter if I set `CoreApplicationViewTitleBar.ExtendViewIntoTitleBar` or not. As soon as I call `Window.SetTitleBar` the top of the window is no longer draggable

mdtauk: Apps which Extend into the titlebar, will often put items into the titlebar area, like the "What's new" button in the Your PC app.

The SetTitleBar API is used to allow the hit target of the panel placed where the Titlebar would be, to be used for dragging, snapping, and double clicking to maximise a window.  Without it, anything you place in that area would eat the taps and clicks, and it would not be possible to move the windows.

``
This method lets you specify a XAML element that interacts with the system as if it’s the app window's title bar. When you use this method, you typically set the CoreApplicationViewTitleBar.ExtendViewIntoTitleBar property to true in order to hide the default system title bar. However, even when the default system title bar is not hidden, this can be used to make additional regions in your app behave like the title bar.
``

https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.window.settitlebar?view=winrt-19041

MikeHillberg: I don't understand/recall the motivation for the existing API to allow you to make an element in the middle of the Window be the title bar (for purposes of drag etc). A Window.TitleBar property sounds more useful to me. For the moment though we're just enabling the existing API on Desktop.

active/Win32/Window_and_Application_API_Spec.md
MikeHillberg: I'm removing this section, we're not trying yet to make this part of the API.

active/Win32/Window_and_Application_API_Spec.md
```diff
@@ -0,0 +1,662 @@
...
+    m_window = new MainWindow();
+
+    m_window.ExtendContentIntoTitleBar = true;
+    m_window.SetTitleBar(m_window.CustomTitleBar);
+
+    m_window.Activate();
+}
+```
+
+Title bars should be have background and foreground colors based on
```
mdtauk, line 223: "Title bars should have background"... the **be** is not needed

active/Win32/Window_and_Application_API_Spec.md
```diff
@@ -0,0 +1,662 @@
...
+
+> Implementation note: Attempting to activate a new window in a UWP app will use 
+[RoOriginateError](https://docs.microsoft.com/en-us/windows/win32/api/roerrorapi/nf-roerrorapi-rooriginateerror) 
+to show an explanatory message in the debugger.
+
+Creating a new Window in a Desktop app creates a new top level HWND.
+
+## Window.ExtendsContentIntoTitleBar
+
+Gets or sets a value that specifies whether the default window title bar should be removed, creating more space for content.
```
mdtauk, line 153: Rather than saying it makes more space, might it not be better and more accurate to say something like **it allows the App's UI to extend to the top of the window's frame, where the titlebar would usually be.**

active/Win32/Window_and_Application_API_Spec.md
```diff
@@ -0,0 +1,662 @@
...
+The rectangular area occupied by the element acts as the title bar for pointer purposes,
+even if the element is blocked by another element,
+even if the element is transparent.
+
+This method is typically used with the Window's `ExtendContentIntoTitleBar` property 
+set to true in order to hide the default system title bar.
+However, even when the default system title bar is not hidden, SetTitleBar can be used to make 
+additional regions in your app behave like the title bar.
+
+The following shows a Window with a no built-in title bar.
```
mdtauk, line 192: "The following shows a Window with ~a~ no built-in title bar." remove **a**

active/Win32/Window_and_Application_API_Spec.md
```diff
@@ -0,0 +1,662 @@
...
+even if the element is transparent.
+
+This method is typically used with the Window's `ExtendContentIntoTitleBar` property 
+set to true in order to hide the default system title bar.
+However, even when the default system title bar is not hidden, SetTitleBar can be used to make 
+additional regions in your app behave like the title bar.
+
+The following shows a Window with a no built-in title bar. 
+Instead it uses the whole window as its content area. To enable features such as window dragging, 
+it designates a TextBlock as the title bar.
```
mdtauk, line 194: Do you want the example to use the TextBlock as the titlebar, or the <Grid> containing the TextBlock?  If so, the Grid should also use a background color of **Transparent** for hit testing.

mdtauk:
Will there be support for WinUI 3 Windows which can hide the Minimise and Maximise buttons, much like the modal dialogs seen commonly in win32 apps - as well as MessageBoxes etc?  I know these will probably be ignored by UWP app windows.

win32 dialog window
_Close only_

Windows 10X titlebar
_10X titlebar with no Maximise/Restore button_

active/Win32/Window_and_Application_API_Spec.md
```diff
@@ -0,0 +1,655 @@
...
+
+### Remarks
+
+The Window class can be used both in a UWP app or a Desktop app.
+
+In a Desktop app you can create (`new`) a Window, even if the current thread already has a Window on it.
+In a UWP app you must create a new 
+[CoreApplicationView](https://docs.microsoft.com/uwp/api/Windows.ApplicationModel.Core.CoreApplicationView),
+which will automatically create a new thread and a Window on it. 
+See the 'Create new Window in UWP application' example.
```
jonwis, line 143: What if I `new` a Window while on the Dispatcher for a CoreWindow? Could that create me a new [AppWindow](https://docs.microsoft.com/en-us/uwp/api/Windows.UI.WindowManagement.AppWindow?view=winrt-19041) instead?

active/Win32/Window_and_Application_API_Spec.md
```diff
@@ -0,0 +1,655 @@
...
+```
+
+## Window.Dispatcher property
+_Behavior update to existing API when running as Desktop app_
+
+Gets the [CoreDispatcher](https://docs.microsoft.com/uwp/api/Windows.UI.Core.CoreDispatcher) 
+object for the Window when called from a UWP app. Returns null when called from a Desktop app.
+
+```CS
+public CoreDispatcher Dispatcher { get; }
```
jonwis, line 283: So we end up with both `Dispatcher` and `DispatcherQueue` as aliases of each other? Should this property be insta-deprecated to help developers move to the preferred one instead?

active/Win32/Window_and_Application_API_Spec.md
```diff
@@ -0,0 +1,655 @@
...
+}
+```
+
+## Application class
+
+> Spec note: Application is not new to this spec, what is new are the following remarks
+
+### Remarks
+
+The following events and virtual methods are not invoked when running in a Desktop app:
```
jonwis, line 335: Maybe link over to the Project Reunion issue thread on Win32-facing activation events so folks know they're not out of luck?

marb2000 (commented):
I don't get it. This is a C# sample, and in C# WindowHandle returns IntPtr. HWND is not even a type. Why this change?

active/Win32/Window_and_Application_API_Spec.md
```diff
@@ -0,0 +1,655 @@
...
+
+
+## Application.RequiresPointerMode property
+_Behavior update to existing API when running as Desktop app_
+
+Gets or sets whether a UWP app supports mouse mode, which emulates pointer interaction experiences 
+with non-pointer input devices such as an Xbox gamepad or remote control. (All nested elements inherit 
+this behavior.)
+
+> Note: this property is ignored in a Desktop app
```
oldnewthing, line 423: So does this mean that Desktop apps always support mouse mode? Or Desktop apps never support mouse mode? Please clarify by saying what the Desktop behavior is:

> Note: This property is ignored in a Desktop app. For Desktop apps, pointer mode is always [ Never | WhenEngaged | WhenFocused | something else ].

active/Win32/Window_and_Application_API_Spec.md
```diff
@@ -0,0 +1,655 @@
...
+Invoked when the application creates a window.
+
+```CS
+protected virtual void OnWindowCreated(XamlWindowCreatedEventArgs args)
+```
+
+### Remarks
+- In a UWP app, this method is called once for the creation of the main window, once for each hosted window, 
+and once each time CreateNewView is called. 
+- In a Desktop app, this method is called every time a Window object is created.
```
oldnewthing, line 437: I assume this method is called from the thread that created the window.

active/Win32/Window_and_Application_API_Spec.md
```diff
@@ -0,0 +1,655 @@
...
+_Behavior update to existing API when running as Desktop app_
+
+Invoked when the application creates a window.
+
+```CS
+protected virtual void OnWindowCreated(XamlWindowCreatedEventArgs args)
+```
+
+### Remarks
+- In a UWP app, this method is called once for the creation of the main window, once for each hosted window,
```
oldnewthing, line 435: For a UWP app, does this mean that OnWindowCreated is called before OnLaunched? This would be a gotcha case for many apps, since they typically do the bulk of their initialization in OnLaunched. May want to add a Note for this case.

active/Win32/Window_and_Application_API_Spec.md
```diff
@@ -0,0 +1,655 @@
...
+
+> Spec note: this is a new XAML version of the existing 
+Windows.UI.Core.[WindowActivatedEventArgs](https://docs.microsoft.com/uwp/api/Windows.UI.Core.WindowActivatedEventArgs).
+
+## XamlWindowActivatedEventArgs properties
+
+**Handled**
+
+Gets or sets whether the window activation changed event was handled.
+ **true** if the event has been handled by the appropriate delegate; **false** if it has not.
```
oldnewthing, line 453: This is an "-ed" event, so the activation has already occurred. The Handled would presumably merely prevent other event handlers from seeing the event? That seems like a weird thing to do. What's the use case for handling the event?

active/Win32/Window_and_Application_API_Spec.md
```diff
@@ -0,0 +1,655 @@
...
+### XamlVisibilityChangedEventArgs
+
+Contains the arguments returned by the event fired when a Window instance's visibility changes.
+
+> Spec note: this is a new XAML version of the existing 
+Windows.UI.Core.[VisibilityChangedEventArgs](https://docs.microsoft.com/uwp/api/Windows.UI.Core.VisibilityChangedEventArgs).
+
+**Handled**
+
+Gets or sets whether the window activation changed event was handled. **true**
```
oldnewthing, line 485: copy/pasta. (Also, same question about what "Handled" means here.)

active/Win32/Window_and_Application_API_Spec.md
```diff
@@ -0,0 +1,655 @@
...
+Windows.UI.Core.[VisibilityChangedEventArgs](https://docs.microsoft.com/uwp/api/Windows.UI.Core.VisibilityChangedEventArgs).
+
+**Handled**
+
+Gets or sets whether the window activation changed event was handled. **true** 
+if the event has been handled by the appropriate delegate; **false** if it has not.
+
+**Visible**
+
+Gets whether the window is visible or not. **true** if the window is visible; otherwise is **false**.
```
oldnewthing, line 490: "is" or "was"? This is an "-ed" event, so the property presumably refers to whether the window was made visible or not, rather than referring to the window's current visibility state (which you can obtain from the Window.Visible property.)

active/Win32/Window_and_Application_API_Spec.md
```diff
@@ -0,0 +1,655 @@
...
+
+> Spec note: this is a new XAML version of the existing Windows.ApplicationModel.Activation.
+[LaunchActivatedEventArgs](https://docs.microsoft.com/uwp/api/Windows.ApplicationModel.Activation.LaunchActivatedEventArgs).
+
+## XamlLaunchActivatedEventArgs Properties
+
+**Arguments**
+
+Gets the arguments that are passed to the app during its launch. 
+In a UWP app, this is equivalent to LaunchActivatedEventArgs.Arguments.
```
oldnewthing, line 524: I assume that for a Desktop app, the Arguments is the raw command line that comes after the program name (leading spaces preserved)? I.e., we don't try to parse it in any way with CommandLineToArgvW or anything like that?

active/Win32/Window_and_Application_API_Spec.md
```diff
@@ -0,0 +1,655 @@
...
+[webhosthidden]
+[contentproperty("Content")]
+unsealed runtimeclass Window
+{
+
+    ...
+
+    // These APIs match the system Xaml versions, except the event args are now from the Xaml namespace
+    event Windows.Foundation.TypedEventHandler<Object,Microsoft.UI.Xaml.XamlWindowActivatedEventArgs> Activated;
+    event Windows.Foundation.TypedEventHandler<Object,Microsoft.UI.Xaml.XamlWindowClosedEventArgs> Closed;
```
oldnewthing, line 557: Missing from spec. Did we intend to keep this? Or did we forget to delete it from the IDL?

active/Win32/Window_and_Application_API_Spec.md
```diff
@@ -0,0 +1,655 @@
...
+runtimeclass XamlWindowVisibilityChangedEventArgs
+{
+    Boolean Handled;
+    Boolean Visible{ get; };
+};
+```
+
+```cs
+[webhosthidden]
+runtimeclass XamlWindowClosedEventArgs
```
oldnewthing, line 650: Missing from spec. Did we intend to keep this? Or did we forget to delete it from the IDL?

active/Win32/Window_and_Application_API_Spec.md
```diff
@@ -0,0 +1,655 @@
...
+On any other thread, the value is null.
+
+## Window.CoreWindow property
+_Behavior update to existing API when running as Desktop app_
+
+Gets the [CoreWindow](https://docs.microsoft.com/uwp/api/Windows.UI.Core.CoreWindow) associated 
+with this Window. In a Desktop app this is always null.
+
+```CS
+public CoreWindow CoreWindow { get; }
```
jeffstall, line 273: There's no reason to expose CoreWindow and CoreDispatcher. If UWP apps really need these, just use the existing static methods on those types. (i.e. CoreWindow.GetForCurrentThread)

active/Win32/Window_and_Application_API_Spec.md
```diff
@@ -0,0 +1,655 @@
...
+
+To create a new Window on a new thread in a Desktop app, create the thread first:
+
+```CS
+var thread = new Thread(() =>
+{
+    Window window = new Window();
+    window.Content = new TextBlock() { Text = "Hello" };
+    window.Activate();
+    window.Closed += (_, __) => w.DispatcherQueue.InvokeShutdown();
```
jeffstall, line 120: Sample feedback. Please use real variable names, not "_".

active/Win32/Window_and_Application_API_Spec.md
```diff
@@ -0,0 +1,655 @@
...
+[described here](https://github.com/microsoft/microsoft-ui-xaml-specs/blob/master/winui3/DispatcherQueueUpdates.md)
+
+
+## IWindowNative COM Interface
+
+This interface is implemented by `Window`, and in a Desktop app can be used to get 
+the Window's underlying HWND.
+
+```CS
+protected override void OnLaunched(Microsoft.UI.Xaml.XamlLaunchActivatedEventArgs e)
```
jeffstall, line 316: I don't understand this sample. Is it in C#, which wouldn't have the IWindowNative interface definition, or in C++, which wouldn't have this language syntax.

active/Win32/Window_and_Application_API_Spec.md
```diff
@@ -0,0 +1,655 @@
...
+   HWND hwnd = windowNative.WindowHandle;
+   ...
+}
+```
+
+## Application class
+
+> Spec note: Application is not new to this spec, what is new are the following remarks
+
+### Remarks
```
jeffstall, line 333: Should we do some refactoring here, since it gets confusing if we have a lot of things that never work for Win32, and this is becoming a front and center scenario?

active/Win32/Window_and_Application_API_Spec.md
```diff
@@ -0,0 +1,655 @@
...
+## Window class
+Microsoft.UI.Xaml namespace
+
+> The Window class' APIs in WinUI3 (**Microsoft**.UI.Xaml) match the existing system Window
+API (**Windows**.UI.Xaml), except for the following differences.
+
+```cs
+[webhosthidden]
+[contentproperty("Content")]
+unsealed runtimeclass Window
```
jeffstall, line 550: Important to specify event order. Does VisibilityChanged get fired before or after SizeChanged? Does VisibilityChanged get fired when minimized?

sylveon: If visibility maps to ShowWindow(SW_HIDE/SW_SHOW) and IsWindowVisible, I don't think it should fire, since a minimized app is still "visible".

PR #72, opened 2020-03-16
Update Window_and_Application_API_Spec.md

thomasclaudiushuber:
I don't know what the property should be called there, but definitely not Visible. :-)

### DIFF
--- a/active/Win32/Window_and_Application_API_Spec.md
+++ b/active/Win32/Window_and_Application_API_Spec.md
@@ -482,9 +482,9 @@ Contains the argument returned by a window size change event.
     public bool Handled { get; set; };
     ```
 
-- WindowSizeChangedEventArgs.Visible: Gets the new size of the window in units of effective (view) pixels.
+- WindowSizeChangedEventArgs.WindowSize: Gets the new size of the window in units of effective (view) pixels.
     ```CS
-    public Size Visible { get; } 
+    public Size WindowSize { get; } 
     ```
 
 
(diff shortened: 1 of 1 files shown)


active/Win32/Window_and_Application_API_Spec.md
```diff
@@ -482,9 +482,9 @@ Contains the argument returned by a window size change event.
     public bool Handled { get; set; };
     ```
 
-- WindowSizeChangedEventArgs.Visible: Gets the new size of the window in units of effective (view) pixels.
+- WindowSizeChangedEventArgs.WindowSize: Gets the new size of the window in units of effective (view) pixels.
     ```CS
-    public Size Visible { get; } 
+    public Size WindowSize { get; }
```
dotMorten: I'd prefer just `Size`

thomasclaudiushuber: Yes, I thought about that too. But maybe when you have an event handler and the parameter is just called "e", e.WindowSize looks better than e.Size. But I'm not an API design expert. :-)

dotMorten:
Then the other size changed event should be called ControlSize. It's more consistent this way.
Also don't see why we can't reuse the existing SizeChangedEventArgs class here?

active/Win32/Window_and_Application_API_Spec.md
dotMorten: Then the other size changed event should be called ControlSize. It's more consistent just calling it Size.
Also don't see why we can't reuse the existing SizeChangedEventArgs class here?

thomasclaudiushuber: Ok, changed to just Size.

Yes, the question with SizeChangedEventArgs is good. @marb2000 might know. :)

PR #74, opened 2020-03-19
ItemsSourceView.IndexOf API spec

chingucoding:
This PR will add the API spec for the new ItemsSourceView.IndexOf function.

See https://github.com/microsoft/microsoft-ui-xaml/issues/1828 for more context.


active/ItemsSourceView-Indexof/ItemsSourceView-IndexOf.md
```diff
@@ -0,0 +1,67 @@
+# Background
+<!-- Use this section to provide background context for the new API(s) 
+in this spec. -->
+
+Currently the ItemsSourceView class does not expose an IndexOf function. However during development, this method was used multiple times internally on derivatives of the class. To make the usage of this existing function easier on subclasses, we should expose this function in the ItemsSourceView.
```
MikeHillberg: I can't remember, why isn't ItemsSourceView just an IVector?

active/ItemsSourceView-Indexof/ItemsSourceView-IndexOf.md
```diff
@@ -0,0 +1,67 @@
...
+# API Notes
+ItemsSourceView.IndexOf(IInspectable item): Returns index in the items source. If an item is not in the items source, it will return -1.
+# API Details
+<!-- The exact API, in MIDL3 format (https://docs.microsoft.com/en-us/uwp/midl-3/) -->
+
+```MIDL
+
+unsealed runtimeclass ItemsSourceView: Windows.UI.Xaml.Interop.INotifyCollectionChanged
+{
+
```
MikeHillberg:
```suggestion
     // ...
```

active/ItemsSourceView-Indexof/ItemsSourceView-IndexOf.md
```diff
@@ -0,0 +1,67 @@
...
+<!-- Use this section to provide background context for the new API(s) 
+in this spec. -->
+
+Currently the ItemsSourceView class does not expose an IndexOf function. However during development, this method was used multiple times internally on derivatives of the class. To make the usage of this existing function easier on subclasses, we should expose this function in the ItemsSourceView.
+
+# Description
+
+The IndexOf function returns the index of the given element in the items source the ItemsSourceView is used for.
+Note that if an element is not contained in the items source, the function will return -1.
+
```
MikeHillberg:
```suggestion
> Spec note: returning -1 matches the behavior of for example IndexOf on [UIElementCollection](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.Controls.UIElementCollection) and List<T>.
```

active/ItemsSourceView-Indexof/ItemsSourceView-IndexOf.md
```diff
@@ -0,0 +1,67 @@
...
+}
+
+```
+
+# Appendix
+<!-- Anything else that you want to write down for posterity, but 
+that isn't necessary to understand the purpose and usage of the API.
+For example, implementation details. -->
+
+When IndexOf of the ItemsSourceView class get's invoked, should we instead of throwing `hresult_not_implemented` return -1 instead?
```
MikeHillberg: Spec already says return -1, which I think is right.

active/ItemsSourceView-Indexof/ItemsSourceView-IndexOf.md
```diff
@@ -0,0 +1,67 @@
+# Background
+<!-- Use this section to provide background context for the new API(s) 
+in this spec. -->
+
```
MikeHillberg:
```suggestion
The Xaml [ItemsRepeater](https://docs.microsoft.com/uwp/api/Microsoft.UI.Xaml.Controls.ItemsRepeater) control displays a list of objects using a specified template (similar to a [ListView](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.Controls.ListView)). You provide the list of objects with the [ItemsRepeater.ItemsSource](https://docs.microsoft.com/uwp/api/Microsoft.UI.Xaml.Controls.ItemsRepeater.ItemsSource) property. The list need only be an IIterable but can be an IVector for better perf. ItemsRepeater exposes a read-only [ItemsRepeater.ItemsSourceView](https://docs.microsoft.com/uwp/api/Microsoft.UI.Xaml.Controls.ItemsRepeater.ItemsSourceView) property that provides a view on this list (an [ItemsSourceView](https://docs.microsoft.com/uwp/api/Microsoft.UI.Xaml.Controls.ItemsSourceView)), which has helpers like Count and GetAt over what might be just an IIterable.

```

active/ItemsSourceView-Indexof/ItemsSourceView-IndexOf.md
MikeHillberg:
```suggestion
Currently the ItemsSourceView class does not expose an IndexOf function. However during development, this method was used multiple times internally on derivatives of the class. To make the usage of this existing function easier on subclasses, this spec adds an IndexOf function to the ItemsSourceView.
```

chingucoding: I think the direction IVector is going and where ItemsSourceView is used are slightly different. The idea of IVector is to enable different collection operations, while ItemsSourceView seems to aim to provide a way to make all kind of collections accessible to the ItemsRepeater and usable for rendering. Adding and removing items wouldn't be an operation that we need in that case I think. But that's just my guess here.

active/ItemsSourceView-Indexof/ItemsSourceView-IndexOf.md
chingucoding: The scenario I want to cover is when one somehow get's a raw ItemsSourceView class (which I think is possible). And in that case, what should we do when someone invokes IndexOf on that raw class? Almost all functions just throw `hresult_not_implemented` . My question would be, whether we should do that or just return -1 in that case.

active/ItemsSourceView-Indexof/ItemsSourceView-IndexOf.md
MikeHillberg: Actually I should have said [IVectorView<T>](https://docs.microsoft.com/uwp/api/Windows.Foundation.Collections.IVectorView_T_). With IndexOf added, the only thing ItemsSourceView won't have is GetMany. Not a big problem, just would have been cleaner if we'd done it that way in the first place.

I don't think it's worth adding it now though, it's better to do it as you have it. The weird thing about doing it now is that it currently has a Count property, which matches [IReadOnlyCollection<T>.Count](https://docs.microsoft.com/dotnet/api/System.Collections.Generic.IReadOnlyCollection-1.Count), but not [IVectorView<T>.Size](https://docs.microsoft.com/uwp/api/Windows.Foundation.Collections.IVectorView-1.Size), so it would just be adding some confusion.

active/ItemsSourceView-Indexof/ItemsSourceView-IndexOf.md
MikeHillberg: I think [UIElementCollection](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.Controls.UIElementCollection) and [List<T>](https://docs.microsoft.com/dotnet/api/System.Collections.Generic.List-1) provide great precedence for returning -1 rather than failing.

active/ItemsSourceView-Indexof/ItemsSourceView-IndexOf.md
```diff
@@ -0,0 +1,68 @@
...
+The IndexOf function returns the index of the given element in the items source the ItemsSourceView is used for.
+Note that if an element is not contained in the items source, the function will return -1.
+> Spec note: returning -1 matches the behavior of for example IndexOf on [UIElementCollection](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.Controls.UIElementCollection) and List<T>.
+# Examples
+C# example
+
+```c#
+/// ************** Behavior ************ ///
+
+var itemsSourceView = new ItemsSourceView()
```
MikeHillberg: The ItemsSourceView needs to take in an IEnumerable to its constructor. Something like this would be easy to read in the docs:

```C#
var items = new List<string>() { "One", "Two" };
var itemsSourceView = new ItemsSourceView(items);
var secondItem = itemsSourceView.GetAt(1);
Assert.AreEqual(1, itemsSourceView.IndexOf(secondItem);
```

active/ItemsSourceView-Indexof/ItemsSourceView-IndexOf.md
```diff
@@ -0,0 +1,68 @@
...
+// Get an item
+var myItem = MyItemsSource.GetAt(1);
+
+Assert.AreEqual(1,itemsSourceView.IndexOf(myItem));
+
+/// ****** Use case in event handler ****** ///
+
+public void ItemInvoked(object sender, RoutedEventArgs e)
+{
+  int index = myItemsSourceView.IndexOf(sender as IInspectable);
```
MikeHillberg: Don't need to type-cast the sender

active/ItemsSourceView-Indexof/ItemsSourceView-IndexOf.md
```diff
@@ -0,0 +1,68 @@
...
+Assert.AreEqual(1,itemsSourceView.IndexOf(myItem));
+
+/// ****** Use case in event handler ****** ///
+
+public void ItemInvoked(object sender, RoutedEventArgs e)
+{
+  int index = myItemsSourceView.IndexOf(sender as IInspectable); 
+
+  // Do something with index, here e.g. :
+  Console.log("Item " + index + " of items source was invoked");
```
MikeHillberg:
```suggestion
  Debug.WriteLine( $"Item {index} of items source was invoked");
```

active/ItemsSourceView-Indexof/ItemsSourceView-IndexOf.md
chingucoding: Oh yes, that makes more sense. I guess there was no value in adding that previously (maybe the method was not needed internally?). The missmatch between Count and Size that we would create is probably not worth the benefit of having ItemsSourceView implement IVectorView.

active/ItemsSourceView-Indexof/ItemsSourceView-IndexOf.md
chingucoding: Oh I see. In that case it may be better to align with those then. My reasoning is that virtually every other method regardless of return value just throws `not implemented`, so having different behaviors of "Hey I do not provide this feature" seems a bit confusing/incoherent for me. But there are valid reasons for both sides here, so I leave it up to you.

active/ItemsSourceView-Indexof/ItemsSourceView-IndexOf.md
chingucoding: That sample is definitely better! I've added your sample instead of the previous one.

active/ItemsSourceView-Indexof/ItemsSourceView-IndexOf.md
```diff
@@ -0,0 +1,64 @@
...
+  // Do something with index, here e.g. :
+  Debug.WriteLine( $"Item {index} of items source was invoked");
+}
+
+```
+
+
+# Remarks
+
+The IndexOf function of ItemsSourceView base class throws a winrt hresult_not_implemented exceptions.
```
MikeHillberg: Confused about this line; when does it return an error?

chingucoding: From how we expose ItemsSourceView, this is a class not an interface. However it does not implement anything, doesn't have a reference to the IIterable it is dealing with. So if you get the base class and call IndexOf we have no way of running this operation. That's why we have to define what should happen. Currently all methods except Count throw a `hresult_not_implemented`.

Here the source:
https://github.com/microsoft/microsoft-ui-xaml/blob/master/dev/Repeater/ItemsSourceView.cpp

MikeHillberg: Those virtual *Core methods are implementation detail though, not something we want to talk about in the docs.microsoft.com pages for the API. For the public API we should say not-found returns -1, and then in the implementation we'll update the subclasses to make it true. Does that work?

chingucoding: I guess that makes sense. I've updated the spec.

active/ItemsSourceView-Indexof/ItemsSourceView-IndexOf.md
```diff
@@ -0,0 +1,58 @@
+# Background
+<!-- Use this section to provide background context for the new API(s) 
+in this spec. -->
+The Xaml [ItemsRepeater](https://docs.microsoft.com/uwp/api/Microsoft.UI.Xaml.Controls.ItemsRepeater) control displays a list of objects using a specified template (similar to a [ListView](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.Controls.ListView)). You provide the list of objects with the [ItemsRepeater.ItemsSource](https://docs.microsoft.com/uwp/api/Microsoft.UI.Xaml.Controls.ItemsRepeater.ItemsSource) property. The list need only be an IIterable but can be an IVector for better perf. ItemsRepeater exposes a read-only [ItemsRepeater.ItemsSourceView](https://docs.microsoft.com/uwp/api/Microsoft.UI.Xaml.Controls.ItemsRepeater.ItemsSourceView) property that provides a view on this list (an [ItemsSourceView](https://docs.microsoft.com/uwp/api/Microsoft.UI.Xaml.Controls.ItemsSourceView)), which has helpers like Count and GetAt over what might be just an IIterable.
```
oldnewthing: Should ItemsSourceView also implement IIterable, so you can
```cs
foreach (var item in ItemsRepeater.ItemsSource)
{
   ... something with item ...
}
```

Or is this not something people do frequently?

active/ItemsSourceView-Indexof/ItemsSourceView-IndexOf.md
```diff
@@ -0,0 +1,58 @@
...
+
+```
+
+
+# Remarks
+
+If an element is not present in the collection, the method returns -1.
+
+# API Notes
+ItemsSourceView.IndexOf(IInspectable item): Returns index in the items source. If an item is not in the items source, it will return -1.
```
oldnewthing: What if the item is present more than once?

Option 1: You get the lowest-index one.
Option 2: You get the highest-index one.
Option 3: You get an arbitrary one.

active/ItemsSourceView-Indexof/ItemsSourceView-IndexOf.md
```diff
@@ -0,0 +1,58 @@
...
+# API Details
+<!-- The exact API, in MIDL3 format (https://docs.microsoft.com/en-us/uwp/midl-3/) -->
+
+```MIDL
+
+unsealed runtimeclass ItemsSourceView: Windows.UI.Xaml.Interop.INotifyCollectionChanged
+{
+     // ...
+    Int32 IndexOf(IInspectable item);
```
oldnewthing: Presumably `null` is an acceptable value and returns the index of any null item in the collection.

active/ItemsSourceView-Indexof/ItemsSourceView-IndexOf.md
```diff
@@ -0,0 +1,58 @@
...
+var itemsSourceView = new ItemsSourceView(items);
+var secondItem = itemsSourceView.GetAt(1);
+Assert.AreEqual(1, itemsSourceView.IndexOf(secondItem);
+
+
+/// ****** Use case in event handler ****** ///
+
+public void ItemInvoked(object sender, RoutedEventArgs e)
+{
+  int index = myItemsSourceView.IndexOf(sender);
```
oldnewthing: IVectorView::IndexOf follows a different pattern. (Returns boolean indicating found/not found, and writes the index to an out parameter.)

Maybe ItemsSourceView should just implement `IVectorView<IInspectable>`? Then C# projection does the work of making it behave like an `IReadOnlyList<object>`. (Then again, `IReadOnlyList<object>` doesn't have an IndexOf method or extension method.)

active/ItemsSourceView-Indexof/ItemsSourceView-IndexOf.md
chingucoding: Looking at the implementation, this is nothing we have control over as we rely on the underlying collection to do this. So if the underlying collection returns the lowest index, so do we.

active/ItemsSourceView-Indexof/ItemsSourceView-IndexOf.md
chingucoding: That's a really good idea and also a really good question. There is definitely a point in having it be IIterable (IEnumerable in .Net). However if one would use the ItemsSourceView, they know its a IIterable, and thus could also do something like this:
```c#
foreach (var item in (IEnumerable)ItemsRepeater.ItemsSource)
{
   ... something with item ...
}
```

I am really not sure what would be best here. Something worth exploring might also be exposing the collection through a read only value, though that kinda defeats the point of ItemsSourceView in some areas.

active/ItemsSourceView-Indexof/ItemsSourceView-IndexOf.md
chingucoding: Should we mention this in the documentation (and should add to the spec)?

active/ItemsSourceView-Indexof/ItemsSourceView-IndexOf.md
chingucoding: My main concern is if there is any benefit to being an IVectorView<IInspectable> except adding `GetMany` and creating this association of "this is an IVectorView"? After all, currently we only implement 2/4 of this interface, and with this spec 3/4.

What would happen to the projection of IndexOf into .Net when we use the IVectorView definition? For me it seems a bit counterintuitive to use an `out int` to get the index, especially when the method is called **IndexOf** and returns a bool. Looking at the documentation of IReadOnlyList, it doesn't seem like it does provide a function for this.

active/ItemsSourceView-Indexof/ItemsSourceView-IndexOf.md
chingucoding: Currently, null is acceptabl as value, but would yield -1 regardless of null values being present in the collection. However I think this does not make much sense to treat `null` different here. Is this worth mentioning in the spec?

oldnewthing: I think the IndexOf method should treat null as just another value instead of special-casing it as always returning `-1`. That would make it align with `IVectorView.IndexOf`, and `List<T>.IndexOf`.

active/ItemsSourceView-Indexof/ItemsSourceView-IndexOf.md
chingucoding: In the case of an IIterable, we create a vector from the IIterable by iterating over the collection and adding it to an empty winrt::vector, so we would use winrt::vector::IndexOf .

active/ItemsSourceView-Indexof/ItemsSourceView-IndexOf.md
chingucoding: Totally agree with you here. I've updated the spec to point that out.

active/ItemsSourceView-Indexof/ItemsSourceView-IndexOf.md
ranjeshj: This was one of the reasons we did not add it initially, but later found out that returning the lowest-index one is reasonable, and in cases where there are no duplicates, this is a non issue. @chingucoding is correct about collections that are just Iterable, we iterate over it and make a copy. Where the provided collection supports IndexOf, it will be forwarded to its implementation.

chingucoding:
So during testing of the existing internal IndexOf, we noticed one problem: Having a collection consisting only pass by value types such as integers, IndexOf does not work if we pass the collection as IIterable.

See following example:
```c#
var itemSourceView = new ItemsSourceView(Enumerable.Rang(0,10));

// The following is false!
// IndexOf(1) returns -1 
Assert.AreEqual(1,itemSourceView.IndexOf(1));
``` 

When passing pass by value types, they get boxed as objects. This happens both when creating an ItemsSourceView and when we call IndexOf. However those boxed objects will not be the same, so finding pass by value types using IndexOf fails.

The big question: Should we support it, or just point out in the documentation that this won't work?
Pass by value types are not a common scenario for ItemsSourceView, so is supporting them really necessary or beneficial?

active/ItemsSourceView-Indexof/ItemsSourceView-IndexOf.md
MikeHillberg: IVector projects to C# as IList, which has an IndexOf, unlike the ICollection interface it derives from.

But IVectorView projects to IReadOnly**Collection**, which was invented for that purpose. But maybe it should have projected to an IReadOnlyList instead.

MikeHillberg: So to the original point, we don't have one precedent we can follow; C++ precedence (IVector) is to return a bool, C# precedence (IList) is to return an int. I think we should go with the C# precedence (int), consistent with markup where we have x:null rather than x:nullptr.

MikeHillberg:
Good point. This case is actually handled in the analogous [ItemCollection.IndexOf](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.Controls.ItemCollection.IndexOf), and similarly by [Selector.SelectedItem](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.Controls.Primitives.Selector.SelectedItem), by unboxing and checking for the fundamental value types and for interfaces. 

I don't know if there's any public API that exposes that functionality though. Does [TabView.SelectedItem](https://docs.microsoft.com/uwp/api/Microsoft.UI.Xaml.Controls.TabView.SelectedItem) have code for it?

chingucoding:
Since the TabView uses a ListView internally, we do whatever ListView does to keep this up to date. 

The main problem I see with doing this manually, is that we would have to unbox all the time, pushing down performance. Since the ItemsRepeater that uses this class, is mainly designed for large collections, this could really make a significant difference. And since most of the time, people actually pass objects and not primitive types, I am not sure if it's worth dropping performance for this "corner case".

If somebody wants to use IndexOf with pass by value types, they could use a different collection. We only would have a say in this if we get passed an IIterable, in other cases we don't even have an option on what we do, we just pass arguments down. So we might just say: 

Pass by value types are not supported, if you want to use them use ... as itemssource.

ranjeshj:
> Since the TabView uses a ListView internally, we do whatever ListView does to keep this up to date.
> 
> The main problem I see with doing this manually, is that we would have to unbox all the time, pushing down performance. Since the ItemsRepeater that uses this class, is mainly designed for large collections, this could really make a significant difference. And since most of the time, people actually pass objects and not primitive types, I am not sure if it's worth dropping performance for this "corner case".
> 
> If somebody wants to use IndexOf with pass by value types, they could use a different collection. We only would have a say in this if we get passed an IIterable, in other cases we don't even have an option on what we do, we just pass arguments down. So we might just say:
> 
> Pass by value types are not supported, if you want to use them use ... as itemssource.

Right. It is quite common for someone to 'try' an enumerable of integers, however in most cases, the collection is going to be a non-value type. we still have the out of saying, if you want to use its, use a collection of some sort (like observablecollection) and In that case we will delegate to the collection. I agree that we should not be doing extra work for this case that might impact perf. I have not checked, but ListView likely has the same issue...

MikeHillberg:
> ListView likely has the same issue

ListView actually works, both SelectedItem and Items.IndexOf. The other case boxing shows up is in a list of structs, like a List&lt;Point&gt;. This will be easier with WinUI3 where we could share the code that ListView uses more easily. Maybe doc it until then.

chingucoding:
> ListView actually works, both SelectedItem and Items.IndexOf. The other case boxing shows up is in a list of structs, like a List<Point>. This will be easier with WinUI3 where we could share the code that ListView uses more easily. Maybe doc it until then.

I've updated the spec to point out how pass by value types behave and how one can use them.

ranjeshj:
> > ListView actually works, both SelectedItem and Items.IndexOf. The other case boxing shows up is in a list of structs, like a List. This will be easier with WinUI3 where we could share the code that ListView uses more easily. Maybe doc it until then.
> 
> I've updated the spec to point out how pass by value types behave and how one can use them.

I did not realize that ListView was working. I think we should make it work here also. I am ok with creating an issue and following up on that since it does not affect the API surface. I can see structs being used in this scenario where this can be a problem.

chingucoding:
> > > ListView actually works, both SelectedItem and Items.IndexOf. The other case boxing shows up is in a list of structs, like a List. This will be easier with WinUI3 where we could share the code that ListView uses more easily. Maybe doc it until then.
> > 
> > 
> > I've updated the spec to point out how pass by value types behave and how one can use them.
> 
> I did not realize that ListView was working. I think we should make it work here also. I am ok with creating an issue and following up on that since it does not affect the API surface. I can see structs being used in this scenario where this can be a problem.

So should we support pass by value in the ItemsSourceView then? In that case, I think I could also fix this with the PR that we already have in WinUI that would add that API.

ranjeshj:
> > > > ListView actually works, both SelectedItem and Items.IndexOf. The other case boxing shows up is in a list of structs, like a List. This will be easier with WinUI3 where we could share the code that ListView uses more easily. Maybe doc it until then.
> > > 
> > > 
> > > I've updated the spec to point out how pass by value types behave and how one can use them.
> > 
> > 
> > I did not realize that ListView was working. I think we should make it work here also. I am ok with creating an issue and following up on that since it does not affect the API surface. I can see structs being used in this scenario where this can be a problem.
> 
> So should we support pass by value in the ItemsSourceView then? In that case, I think I could also fix this with the PR that we already have in WinUI that would add that API.

Let's follow up with a separate PR for that specific change. We could possibly share some of this code in WinUI3. Thanks.

chingucoding:
So is this spec then fine as it currently is or should I revert my last commit adding a comment regarding pass by value?

active/ItemsSourceView-Indexof/ItemsSourceView-IndexOf.md
```diff
@@ -41,6 +41,9 @@ If an element is not present in the collection, the method returns -1.
 
 # API Notes
 ItemsSourceView.IndexOf(IInspectable item): Returns index in the items source. If an item is not in the items source, it will return -1. Null is an acceptable value for IndexOf and calling IndexOf with null will return the index of null in the ItemsSource or -1 if the ItemsSource does not contain null.
+
+Note that to use `IndexOf` with pass by value types (e.g. int,float,etc) you need to create the ItemsSourceView with a collection that uses unboxing for values passed to `IndexOf`.
+Otherwise calling `IndexOf` will always return `-1` regardless of the items contained in the collection.
```
MikeHillberg: I understand what this is about, but I don't understand the resolution. What does it mean that I need to use unboxing for values passed to IndexOf?

active/ItemsSourceView-Indexof/ItemsSourceView-IndexOf.md
MikeHillberg:
```suggestion
ItemsSourceView.IndexOf(IInspectable item): Returns index in the items source. If an item is not in the items source, it will return -1.

If the ItemsSourceView's underlying collection supports [IVector<T>.IndexOf](https://docs.microsoft.com/uwp/api/Windows.Foundation.Collections.IVector-1.IndexOf) or [IVectorView<T>.IndexOf](https://docs.microsoft.com/uwp/api/Windows.Foundation.Collections.IVectorView-1.IndexOf), it will be used and that value returned. If the underlying collection only supports [IIterable<T>](https://docs.microsoft.com/uwp/api/Windows.Foundation.Collections.IIterable_T_), then the first matching value from its enumeration will be returned.
```

MikeHillberg: Added new suggested text in a separate comment to cover this.

active/ItemsSourceView-Indexof/ItemsSourceView-IndexOf.md
MikeHillberg: Or can we remove this from the API doc because we're going to address it in the code?

chingucoding: Oh, yes my wording is a bit difficult here and definitely needs improvement. So the idea is: primitive types get wrapped in objects (aka boxing). To be able to "find" them again, you need to get the primitive of those values again, that is compare by value and not by reference. So you need to unbox the value passed to IndexOf to be able to properly find a pass by value type in the collection.

If you have an idea how we could convey that in the most understandable and readable way, please let me know.

active/ItemsSourceView-Indexof/ItemsSourceView-IndexOf.md
chingucoding: Unfortunately, GitHub won't let me commit that here, so I've added it manually.

active/ItemsSourceView-Indexof/ItemsSourceView-IndexOf.md
chingucoding: I am not sure if we are gonna do that. Currently, we just don't deal with pass by value's. And in the case of any collection that has IndexOf, we don't have any control anyway.

MikeHillberg: For the user of this API, I think all we can do is state a restriction that it doesn't work. But are we going to address it in the code instead?

chingucoding: We could circumvent that for simple IIterable's by using our internal vector and go over the values manually instead of using the build in Vector.IndexOf function. But would it be much of a gain here?

oldnewthing: Note that `string` is a value type in C++/WinRT but a reference type in C# and C++/CX, so saying "pass by value types" is ambiguous.

I think this discussion in the documentation is going too deep and will end up confusing people. How about

> If the `ItemsSource` supports `IVector<Object>`, `IVectorView<Object>`, or `IObservableVector<Object>`, then the collection's `IndexOf` method is used.
>
> If the `ItemsSource` supports only `IIterable<Object>`, then the `IndexOf` method uses object identity. (Note that for strings and value types, object identity is performed on the boxed value.)

Because it's not the case that `IndexOf` will always return `-1` on value types. You could pass the same box back into `IndexOf` and it will find it.

chingucoding: Great wording @oldnewthing ! I've merged it into the existing text.

ranjeshj: Should we also add IEnumerable to the list along with IIterable<Object> for C# developers ?

chingucoding: Good question. Looking at the documentation, wouldn't IIterable be IEnumerable, because C# would use the projected type? 
I've updated the comment to note that IEnumerable is the appropiate .Net type for clarity.

ranjeshj: Yes they are. Since we have a lot of users in C#, calling out explicitly might make sense. @MikeHillberg @oldnewthing would you agree ?

oldnewthing: I think it's reasonable to mention the C# equivalents. Would need to wordsmith a bit to avoid separating the antecedent from consequent.

> The `IndexOf` method uses the `IndexOf` method of the `ItemsSource` if it supports `IVector<Object>`, `IVectorView<Object>`, or `IObservableVector<Object>` (in C#, `IList<object>`, `IReadOnlyList<object>`, or `IObservableVector<object>`).
>
> The `IndexOf` method uses object identity if the `ItemsSource` supports only `IIterable<Object>` (in C#, `IEnumerable<object>`). Note that for strings and value types, object identity is performed on the boxed value.

chingucoding: Should we state "in C# ..." or "in .Net ..." for the differences? After all there are other .Net languages which could potentially use the classes.

Also, IReadOnlyList (the projection of IVectorView) does not provide IndexOf, so we need to leave that out unfortunately. Instead of having that long lists, would it be easier to have a document for C++ and one for .Net listing the interfaces where we would use IndexOf instead of implementing it ourself?

Something we could also do is simply say: In case of IIterable, we provide it, and in any interface that provides IndexOf, we use that. That way we wouldn't have to provide a list of interfaces that might get outdated

MikeHillberg: I just left it at IVector for this spec, because I don't understand the editorial approach to this problem on docs.microsoft.com. When we copy this there we can get help on it from them.

active/ItemsSourceView-Indexof/ItemsSourceView-IndexOf.md
```diff
@@ -0,0 +1,63 @@
...
+
+
+# Remarks
+
+If an element is not present in the collection, the method returns -1.
+
+# API Notes
+ItemsSourceView.IndexOf(IInspectable item): Returns index in the items source. If an item is not in the items source, it will return -1. Null is an acceptable value for IndexOf and calling IndexOf with null will return the index of null in the ItemsSource or -1 if the ItemsSource does not contain null.
+
+If the ItemsSourceView's underlying collection supports [IVector<T>.IndexOf](https://docs.microsoft.com/uwp/api/Windows.Foundation.Collections.IVector-1.IndexOf) or [IVectorView<T>.IndexOf](https://docs.microsoft.com/uwp/api/Windows.Foundation.Collections.IVectorView-1.IndexOf), it will be used and that value returned.
```
MikeHillberg: These interfaces are specifically IInspectable, not T, I assume?

chingucoding: Apparently they are not, however calling them still works as expected (at least in our unit tests).

MikeHillberg: It looks like in the implementation it's an IVector<IInspectable>. If the unit test is passing a List<T>, that class implements IList, and so the collection will project to WinRT as implementing both IVector<IInspectable> and IVector<T>. Do we actually use IVectorView's IndexOf, or does that fall back to the IIterable?

chingucoding: The order of casting is: IVector->IBindableVector->IIterable->IBindableIterable
The first cast that succeeds will be used. So actually we would ignore any pure IVectorView. An C# IList would be used as IVector then.

MikeHillberg: And specifically IVector<IInspectable>?

chingucoding: If you pass an IVector, we use that IVector directly. According to the docmentation, the .Net projection of IVector<T> is IList, so IList would be used as IVector.

MikeHillberg: The collection projections are complicated because in c# we want to use the language's standard interfaces. There isn't actually an IVector without the T, and the c# List&lt;T&gt; _class_ makes things even more interesting by implementing both IList&lt;T&gt; and IList.

| WinRT | C# |
| - | - |
| IVector&lt;T&gt; | IList&lt;T&gt; |
| IVectorView&lt;T&gt; | IReadOnlyCollection&lt;T&gt; |
| IBindableVector | IList |

I think diving into all of that is getting too complicated for this method description. It sounds like the correct technical description is that it uses the underlying IndexOf if the collection supports IVector&lt;T&gt; or IBindableVector or IVector&lt;IInspectable&gt;, otherwise it enumerates in order using IIterable&lt;IInspectable&gt; or IBindableIterable. And then translate this into c# terms, and I think it's just getting too complicated.

How about we just leave this at:

> If the ItemsSourceView's underlying collection supports a compatible IndexOf, it will be used and that value returned. Otherwise the collection will be enumerated and the index of the first matching item returned.

And then if looks too vague in the docs we can add a link on "compatible" to a page that generally describes the supported collection interfaces and how they project to different languages.

chingucoding: Yes that whole projection situation is to much for the method description, I agree,

However, your proposed text is not entirely correct, since we still use Vector.IndexOf, if there is no suitable IndexOf we could use. So I think either we should change the way of handling this or we should change the section to something more like this:

> If the ItemsSourceView's underlying collection supports a compatible IndexOf, it will be used and that value returned. Otherwise the first matching value from its enumeration will be returned and IndexOf uses object identity. Note that for strings and value types, object identity is performed on the boxed value.

What do you think about this @mikehillberg ?

MikeHillberg: OK, looks good to me :)

chingucoding: Great, doc's updated now.

active/ItemsSourceView-Indexof/ItemsSourceView-IndexOf.md
```diff
@@ -0,0 +1,61 @@
...
+
+
+# Remarks
+
+If an element is not present in the collection, the method returns -1.
+
+# API Notes
+ItemsSourceView.IndexOf(IInspectable item): Returns index in the items source. If an item is not in the items source, it will return -1. Null is an acceptable value for IndexOf and calling IndexOf with null will return the index of null in the ItemsSource or -1 if the ItemsSource does not contain null.
+
+If the ItemsSourceView's underlying collection supports a compatible `IndexOf`, it will be used and that value returned. Otherwise the first matching value from its enumeration will be returned and `IndexOf` uses object identity. Note that for strings and value types, object identity is performed on the boxed value.
```
MikeHillberg, line 45: Is string really an issue? String should project as HSTRING rather than being boxed. Case in point, does the above string-based sample work?

active/ItemsSourceView-Indexof/ItemsSourceView-IndexOf.md
```diff
@@ -0,0 +1,61 @@
...
+
+var items = new List<string>() { "One", "Two" };
+var itemsSourceView = new ItemsSourceView(items);
+var secondItem = itemsSourceView.GetAt(1);
+Assert.AreEqual(1, itemsSourceView.IndexOf(secondItem);
+
+
+/// ****** Use case in event handler ****** ///
+
+public void ItemInvoked(object sender, RoutedEventArgs e)
```
MikeHillberg, line 27: What event is this listening to? Is the sender really an item, rather than an element?

chingucoding: I think in the case of a collection of controls, instead of wrapping them, we render them directly, so there is a scenario where the sender could be the item.

active/ItemsSourceView-Indexof/ItemsSourceView-IndexOf.md
chingucoding: Turns out, when trying to pass an IEnumerable to the ItemsSourceView and you call IndexOf with a string, it returns -1.

MikeHillberg: Oh, sorry, right. IEnumerable is like IEnumerable&lt;Object&gt; and forces the string to be boxed. So we need to update the string-based example until this is fixed?

active/ItemsSourceView-Indexof/ItemsSourceView-IndexOf.md
MikeHillberg: For the example we should be specific about what event it's listening to. We want the example code to be highly copy/pasteable. What was the code you were writing when you originally needed this?

active/ItemsSourceView-Indexof/ItemsSourceView-IndexOf.md
chingucoding: Since in the sample, we are passing a list with type string, that sample works as the string won't get boxed.

active/ItemsSourceView-Indexof/ItemsSourceView-IndexOf.md
chingucoding: There wasn't a specific reason. Since the context would be a bit large for a sample, we can also get rid of it. What do you think?

ranjeshj: I think the event handler code is a bit confusing. Perhaps removing that part would be better. We could also make it more concrete and show the event handler for one of ItemsRepeater's events like ElementPrepared or ElementIndexChanged if we want to show usage in a handler.

MikeHillberg: I think it's OK to remove it.

active/ItemsSourceView-Indexof/ItemsSourceView-IndexOf.md
MikeHillberg: List<string> should show up in the InspectingDataSource constructor as not implementing implementing IBindableVector but not IVector<IInspectable>. And so boxed values, and then IndexOf should be receiving a different box instance, and therefore not find a match. Is that not what you're seeing?

active/ItemsSourceView-Indexof/ItemsSourceView-IndexOf.md
chingucoding: I've removed it now.

active/ItemsSourceView-Indexof/ItemsSourceView-IndexOf.md
chingucoding: If you copy the sample, it works just as expected. Calling IndexOf in that case returns the correct index.

MikeHillberg: I see, the above sample works for any value type, because the input supports IndexOf. It's the IEnumerable case where the value type doesn't work yet.

Can we update the wording?

```suggestion
If the ItemsSourceView's underlying collection supports a compatible `IndexOf`, it will be used and that value returned. Otherwise the first matching value from its enumeration will be returned. Note in this case that for strings and value types, -1 will always be returned, because items will be compared by boxed references rather than the values.
```

PR #77, opened 2020-03-25
TabView v2 (#70)

stmoy:
* Move completed TabView v1 docs to archive
* Document the TabView v2 features


active/TabView/TabView_v2.md
```diff
@@ -0,0 +1,163 @@
...
+[WUXC_VERSION_MUXONLY]
+[webhosthidden]
+enum TabViewWidthMode
+{
+    Equal = 0,
+    SizeToContent = 1,
+    Compact = 2,
+};
+
+[WUXC_VERSION_PREVIEW]
```
chingucoding: Should this be preview or should it be [WUXC_VERSION_MUXONLY] ? I thought the goal is to finalize this API so we can mark it as public, having it be marked as preview here is a bit counterintuitive in my opinion.

active/TabView/TabView_v2.md
```diff
@@ -0,0 +1,142 @@
...
+};
+
+[WUXC_VERSION_MUXONLY]
+[webhosthidden]
+[contentproperty("TabItems")]
+unsealed runtimeclass TabView : Windows.UI.Xaml.Controls.Control
+{
+
+    [WUXC_VERSION_PREVIEW]
+    [MUX_DEFAULT_VALUE("winrt::TabViewCloseButtonOverlayMode::Always")]
```
oldnewthing, line 116: Shouldn't the default be `Auto`, not `Always`? Otherwise, what's the point of `Auto`?

chingucoding: Yes, `Auto` should be the default

active/TabView/TabView_v2.md
stmoy: cc: @Mikehillberg - I think it should be WUXC_VERSION_MUXONLY?

stmoy: cc: @Ranjeshj

ranjeshj: Yes. Once the review is complete, we can switch the attribute from preview to WUXC_VERSION_MUXONLY. Thanks!

PR #84, opened 2020-04-29
XamlServiceProvider API for markup extensions

MikeHillberg:


active/XamlServiceProvider/XamlServiceProviderApi.md
```diff
@@ -0,0 +1,251 @@
...
+_This spec adds a serviceProvider to the existing [MarkupExtension](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.Markup.MarkupExtension) class. ([Proposal](https://github.com/microsoft/microsoft-ui-xaml/issues/741))_
+
+Xaml markup has a [markup extension](https://docs.microsoft.com/en-us/dotnet/desktop-wpf/xaml-services/markup-extensions-overview) feature which is a builder pattern, similar to a .Net StringBuilder. It lets you create, modify, and then retrieve a value.
+
+When you write a custom markup extension (see the Description section below), you override the `ProvideValue` method to return the built value. The ProvideValue method is different in  Xaml (WinUI) than it is in WPF, because WPF passes an extra [IServiceProvider](http://msdn.microsoft.com/library/System.IServiceProvider) parameter:
+
+Xaml:
+
+```cs
+protected override object ProvideValue(IServiceProvider serviceProvider)
```
alwu-msft, line 13: This should be parameterless.

active/XamlServiceProvider/XamlServiceProviderApi.md
```diff
@@ -0,0 +1,251 @@
...
+{
+    return DateTime.Now.ToString(Format);
+}
+```
+
+The service provider can be used, for example, to discover the property being assigned to.
+
+This spec is adding a new version of ProvideValue that passes a service provider.
+
+> To do: This is defining an IXamlServiceProvider, can/should this project in .Net to [System.IServiceProvider](http://msdn.microsoft.com/library/System.IServiceProvider)?
```
alwu-msft, line 32: Definitely!

active/XamlServiceProvider/XamlServiceProviderApi.md
```diff
@@ -0,0 +1,251 @@
...
+
+## MarkupExtension.ProvideValue virtual method
+
+A markup extension overrides the ProvideValue method, and returns a value from it that the Xaml loader sets as a property value.
+
+There are two versions of ProvideValue, one that passes no parameters and one that passes a service provider. From the service provider, a markup extension implementation can retrieve the following services:
+* `IProvideValueTarget`, which provides information about the type/property the markup extension is being applied to. This aligns with WPF's [IProvideValueTarget](https://docs.microsoft.com/dotnet/api/System.Windows.Markup.IProvideValueTarget).
+* `IRootObjectProvider`, which provides a reference to the root object in the markup. This aligns with WPF's [IRootObjectProvider](https://docs.microsoft.com/dotnet/api/System.Xaml.IRootObjectProvider).
+* `IUriContext` , which provides the base URI of the markup. This aligns with WPF's [IUriContext](https://docs.microsoft.com/dotnet/api/System.Windows.Markup.IUriContext).
+
```
alwu-msft, line 72: Was ['IXamlTypeResolver'](https://docs.microsoft.com/en-us/dotnet/api/system.windows.markup.ixamltyperesolver?view=netcore-3.1) inadvertently omitted from this list?

active/XamlServiceProvider/XamlServiceProviderApi.md
```diff
@@ -0,0 +1,251 @@
...
+> Todo
+
+# API Notes/Remarks
+
+## IXamlServiceProvider interface
+Gets the service object of the specified type.
+
+There are two versions of the virtual ProvideValue method, you should only override one.
+
+> Issue: how do we know which version has been overridden? We have the same scenario with [DataTemplateSelector.SelectTemplateCore](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.Controls.DataTemplateSelector.SelectTemplateCore). Proposal: the virtual (base) version of the parameterless override returns an internal sentinal value, and is called first.
```
alwu-msft, line 162: The base class's implementation of ProvideValue(IXamlServiceProvider) invokes ProvideValue(). So if the new overload isn't overridden, the parameterless version of the function will be automatically invoked.

active/XamlServiceProvider/XamlServiceProviderApi.md
```diff
@@ -0,0 +1,251 @@
+
+# Background
+
+_This spec adds a serviceProvider to the existing [MarkupExtension](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.Markup.MarkupExtension) class. ([Proposal](https://github.com/microsoft/microsoft-ui-xaml/issues/741))_
+
+Xaml markup has a [markup extension](https://docs.microsoft.com/en-us/dotnet/desktop-wpf/xaml-services/markup-extensions-overview) feature which is a builder pattern, similar to a .Net StringBuilder. It lets you create, modify, and then retrieve a value.
+
+When you write a custom markup extension (see the Description section below), you override the `ProvideValue` method to return the built value. The ProvideValue method is different in  Xaml (WinUI) than it is in WPF, because WPF passes an extra [IServiceProvider](http://msdn.microsoft.com/library/System.IServiceProvider) parameter:
+
+Xaml:
```
mrlacey, line 10: Label this `WinUI` as the actual code example is in C#

active/XamlServiceProvider/XamlServiceProviderApi.md
```diff
@@ -0,0 +1,251 @@
+
+# Background
+
+_This spec adds a serviceProvider to the existing [MarkupExtension](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.Markup.MarkupExtension) class. ([Proposal](https://github.com/microsoft/microsoft-ui-xaml/issues/741))_
+
+Xaml markup has a [markup extension](https://docs.microsoft.com/en-us/dotnet/desktop-wpf/xaml-services/markup-extensions-overview) feature which is a builder pattern, similar to a .Net StringBuilder. It lets you create, modify, and then retrieve a value.
+
+When you write a custom markup extension (see the Description section below), you override the `ProvideValue` method to return the built value. The ProvideValue method is different in  Xaml (WinUI) than it is in WPF, because WPF passes an extra [IServiceProvider](http://msdn.microsoft.com/library/System.IServiceProvider) parameter:
```
mrlacey, line 8: Referring to ' Xaml (WinUI)' gives the impression that you're thinking of WinUI as a replacement to XAML.
XAML is a markup language used by WinUI and WPF (& Xamarin.Forms, and more.)
Using "XAML" to mean "WinUI" limits the scope of WinUI and is potentially confusing.

active/XamlServiceProvider/XamlServiceProviderApi.md
```diff
@@ -0,0 +1,251 @@
...
+        Go bears!
+    </Button>
+    <Rectangle x:Name="SolidColor" Fill="{local:BrushSelector Color=Green}" />
+</StackPanel>
+```
+
+
+## IRootObjectProvider
+
+> Some more realistic example?
```
mrlacey, line 121: Get the `Title` from the page (if the rootobject is a `Page`)?

active/XamlServiceProvider/XamlServiceProviderApi.md
```diff
@@ -0,0 +1,251 @@
...
+## IUriContext 
+
+> Todo
+
+# API Notes/Remarks
+
+## IXamlServiceProvider interface
+Gets the service object of the specified type.
+
+There are two versions of the virtual ProvideValue method, you should only override one.
```
mrlacey, line 160: Can a warning be raised if a developer overrides both versions?

active/XamlServiceProvider/XamlServiceProviderApi.md
MikeHillberg: Page doesn't have a Title in WinUI

active/XamlServiceProvider/XamlServiceProviderApi.md
MikeHillberg: I don't see a way, all the base class can do is call it. This is the same "fun" we had with DataTemplateSelector.SelectTemplateCore.

active/XamlServiceProvider/XamlServiceProviderApi.md
MikeHillberg: But how does the base know if the new overload isn't overridden? We could potentially separate the Overrides interfaces and do a QI test, but we don't want that many interfaces, and we can't guarantee how a language projection will implement the overrides.

StephaneDelcroix:
I'm wondering why you choose to name this `IXamlServiceProvider` instead of already existing `IServiceProvider`

active/XamlServiceProvider/XamlServiceProviderApi.md
```diff
@@ -0,0 +1,272 @@
...
+{
+    return ...
+}
+```
+
+The service provider can be used, for example, to discover the property being assigned to.
+
+This spec is adding a new version of ProvideValue to WinUI that passes a similar service provider. All the APIs match WPF, except where WPF uses the System.IServiceProvider, WinUI will use Microsoft.UI.Xaml.Markup.IXamlServiceProvider.
+
+> To do: Add an Issue to cs/winrt to project IServiceProvider to .Net5 as [System.IServiceProvider](http://msdn.microsoft.com/library/System.IServiceProvider)?
```
MikeHillberg, line 32: Also, should this be in the Xaml namespace or someplace more general?

alwu-msft: What namespace did we put our versions of INPC and INCC under?

MikeHillberg: Xaml. But for example we moved [Color](https://docs.microsoft.com/uwp/api/Windows.UI.Color) out of the Xaml namespace, since it's useful outside that one domain. There's nothing specific to Xaml about the service provider pattern.

alwu-msft: I think what I was trying to ask is: how does CsWinrt project INPC for a non-WinUI project? Is that project still required to include WinUI so that it knows about Microsoft.UI.Xaml.Data.INotifyPropertyChanged? Or is that a nonsensical question?

MikeHillberg: That's a sensical question; I don't think INPC should have been put into the Xaml namespace.

active/XamlServiceProvider/XamlServiceProviderApi.md
```diff
@@ -0,0 +1,272 @@
...
+The following example shows a custom markup extension whose provided value changes based on the specific property being targeted by the custom markup extension. In this case, the developer wants something that will automatically use an AcrylicBrush for the [Control.Background](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.Controls.Control.Background) or [Border.Background](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.Controls.Border.Background) properties, but uses a [SolidColorBrush](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.Media.SolidColorBrush) for all other properties. The ProvideValue implementation deterimines what value to return by using the IProvideValueTarget it gets from the IXamlServiceProvider parameter passed to ProvideValue.
+
+```cs
+public class BrushSelectorExtension : MarkupExtension
+{
+    public Color Color { get; set; }
+
+    protected override object ProvideValue(IXamlServiceProvider serviceProvider)
+    {
+        Brush brushToReturn = new SolidColorBrush() { Color = Color }; ;
```
oldnewthing, line 77: `; ;` typo. Also, `Color = Color` looks weird but it's okay. Is this a common idiom? If not, we may want to clarify `Color = this.Color`.

active/XamlServiceProvider/XamlServiceProviderApi.md
```diff
@@ -0,0 +1,272 @@
...
+* `IProvideValueTarget`, which provides information about the type/property the markup extension is being applied to. This aligns with WPF's [IProvideValueTarget](https://docs.microsoft.com/dotnet/api/System.Windows.Markup.IProvideValueTarget).
+* `IRootObjectProvider`, which provides a reference to the root object in the markup. This aligns with WPF's [IRootObjectProvider](https://docs.microsoft.com/dotnet/api/System.Xaml.IRootObjectProvider).
+* `IUriContext` , which provides the base URI of the markup. This aligns with WPF's [IUriContext](https://docs.microsoft.com/dotnet/api/System.Windows.Markup.IUriContext).
+* `IXamlTypeResolver`, which binds a Xaml markup name to a type. This aligns with WPF's [IXamlTypeResolver](https://docs.microsoft.com/en-us/dotnet/api/system.windows.markup.ixamltyperesolver).
+
+There are two overloads of the virtual ProvideValue method, with/without the IXamlServiceProvider parameter. During Xaml markup loading, ProvideValue(IXamlServiceProvider) is called. Its virtual implementatiohn calls ProvideValue(). So not overriding ProvideValue(IXamlServiceProvider), or calling the base implementation of it, calls ProvideValue().
```
oldnewthing, line 196: TYPO: `implementation`.

active/XamlServiceProvider/XamlServiceProviderApi.md
```diff
@@ -0,0 +1,272 @@
...
+            Background="{local:BrushSelector Color=Gold}">
+        Go bears!
+    </Button>
+    <Rectangle x:Name="SolidColor" Fill="{local:BrushSelector Color=Green}" />
+</StackPanel>
+```
+
+## IRootObjectProvider
+
+> Some more realistic example?
```
oldnewthing, line 109: How about using the RootObject as a default data source?

```
public class TestMarkupExtension : MarkupExtension
{
    string PropertyName { get; set; }
    object Source { get; set; }
    public override object ProvideValue(IXamlServiceProvider serviceProvider)
    {
        var source = Source ?? ((IRootObjectProvider)serviceProvider.GetService(typeof(IRootObjectProvider))).RootObject;
        var info = source.GetType().GetProperty(PropertyName);
        return info.GetValue(target);
    }
}
```

active/XamlServiceProvider/XamlServiceProviderApi.md
```diff
@@ -0,0 +1,272 @@
...
+The following example shows a custom markup extension that returns a string describing the object at the root of the markup. It does this by getting the IRootObjectProvider from the IXamlServiceProvider parameter passed to ProvideValue.
+
+With this markup extension:
+
+```cs
+public class TestMarkupExtension : MarkupExtension
+{
+    public override object ProvideValue(IXamlServiceProvider serviceProvider)
+    {
+        var target = serviceProvider.GetService(typeof(IRootObjectProvider)) as IRootObjectProvider;
```
oldnewthing, line 120: `target` is a bad name for this variable because it's not the target of the markup extension.

active/XamlServiceProvider/XamlServiceProviderApi.md
```diff
@@ -0,0 +1,272 @@
...
+public class BrushSelectorExtension : MarkupExtension
+{
+    public Color Color { get; set; }
+
+    protected override object ProvideValue(IXamlServiceProvider serviceProvider)
+    {
+        Brush brushToReturn = new SolidColorBrush() { Color = Color }; ;
+
+        var provideValueTarget = serviceProvider.GetService(typeof(IProvideValueTarget))
+            as IProvideValueTarget;
```
oldnewthing, line 80: If you're going to crash if `IProvideValueTarget` is not present, then use a cast instead of `as` so the crash is right at the line of code that caused the problem, instead of deferring the crash until some later point. (Repeat for other uses of "as" where the result is assumed to be non-null.)

active/XamlServiceProvider/XamlServiceProviderApi.md
oldnewthing, line 196: If my MarkupProvider overrides the `ProvideValue(IXamlServiceProvider)` but not the `ProvideValue()`, what happens? Are there ways that the 0-parameter could get called even if the 1-parameter one is present? Say, because some other app code grabs the MarkupProvider and starts calling ProvideValue directly?

MikeHillberg: This paragraph tries to answer this: the virtual implementation of ProvideValue(IXamlServiceProvider) calls ProvideValue(), which returns null.

active/XamlServiceProvider/XamlServiceProviderApi.md
```diff
@@ -0,0 +1,272 @@
...
+# API Notes/Remarks
+
+## MarkupExtension.ProvideValue(IXamlServiceProvider) virtual method
+
+When implemented in a derived class, returns an object that is provided as the value of the target property for this markup extension.
+
+**Remarks**
+
+The IXamlServiceProvider argument can be used to retrieve the following services:
+* `IProvideValueTarget`, which provides information about the type/property the markup extension is being applied to. This aligns with WPF's [IProvideValueTarget](https://docs.microsoft.com/dotnet/api/System.Windows.Markup.IProvideValueTarget).
```
jonwis, line 191: Lines 190-196 seem like they're on the wrong place? Shouldn't they be under the `IXamlServiceProvider` section?

active/XamlServiceProvider/XamlServiceProviderApi.md
```diff
@@ -0,0 +1,272 @@
...
+    {
+        protected MarkupExtension();
+    
+        overridable Object ProvideValue();
+
+        // New
+        overridable Object ProvideValue(Windows.UI.Xaml.IXamlServiceProvider serviceProvider);
+    };
+
+    runtimeclass ProvideValueTargetProperty
```
jonwis, line 260: What other Targets are expected?

active/XamlServiceProvider/XamlServiceProviderApi.md
MikeHillberg, line 32: Typo: IServiceProvider should be IXamlServiceProvider

active/XamlServiceProvider/XamlServiceProviderApi.md
MikeHillberg: Do you mean what other types are expected to be queried for?

active/XamlServiceProvider/XamlServiceProviderApi.md
MikeHillberg: Actually I'm not understanding this. If getting a PropertyInfo from Source, and then using it to read from 'target'. I'm not sure if that's IPropertyValueTarget, or was meant to be 'source'. It kind of looks like it's manually doing an x:Bind.

active/XamlServiceProvider/XamlServiceProviderApi.md
```diff
@@ -0,0 +1,272 @@
...
+The following example shows a custom markup extension that converts a relative URI to an absolute URI. It does this by getting the IUriContext from the IXamlServiceProvider parameter passed in to ProvideValue. IUriContext provides the URI of the markup being loaded.
+
+This markup extension enables the following markup, where a property requires an absolute URI:
+
+```xml
+<SomeControl AbsoluteUri="{RelativeUriResolver 'image.jpg'}"
+```
+
+```cs
+public class RelativeUriResolver : MarkupExtension
```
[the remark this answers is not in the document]
MikeHillberg: We'll consider and open an Issue.

active/XamlServiceProvider/XamlServiceProviderApi.md
MikeHillberg: As for where the interface belongs, we'll follow up with the above-mentioned cs/winrt issue

active/XamlServiceProvider/XamlServiceProviderApi.md
oldnewthing: `Direction="{local:MyBind PropertyName=DirectionToUse, Source=Somebody}"` will set the Direction property to the current value of `Somebody.DirectionToUser`. And if you don't specify a Source, then it defaults to the root page. So yeah, this is basically x:Bind reimplemented, but at least it's something less artificial than "Give me the title of the root element."

MikeHillberg: Cool, basically a dynamic x:Bind

active/XamlServiceProvider/XamlServiceProviderApi.md
MikeHillberg: Oh, as in what targets other than ProvideValue. MarkupExtension exists for the ProvideValue call, with the output being set on a target. So it's really just about this one target.

MikeHillberg:
> I'm wondering why you choose to name this `IXamlServiceProvider` instead of already existing `IServiceProvider`

IServiceProvider is specific to .NET. But we should project it to IServiceProvider in the cs/winrt projection:
https://github.com/microsoft/CsWinRT/issues/358

oldnewthing:
> > I'm wondering why you choose to name this `IXamlServiceProvider` instead of already existing `IServiceProvider`
> 
> IServiceProvider is specific to .NET. But we should project it to IServiceProvider in the cs/winrt projection:
> [microsoft/CsWinRT#358](https://github.com/microsoft/CsWinRT/issues/358)

There is already [a classic COM interface called IServiceProvider](https://docs.microsoft.com/en-us/previous-versions/windows/internet-explorer/ie-developer/platform-apis/cc678965(v=vs.85)), so C++ code may encounter name conflicts if we called it IServiceProvider.

active/XamlServiceProvider/XamlServiceProviderApi.md
MikeHillberg: Current plan is to leave this as is in this spec (in the Xaml.Data namespace).

PR #90, opened 2020-07-07
Updates to CornerRadiusToThicknessConverter for more filters

MikeHillberg:


active/CornerRadius/CornerRadiusToThickness2.md
```diff
@@ -0,0 +1,108 @@
...
+* FilterTopFromTopLeft,
+* FilterTopFromTopRight,
+* FilterRightFromTopRight,
+* FilterRightFromBottomRight,
+* FilterBottomFromBottomRight,
+* FilterBottomFromBottomLeft,
+* FilterLeftFromBottomLeft,
+* FilterLeftFromTopLeft,
+
+The CornerRadiusFilterConverter is also updated with a new property -- Multiplier -- which is applied during the converstion. For example, setting this to -1 negates the value.
```
chingucoding, line 46: 1. `conversion` instead of `converstion`
2. The new multiplier property should also be present in the API details down
3. Maybe Multiplier could be of type thickness, allowing scaling different parts by different factors. Maybe we could also allow "FilterAllToAll" (semantically speaking) to make the converter even more powerful.

MikeHillberg: What would FilterAllToAll do? Both structs have 4 fields, but really we're mapping 8 values onto 4.

chingucoding: FilterAllToAll would do something like this:

image

We could also do the max/min/average of the corner radii on a side to get the thickness for that side, e.g. `Left = min(BottomLeft,TopLeft)`.

MikeHillberg: Those work but I can't see the use case for them, for example when a UI would want to align the top margin of one element with the top/left corner radius of another element. I think we should keep what you have and add more when we need it.

chingucoding: Fair point, it would definitely not be a common case. Let's not add it then and stick to what is in this proposal already.

active/CornerRadius/CornerRadiusToThickness2.md
```diff
@@ -0,0 +1,151 @@
+> See comments in Markdown for how to use this spec template
+
```
chingucoding, line 2: nit: Remove this

active/CornerRadius/CornerRadiusToThickness2.md
```diff
@@ -0,0 +1,149 @@
...
+* FilterTopFromTopLeft,
+* FilterTopFromTopRight,
+* FilterRightFromTopRight,
+* FilterRightFromBottomRight,
+* FilterBottomFromBottomRight,
+* FilterBottomFromBottomLeft,
+* FilterLeftFromBottomLeft,
+* FilterLeftFromTopLeft,
+
+The CornerRadiusFilterConverter is also updated with a new property -- Multiplier -- which is applied during the conversion. For example, setting this to -1 negates the value.
```
oldnewthing, line 44: Would `Scale` be a more natural name than `Multiplier`?

active/CornerRadius/CornerRadiusToThickness2.md
```diff
@@ -0,0 +1,149 @@
...
+
+**New values:**
+* FilterTopFromTopLeft,
+* FilterTopFromTopRight,
+* FilterRightFromTopRight,
+* FilterRightFromBottomRight,
+* FilterBottomFromBottomRight,
+* FilterBottomFromBottomLeft,
+* FilterLeftFromBottomLeft,
+* FilterLeftFromTopLeft,
```
oldnewthing, line 42: Looks like you can't mix them, so if you want to filter top and left, you're out of luck. (I'm okay with that, just checking.)

active/CornerRadius/CornerRadiusToThickness2.md
```diff
@@ -0,0 +1,149 @@
+
+# Background
+
+_This spec updates the existing [CornerRadiusToThicknessConverter](https://docs.microsoft.com/uwp/api/Microsoft.UI.Xaml.Controls.Primitives.CornerRadiusToThicknessConverter) class with new filters._
```
oldnewthing, line 4: Curiously, the existing documentation lacks these illustrations and examples.

active/CornerRadius/CornerRadiusToThickness2.md
```diff
@@ -0,0 +1,149 @@
...
+
+```xml
+<Path 
+    Margin="{Binding Source={ThemeResource OverlayCornerRadius},
+        Converter={StaticResource CornerRadiusToThickness1}}"
+    Height="100"
+    Margin="100"
+    Fill="Red"
+    Data="M4 0 L4 4 L0 4 A4,4 90 0 0 4 0 Z" /> 
+```
```
oldnewthing, line 68: Would like to see the results of this example. I tried pasting this into a test app and saw nothing because there is no stroke thickness or brush, and because I'm not sure what "M4 0 ..." produces. A labeled illustration would help.

active/CornerRadius/CornerRadiusToThickness2.md
```diff
@@ -0,0 +1,149 @@
...
+
+This converter can then be used elsewhere in Xaml:
+
+```xml
+<Path 
+    Margin="{Binding Source={ThemeResource OverlayCornerRadius},
+        Converter={StaticResource CornerRadiusToThickness1}}"
+    Height="100"
+    Margin="100"
```
oldnewthing, line 65: TYPO: `Margin` being set twice.

active/CornerRadius/CornerRadiusToThickness2.md
oldnewthing, line 44: I'm having trouble understanding the scenario for a negative multiplier. I tried

```xaml
                    <StackPanel Orientation="Horizontal" Padding="50,0,0,0">
                        <Rectangle Height="50" Width="50" Fill="White"/>
                        <Border BorderBrush="White" BorderThickness="2" CornerRadius="30" Margin="-30">
                            <Rectangle Height="200" Width="200" Fill="Red" Opacity="0.5"/>
                        </Border>
                        <Rectangle Height="50" Width="50" Fill="White"/>
                    </StackPanel>
```

in a test app and the result didn't make sense.

image

active/CornerRadius/CornerRadiusToThickness2.md
```diff
@@ -0,0 +1,149 @@
...
+```
+
+
+# API Notes
+
+## CornerRadiusToThicknessConverter.Multiplier property
+
+Type: [Thickness](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.Thickness)
+
+Multiplied to the value in the target Thickness. Defaults to (1,1,1,1)
```
oldnewthing, line 77: TYPO: "Multiplied **by**"

active/CornerRadius/CornerRadiusToThickness2.md
```diff
@@ -0,0 +1,149 @@
...
+
+This multiplier is a Thickness with left/top/right/bottom values, which are applied to the matching fields in the target. Note that in Xaml markup you can specify a single number and it's set to all 4 fields. For example the following sets the Multipler to (-1, -1, -1, -1):
+
+```xml
+Multipler = '-1'
```
oldnewthing, line 82: TYPO: `Multiplier`.

Also, I believe XML style is to remove spaces around the `=` sign.

active/CornerRadius/CornerRadiusToThickness2.md
chingucoding: Yes, you can't mix them. Maybe we should do this, but how should we do that? If we add more enum values for those cases, we would end up with quite a large enum.

active/CornerRadius/CornerRadiusToThickness2.md
chingucoding: For once, a (hacky) use case, where this spec is also coming from is the inset corner radii of the WinUI TabView control: https://github.com/microsoft/microsoft-ui-xaml/pull/2201

What negative margin does (from my understanding) is the following: If you have content, the width and height is calculated based on the available size, and the margin. If a control has 200px * 200px, and 50px margin, then it will render with an actual size of 100px * 100px. So with negative margin, the space used for rendering is 200px - (-50px + -50px) = 300px, making it render with a width and height of 300px. Since the control has a fixed position and "available size", it starts to render out of those bounds, extending into other content.

active/CornerRadius/CornerRadiusToThickness2.md
chingucoding: Good point, the example should show something when simply copy pasted. Fun fact, the path renders this kind of shape: 

image

active/CornerRadius/CornerRadiusToThickness2.md
chingucoding: For strictly positive values, I would say yes, but I am not sure if you refer to a scale of "-1".

oldnewthing: Negative scale factors are a thing.

chingucoding: That's true, though I mostly have encountered them in a very mathematical scenario. Given that negative values are not that common as use case, `Scale` might be the better wording.

active/CornerRadius/CornerRadiusToThickness2.md
MikeHillberg: Recommend: Carry over to the doc update

active/CornerRadius/CornerRadiusToThickness2.md
MikeHillberg: Discussion: It's not a [flags] enum, so it would be a breaking change to make it one now.

active/CornerRadius/CornerRadiusToThickness2.md
MikeHillberg: Recommend: add some high level explanation about how the paths are using negative margins to extend out to the left and right

active/CornerRadius/CornerRadiusToThickness2.md
MikeHillberg: Recommend: everything should be copy/paste-able.

active/CornerRadius/CornerRadiusToThickness2.md
MikeHillberg: Discussion: We have two "multipliers" today (e.g. [PointerDeviceUsage.PhysicalMultiplier](https://docs.microsoft.com/uwp/api/Windows.Devices.Input.PointerDeviceUsage.PhysicalMultiplier)). Note that both are floats rather than this (Thickness). Scales are mostly scalars, vector2, or vector3s.

"Scale" might be confusing when negative, because a scale transform does a flip, which this does not.

Recommend: let's stick with multiplier

PR #91, opened 2020-07-09
Updates to GridSyntaxSpec

MikeHillberg:
This file is an [updated copy](./GridSyntaxSpec.md), split off to make all of the lines commentable in PR. But this will be merged back when complete.

### DIFF
--- a/active/gridsyntax/GridSyntaxSpec2.md
+++ b/active/gridsyntax/GridSyntaxSpec2.md
@@ -1,0 +1,368 @@
+
+
+# Background
+_This adds a new short-hand Xaml syntax feature that makes it much easier to define a Grid in Xaml makrup. The proposal for this feature can be found [here.](https://github.com/microsoft/microsoft-ui-xaml/issues/673) The existing API documentation for Grid can be found [here.](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.controls.grid)_
+
+_This file is an [updated copy](./GridSyntaxSpec.md), split off to make all of the lines commentable in PR. But this will merged back when complete._
+
+The Xaml Grid  is one of the most widely used panels, yet it has a noticeably long and repetitive syntax in markup, which leads to a tough learning curve for new developers. Currently, to create a Grid, developers must define each row height and column width with separate XML tags:
+
+```xml
+<Grid>
+    <Grid.ColumnDefinitions>
+          <ColumnDefinition Width="1*" />
+          <ColumnDefinition Width="2*" />
+          <ColumnDefinition Width="Auto" />
+          <ColumnDefinition Width="*" />
+          <ColumnDefinition Width="300" />
+    </Grid.ColumnDefinitions>
+    <Grid.RowDefinitions>
+          <RowDefinition Height="1*" />
+          <RowDefinition Height="Auto" />
+          <RowDefinition Height="25" />
+          <RowDefinition Height= "14" />
+          <RowDefinition Height="20" />
+    </Grid.RowDefinitions>
+</Grid>
+```
+
+The code in that example creates the ColumnDefinition objects and RowDefinition objects and adds them to the Grid's [Grid.ColumnDefinitions](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.Controls.Grid.ColumnDefinitions) and [Grid.RowDefinitions](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.Controls.Grid.RowDefinitions) collections, respectively.  This is currently the only way to set these properties in markup.
+
+The new features in this spec allow for this equivalent syntax:
+
+```xml
+<Grid ColumnDefinitions="1*, 2*, Auto, *, 300" RowDefinitions="1*, Auto, 25, 14, 20" />
+```
+
+This new syntax is enabled by a new Xaml language feature and associated APIs that can be used on any collection-typed property, and includes updates to Grid to use these APIs.
+
+# Description
+
+We've introduced a new XAML language feature that allows the initialization of collection-type properties (including read-only properties) using Xaml attribute syntax. We've also updated the Grid APIs to make sure that this syntax works properly in particular for Grid's use cases.
+
+Grid initialization in Xaml markup has always required a lot of XML:
+
+```xml
+<Grid>
+    <Grid.ColumnDefinitions>
+          <ColumnDefinition Width="Auto" />
+          <ColumnDefinition Width="*" />
+    </Grid.ColumnDefinitions>
+</Grid>
+```
+
+You can now accomplish the same thing in a single tag:
+
+```xml
+<Grid ColumnDefinitions="Auto, *">
+```
+
+Before this change, if you write that shorter Xaml markup, the  Xaml loader would try to convert the string `Auto, *` into a [ColumnDefinitionCollection](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.Controls.ColumnDefinitionCollection) object, as that's the type of the [Grid.ColumnDefinitions](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.Controls.Grid.ColumnDefinitions) property. That fails because there is no such conversion available. (And it wouldn't help if there was, because the property is read-only.)
+
+But the Xaml syntax rules now support this scenario by recognizing that the target property is a collection type, and that there's no conversion from string available, and then by treating the attribute value as a comma-separated list of collection items. The items of the collection are of type [ColumnDefinition](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.Controls.ColumnDefinition), so a check is made to see if strings can be converted to ColumnDefinition instances. Since they can, the Xaml loader creates those instances and adds them to the collection.
+
+| Do we have documentation that explains what exactly a "collection" is? Can't remember if it's just IVector, or if it has more rules like the C# initialization syntax.
+
+## Role of the [ContentProperty] attribute
+
+In order for a ColumnDefinition (or RowDefinition) to be created from a string there is a second new Xaml syntax rule: if a type can't be created from string, but the type's [[ContentProperty](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.Markup.ContentPropertyAttribute)] property can, the type is created and the property set.
+
+About this "content property", it's not literally a property named "content" (although it can be and [is sometimes](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.Controls.ContentControl.Content)). This is essentially the default property of a class, and is marked with the [ContentProperty] attribute.
+
+For example the [SolidColorBrush](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.Media.SolidColorBrush) is defined roughly like this (in C# terms):
+
+```cs
+[ContentProperty(Name = "Color")]
+public sealed class SolidColorBrush : Brush
+{
+    ...
+    public Color Color { get; set; }
+}
+```
+
+Because [SolidColorBrush.Color](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.Media.SolidColorBrush.Color) is marked as the [ContentProperty] (and because a Color can be created from a string), you've always been able to write this short-hand Xaml:
+
+```xml
+<SolidColorBrush>Red</SolidColorBrush>
+```
+
+If it weren't for the [ContentProperty] on SolidColorBrush, you'd have to use the equivalent full syntax (and that's still valid to use if you prefer):
+
+```xml
+<SolidColorBrush>
+    <SolidColorBrush.Color>
+        Red
+    </SolidColorBrush.Color>
+</SolidColorBrush>
+```
+
+## Making things work for Grid
+
+Back to Grid, with the new syntax feature and with ColumnDefinition.Width now marked as the [ContentProperty]:
+
+```cs
+[ContentProperty(Name = "Width")]
+public sealed class ColumnDefinition : DependencyObject
+{
+        public GridLength Width { get; set; }
+}
+```
+
+you can now write
+
+```xml
+<ColumnDefinition>Auto</ColumnDefinition>
+```
+
+(There's already a converter to create a GridLength from a string.)
+
+Putting that all together, back to the original goal:
+
+<Grid ColumnDefinitions="Auto, *">
+
+Two ColumnDefinition objects are created and added to the ColumnDefinitions collection property, which is ready-only but defaults to an empty collection. The first item has its Width set to GridLength.Auto, the second has it's Width set to 1 GridUnitType.Star.
+
+
+## Special characters and escaping
+
+The XML attribute value is interpreted as a comma-separated list of items, which creates a problem if you want to use a comma an item value. You can solve this using quotes, for example, this initializes a collection with two items:
+
+```xml
+<WordWheel Words = " 'Hello, world', 'how is the weather?' " />
+```
+
+(Xaml markup extensions have the same command issue and the same quote syntax. In general the syntax rules follow those of markup extensions.)
+
+Alternatively you can use a backslash as an escape character before the comma:
+
+```xml
+Words = " I weigh 90\,7 KG    "
+```
+| todo: handle apostrophe in "How's the weather?" case
+
+Note that here, like everywhere in XML, we're using "opposite quotes" (whichever kind of quote you use to delineate the attribute value, don't use it _in_ the attribute value, except as an entity). So these are equvalent:
+
+```xml
+Words = " 'Hello', 'world' "
+Words = ' "Hello", "world" '
+```
+
+## Don't mix syntaxes
+
+In Xaml today you can't set a property twice. For example this isn't valid because Background is getting set twice:
+
+```xml
+<Button Content='Click' Background='Red'>
+    <Button.Background>
+        <SolidColorBrush Color='Blue' />
+    </Button.Background>
+</Button>
+```
+
+That rule still applies, so the following is an error:
+
+```xml
+//Warning: Incorrect syntax
+<Grid ColumnDefinitions="Auto,*,Auto">
+    <Grid.ColumnDefinitions>
+          <ColumnDefinition Width="*" />
+          <ColumnDefinition Width="Auto" />
+          <ColumnDefinition Width="*" />
+    </Grid.ColumnDefinitions>
+</Grid>
+```
+
+## Things that don't change
+
+There are some interesting details with collection properties that don't change with the new syntax:
+* Items are simply _added_ to the collection, so if the collection defaults to non-empty, it will have more items in it than what you see in the markup.
+* The collection property must be non-null. If it's null, then the value being assigned to it must be the type of the collection, and so this new syntax isn't applicable.
+
+
+## Other Use Cases
+
+This more succinct syntax for initializing collections was created with the goal of making Grid easier and more intuitive. However, this new syntax will be implemented as a language feature and work properly for any collection-type property. The current syntaxes will still be functional for all other scenarios, as it will be with Grid. Some more examples:
+
+### ColorWheel
+
+The code below shows an example of how the syntax could be used for a property that's (in C# terms) an IList<Color>. 
+
+```xml
+<ColorWheel Colors="Red, Green, Blue"/>
+```
+
+### CalendarView
+
+Initialize the selected dates on a calendar:
+
+```xml 
+<CalendarView SelectedDates="'1/5/1977', '6/19/1997'" />
+```
+
+### ListView
+
+Previous syntax to populate ListView Items collection (which still works):
+
+```xml
+<ListView>
+    <ListView.Items>
+        <x:string>Hello</x:string>
+        <x:string>World</x:string>
+    </ListView.Items>
+</ListView>
+```
+
+New syntax:
+
+```xml 
+<ListView Items="'Hello', 'world'" />
+```
+
+## MS-XAML
+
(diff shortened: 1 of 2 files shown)


active/gridsyntax/GridSyntaxSpec2.md
```diff
@@ -0,0 +1,368 @@
+
+
+# Background
+_This adds a new short-hand Xaml syntax feature that makes it much easier to define a Grid in Xaml makrup. The proposal for this feature can be found [here.](https://github.com/microsoft/microsoft-ui-xaml/issues/673) The existing API documentation for Grid can be found [here.](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.controls.grid)_
```
marb2000, line 4: It recommend to use the trademark (official name) XAML - in upper case, instead of Xaml. Xaml is correct when you refers to a namespace. This is the same case of .NET, it's .NET neither .Net nor .net.

active/gridsyntax/GridSyntaxSpec2.md
```diff
@@ -0,0 +1,368 @@
...
+
+But the Xaml syntax rules now support this scenario by recognizing that the target property is a collection type, and that there's no conversion from string available, and then by treating the attribute value as a comma-separated list of collection items. The items of the collection are of type [ColumnDefinition](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.Controls.ColumnDefinition), so a check is made to see if strings can be converted to ColumnDefinition instances. Since they can, the Xaml loader creates those instances and adds them to the collection.
+
+| Do we have documentation that explains what exactly a "collection" is? Can't remember if it's just IVector, or if it has more rules like the C# initialization syntax.
```
marb2000, line 64: IVector is the WinRT type and IList is the .NET projection: 
- https://docs.microsoft.com/en-us/uwp/api/windows.foundation.collections.ivector-1?view=winrt-19041 
https://docs.microsoft.com/en-us/windows/uwp/cpp-and-winrt-apis/collections

anawishnoff: This link also gives an overview of collections in .NET: https://docs.microsoft.com/en-us/uwp/api/windows.foundation.collections?view=winrt-19041

active/gridsyntax/GridSyntaxSpec2.md
```diff
@@ -0,0 +1,368 @@
...
+<ColumnDefinition>Auto</ColumnDefinition>
+```
+
+(There's already a converter to create a GridLength from a string.)
+
+Putting that all together, back to the original goal:
+
+<Grid ColumnDefinitions="Auto, *">
+
+Two ColumnDefinition objects are created and added to the ColumnDefinitions collection property, which is ready-only but defaults to an empty collection. The first item has its Width set to GridLength.Auto, the second has it's Width set to 1 GridUnitType.Star.
```
anawishnoff, line 123:
```suggestion
Two ColumnDefinition objects are created and added to the Grid's ColumnDefinitions collection property, which is read-only but defaults to an empty collection. The first item has its Width set to GridLength.Auto, the second has its Width set to 1 GridUnitType.Star.
```

active/gridsyntax/GridSyntaxSpec2.md
```diff
@@ -0,0 +1,368 @@
...
+
+## Special characters and escaping
+
+The XML attribute value is interpreted as a comma-separated list of items, which creates a problem if you want to use a comma an item value. You can solve this using quotes, for example, this initializes a collection with two items:
+
+```xml
+<WordWheel Words = " 'Hello, world', 'how is the weather?' " />
+```
+
+(Xaml markup extensions have the same command issue and the same quote syntax. In general the syntax rules follow those of markup extensions.)
```
anawishnoff, line 134: command issue?

active/gridsyntax/GridSyntaxSpec2.md
```diff
@@ -0,0 +1,368 @@
...
+
+Alternatively you can use a backslash as an escape character before the comma:
+
+```xml
+Words = " I weigh 90\,7 KG    "
+```
+| todo: handle apostrophe in "How's the weather?" case
```
anawishnoff, line 141: I remember that we can't use &apos; in XAML. Just to refresh my mind, I'm assuming we can't just escape the apostrophe with \' either? 

In that case, do we just suggest using single quotes on the outside and double quotes for each of your values?

Words = ' "How's the weather", "It's sunny" '

active/gridsyntax/GridSyntaxSpec2.md
```diff
@@ -0,0 +1,368 @@
...
+```xml <TextBlock Text="{x:Bind VM.MyText, FallbackValue='Hello, world', Mode=TwoWay}" />```
+
+This uses a comma as a delimiter, but also allows strings to have commas within them. For these reasons, we chose to use the syntax in which a collection can be delimited by commas but have more complex values be encased within single quotes.
+
+
+## Release Details
+The xmlns will not be updated for this XAML language feature addition, i.e. there will be no formal update to the XAML language. Visual Studio, however, will have to be updated to accept the new syntax structure. This feature will be released with WinUI and not with the XAML SDK.
+
+## Tooling/Visual Studio Considerations
+The Designer is able to tell when an element is createable from string, and this language feature will make it so all collection-type properties are createable from string (if that string is formatted as a comma-separated list). So, the Designer will recognize that a collection-type property is assigned to a string, check if it is assignable from string, and then attempt to treat the string as a comma-separated list, similar to the way the XAML Parser works.
```
anawishnoff, line 339: I believe this isn't completely true anymore. If a collection-typed property defaults to null, it isn't able to be creatable from string via this new syntax.

active/gridsyntax/GridSyntaxSpec2.md
```diff
@@ -0,0 +1,368 @@
...
+    * KPI: Multiple new commenters on this thread have gone to be active on other threads, and all experienced commenters from this thread have gone on to create at least one other thread or issue on the WinUI repo. 
+    * Measurement: Number of *new* commenters on this thread (i.e. made their first WinUI Github comment/contribution on this thread), and how many of them went on to be active on other WinUI or Microsoft threads.
+    * Measurement: Number of the commenters from this thread went on to create other issues after this one was resolved (beleiving that their voices would be heard!).
+
+### P1: Feature Performance Indictators
+* Syntax has decreased Grid's learning curve
+    * KPI: A slight increase (<5%) of apps released to the Windows store were created by new developers
+    * Measurement: How many apps using this new syntax were created by new/first-time developers
+
+# Open Questions
```
anawishnoff, line 362: I think these open questions are super old and can be deleted :)

active/gridsyntax/GridSyntaxSpec2.md
```diff
@@ -0,0 +1,368 @@
+
+
+# Background
+_This adds a new short-hand Xaml syntax feature that makes it much easier to define a Grid in Xaml makrup. The proposal for this feature can be found [here.](https://github.com/microsoft/microsoft-ui-xaml/issues/673) The existing API documentation for Grid can be found [here.](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.controls.grid)_
+
+_This file is an [updated copy](./GridSyntaxSpec.md), split off to make all of the lines commentable in PR. But this will merged back when complete._
```
asklar, line 6:
```suggestion
_This file is an [updated copy](./GridSyntaxSpec.md), split off to make all of the lines commentable in PR. But this will be merged back when complete._
```

active/gridsyntax/GridSyntaxSpec2.md
```diff
@@ -0,0 +1,368 @@
...
+If it weren't for the [ContentProperty] on SolidColorBrush, you'd have to use the equivalent full syntax (and that's still valid to use if you prefer):
+
+```xml
+<SolidColorBrush>
+    <SolidColorBrush.Color>
+        Red
+    </SolidColorBrush.Color>
+</SolidColorBrush>
+```
+
```
asklar, line 98: should this be in its own spec or doc since it's used by the grid syntax but not specific to it?

active/gridsyntax/GridSyntaxSpec2.md
```diff
@@ -0,0 +1,368 @@
...
+```xml
+<ColumnDefinition>Auto</ColumnDefinition>
+```
+
+(There's already a converter to create a GridLength from a string.)
+
+Putting that all together, back to the original goal:
+
+<Grid ColumnDefinitions="Auto, *">
```
asklar, line 121: make markdown?

active/gridsyntax/GridSyntaxSpec2.md
```diff
@@ -0,0 +1,368 @@
...
+
+## Role of the [ContentProperty] attribute
+
+In order for a ColumnDefinition (or RowDefinition) to be created from a string there is a second new Xaml syntax rule: if a type can't be created from string, but the type's [[ContentProperty](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.Markup.ContentPropertyAttribute)] property can, the type is created and the property set.
+
+About this "content property", it's not literally a property named "content" (although it can be and [is sometimes](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.Controls.ContentControl.Content)). This is essentially the default property of a class, and is marked with the [ContentProperty] attribute.
+
+For example the [SolidColorBrush](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.Media.SolidColorBrush) is defined roughly like this (in C# terms):
+
+```cs
```
asklar, line 74: here and in 103 you use cs but in others you used csharp, make them match?

active/gridsyntax/GridSyntaxSpec2.md
MikeHillberg: C# has a definition of "collection" that's not defined by an interface (in the initialization syntax). For example, among other things, it looks for an "Add" method with a particular signature. IList has a matching Add method, so you can implement that and it works. (Similarly C# 'foreach' looks for a method name/signature, not IEnumerable.) 

I think the answer is that you're right, the WinUI Xaml compiler/loader doesn't have a definition like that, and relies specifically on IVector/IList.

active/gridsyntax/GridSyntaxSpec2.md
MikeHillberg: There's a bug in markup extension syntax today that prevents the quote from being escaped. But good point, there's no reason to reproduce that bug in this syntax.

active/gridsyntax/GridSyntaxSpec2.md
MikeHillberg: This whole Discussion section is written like a blog post. This section on the CPA is a backgrounder on how it works, in order to understand what follows in the spec.

active/gridsyntax/GridSyntaxSpec2.md
MikeHillberg: Oops, it still colors correctly in VSCode :)

active/gridsyntax/GridSyntaxSpec2.md
MikeHillberg: I don't think anything in here refers to the XAML product, it's all the Xaml language, which I believe follow the API casing guidelines (PascalCasing).

active/gridsyntax/GridSyntaxSpec2.md
MikeHillberg: https://github.com/microsoft/microsoft-ui-xaml/issues/2904

active/gridsyntax/GridSyntaxSpec2.md
```diff
@@ -0,0 +1,368 @@
...
+## Syntax Considerations and Rationale
+In developing the syntax, the team considered developer familiarty and patterns as much as possible to ensure that the Grid learning curve could be decreased. One of the considerations was how similar the syntax should be to JSON, as that's a popular notation style that often addresses the same problems that Xaml does. 
+
+JSON defines their arrays as: ```"words" : ["apple", "banana", "cherry"]```
+
+Xaml defines a collection-typed property as: ```<Words="'Apple', 'Banana', 'Cherry'"/>```
+
+The syntax that we chose encloses strings individually in quotes, as JSON does, but it does so in a way that is more consistent with existing Xaml style and Xaml functions. An example of this syntax used elsewhere in Xaml is for the x:Bind syntax:
+
+```xml <TextBlock Text="{x:Bind VM.MyText, FallbackValue='Hello, world', Mode=TwoWay}" />```
```
oldnewthing, line 341: Need to split onto three lines. Cramming into one line doesn't work because the `xml` is being interpreted as literal text rather than a syntax highlighter hint. It also confuses github's markdown syntax-highlighter.

~~~
```xml
<TextBlock Text="..." />
``` 
~~~

active/gridsyntax/GridSyntaxSpec2.md
```diff
@@ -0,0 +1,368 @@
...
+
+This uses a comma as a delimiter, but also allows strings to have commas within them. For these reasons, we chose to use the syntax in which a collection can be delimited by commas but have more complex values be encased within single quotes.
+
+
+## Release Details
+The xmlns will not be updated for this Xaml language feature addition, i.e. there will be no formal update to the Xaml language. Visual Studio, however, will have to be updated to accept the new syntax structure. This feature will be released with WinUI and not with the Xaml SDK.
```
oldnewthing, line 347: Is it legal to change the language without revving the language? This will lead to confusion if somebody takes some new-style markup and brings it to an old xaml compiler. The compiler will be totally happy with the version number but balk at the new synax.

active/gridsyntax/GridSyntaxSpec2.md
```diff
@@ -0,0 +1,368 @@
...
+
+## Tooling/Visual Studio Considerations
+In terms of usage of this new Xaml language feature in Visual Studio, VS needs a way to recognize the current Xaml version that the developer is using. This will be likely be done by accessing metadata or accessing an attached features.xml file - the VS team is still in discussion as to which method will be used, but the overall cost is not high. 
+
+## Data and Intelligence Metrics
+### P0: Feature Key Performance Indictors
+* Syntax has increased developer/consumer satisfaction
+    * KPI: At least 80% of new apps released to Windows store after release of feature are using the new syntax, and 50% of developers express a positive reaction to this change.
```
oldnewthing, line 355: I think 80% is too aggressive. There's a lot of pre-existing XAML out there that doesn't use the new syntax and nobody is going to go through their existing XAML and refresh it before pushing a bugfix to Store.

active/gridsyntax/GridSyntaxSpec2.md
```diff
@@ -0,0 +1,368 @@
...
+
+But the Xaml syntax rules now support this scenario by recognizing that the target property is a collection type, and that there's no conversion from string available, and then by treating the attribute value as a comma-separated list of collection items. The items of the collection are of type [ColumnDefinition](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.Controls.ColumnDefinition), so a check is made to see if strings can be converted to ColumnDefinition instances. Since they can, the Xaml loader creates those instances and adds them to the collection.
+
+## Role of the [ContentProperty] attribute
+
+In order for a ColumnDefinition (or RowDefinition) to be created from a string there is a second new Xaml syntax rule: if a type can't be created from string, but the type's [[ContentProperty](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.Markup.ContentPropertyAttribute)] property can, the type is created and the property set.
```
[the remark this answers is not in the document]
MikeHillberg: Recommend: Articulate this as three rules.

active/gridsyntax/GridSyntaxSpec2.md
```diff
@@ -0,0 +1,368 @@
...
+
+## Special characters and escaping
+
+The XML attribute value is interpreted as a comma-separated list of items, which creates a problem if you want to use a comma an item value. You can solve this using quotes, for example, this initializes a collection with two items:
+
+```xml
+<WordWheel Words = " 'Hello, world', 'how is the weather?' " />
+```
+
+(Xaml markup extensions have the same comma issue and the same quote syntax. In general the syntax rules follow those of markup extensions.)
```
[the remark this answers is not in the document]
MikeHillberg: Discussion: If we fix 2904 we need to ensure we don't break the workaround for the hat rule

This is the workaround (for C#), which is adding a whack to preserve the hat:

```xml
<TextBlock Text="{x:Bind Foo(&quot;a\\^&quot;b&quot;)}"/>
```

active/gridsyntax/GridSyntaxSpec2.md
MikeHillberg: Discussion: In other words, do we need to update xmlns:x='http://schemas.microsoft.com/winfx/2006/xaml'?

For example, the namespace was updated for 2009.

We didn't update the language version number for other additive updates:
* x:Bind
* Conditional Xaml

active/gridsyntax/GridSyntaxSpec2.md
MikeHillberg: Recommend: We can leave these D&I goals out

active/gridsyntax/GridSyntaxSpec2.md
```diff
@@ -0,0 +1,370 @@
...
+    </Grid.RowDefinitions>
+</Grid>
+```
+
+The code in that example creates the ColumnDefinition objects and RowDefinition objects and adds them to the Grid's [Grid.ColumnDefinitions](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.Controls.Grid.ColumnDefinitions) and [Grid.RowDefinitions](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.Controls.Grid.RowDefinitions) collections, respectively.  This is currently the only way to set these properties in markup.
+
+The new features in this spec allow for this equivalent syntax:
+
+```xml
+<Grid ColumnDefinitions="1*, 2*, Auto, *, 300" RowDefinitions="1*, Auto, 25, 14, 20" />
```
MikeHillberg, line 34: Discussion: DoubleCollection has a type converter that support spaces, but also supports commas (Shape.StrokeDashArray). So to be consistent with that we could support both, but prefer to simplify and just support commas going forward (and this is still consistent or not inconsistent).

active/gridsyntax/GridSyntaxSpec2.md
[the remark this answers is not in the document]
anawishnoff: Just pushed a fix, but wasn't sure what to put for the third rule. I clarified the first two rules.

active/gridsyntax/GridSyntaxSpec2.md
anawishnoff: I agree that it's best to only support commas. The spec only mentions comma-separated lists so I don't think there have to be any changes made here.

active/gridsyntax/GridSyntaxSpec2.md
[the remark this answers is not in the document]
oldnewthing: Rule 3 is the rule being introduced in this spec.

* If rules 1 and 2 do not apply, but the type is a collection, and the collection item type has a [ContentProperty] attribute, then split the string on commas (allowing quotation marks to protect commas), and construct a new item for each substring, adding it to the collection.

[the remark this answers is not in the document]
MikeHillberg: I think the text was confusing. It was trying to say that there are two _new_ syntax rules: the comma separated list, and creating at type using it's [ContentProperty].

As for creating a type using the [ContentProperty], what's confusing is that it already partially works today. If a type can't be created from string, but it's [ContentProperty] is of a type that can be _cast_ to from string, then do that. That's why `<TextBlock>Hello</TextBlock>` works (TextBlock.Text is a string type). Basically the property has to be of type string or object.

The second new rule is that if the [ContentProperty] type can be _created_ from string, we'll use that too. So if the property's type is string or object, that's still good, but now if the property's type has a type converter we'll use that too.

I'm updating the text of the spec.

active/gridsyntax/GridSyntaxSpec2.md
```diff
@@ -0,0 +1,364 @@
...
+
+(Xaml markup extensions have the same comma issue and the same quote syntax. In general the syntax rules follow those of markup extensions.)
+
+Alternatively you can use a backslash as an escape character before the comma:
+
+```xml
+Words = " I weigh 90\,7 KG    "
+```
+
+Note that here, like everywhere in XML, we're using "opposite quotes" (whichever kind of quote you use to delineate the attribute value, don't use it _in_ the attribute value, except as an entity). So these are equvalent:
```
oldnewthing, line 152: TYPO: "equvalent"

active/gridsyntax/GridSyntaxSpec2.md
```diff
@@ -0,0 +1,364 @@
...
+<Grid ColumnDefinitions="Auto, *">
+```
+
+## Collection property values as comma-separated lists
+
+Before these changes, if you write that shorter Xaml markup, the  Xaml loader would try to convert the string `Auto, *` into a [ColumnDefinitionCollection](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.Controls.ColumnDefinitionCollection) object, as that's the type of the [Grid.ColumnDefinitions](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.Controls.Grid.ColumnDefinitions) property. That fails because there is no such conversion available. (And it wouldn't help if there was, because the property is read-only.)
+
+A new Xaml syntax rule now takes a next step though and recognizes that the target property is a collection type, and then treats the attribute value as a comma-separated list of collection items. The items of the collection are of type [ColumnDefinition](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.Controls.ColumnDefinition), so a check is made to see if strings can be converted to ColumnDefinition instances. 
+
+But it can't.
```
oldnewthing, line 66: I like the new exposition, especially clarifying that this proposal encompasses *two* new rules.

PR #94, opened 2020-07-31
InfoBar Control

gabbybilka:
For [issue 913](https://github.com/microsoft/microsoft-ui-xaml/issues/913): 
Expected changes to be added before this will be merged:

-  [x] Consistent mockups for the examples provided (and removal of sketches/placeholder images)
-  [x] Improved example code snippets
-  [x] Clarification on Assistive Technology behavior especially regarding it changing based on the Severity level set
-  [x] Behavior on updated properties, especially in regards to events


Felix-Dev:
As proposed in https://github.com/microsoft/microsoft-ui-xaml/issues/2916 I would like to see any theme resources exposed for this control to be part of the API section in this spec as well.

gabbybilka:
> As proposed in [microsoft/microsoft-ui-xaml#2916](https://github.com/microsoft/microsoft-ui-xaml/issues/2916) I would like to see any (theme) resources exposed for this control to be part of the API section in this spec as well.

Hi @Felix-Dev thanks for pointing those issues out to me! I've added in our preliminary theme resources that are unique to InfoBar but there are more to come.

gabbybilka:
Open questions related to the Background/Guidance/Recommendations section of the spec:
- Thoughts on the name? InfoBar? Is this flexible enough for a future floating/overlay/toast mode?
- Referred to as a notification, status message, and information bar synonymously throughout the document. Is there another term that sounds right to you?
- Are the terms directly and indirectly impactful to the UX make sense? Or does error and informational state work better? This is the comparison of terminology in the description vs "when to show an information bar" section.
- Is there anything else that should be added to Remarks? It currently only has the "when to show an information bar" section.

active/InformationBar/InformationBar.md
```diff
@@ -0,0 +1,536 @@
...
+
+ Note: Users should **always** have the option to dismiss informational notifications.
+
+# API Notes
+
+### Notable Properties  
+
+| Name | Description |
+|:-:|:--|
+| IsOpen| Gets or sets a value that determines the visibility of the InfoBar. By default, is set to false. |
```
teaP, line 367: What is the reason to have this in addition to the usual Visibility property?

active/InformationBar/InformationBar.md
```diff
@@ -0,0 +1,536 @@
...
+| Severity | Gets or sets a value that indicates the  color and icon to style the InfoBar. It will also define the assistive technology settings. |
+| IsCloseButtonVisible| Gets or sets a boolean that indicates whether a close button will appear
+
+
+### Events  
+| Name | Description |
+|:-:|:--|
+| ActionButtonClick | Occurs after the action button has been tapped. |
```
teaP, line 375: The comps sometimes show two action buttons, but it looks like there's only ever one, right?

active/InformationBar/InformationBar.md
teaP: (I'm guessing, after further reading, that it's so there are events around Closing/Closed and the reason for it, etc.)

active/InformationBar/InformationBar.md
```diff
@@ -0,0 +1,536 @@
...
+> This spec corresponds to [issue 913](https://github.com/microsoft/microsoft-ui-xaml/issues/913) on the WinUI repo.
+
+Users should be informed about essential status changes that occur on an app level.
+These status changes affect the app as a whole and can be either critical or informational.
+Critical status changes like lost internet connectivity are directly impactful to app functionality while informational status changes like an update has completed and been applied are indirectly impactful to app functionality.
+These notifications and corresponding information should be presented in a consistent, predictable, and relevant way to the user depending on the specific scenario.
+
+Currently, Teaching Tip, Content Dialog, and customizations of other flyouts and dialogs exist as options to show these notifications but these controls were not specifically designed to handle app-wide status change notifications. 
+Due to their visual layouts, inherent intrusiveness, or available features they are not sufficient for displaying notifications at an app-wide level.
+
```
MikeHillberg, line 11:
```suggestion

This spec introduces an InfoBar API to ...
```

active/InformationBar/InformationBar.md
```diff
@@ -0,0 +1,536 @@
+# Background
+> This spec corresponds to [issue 913](https://github.com/microsoft/microsoft-ui-xaml/issues/913) on the WinUI repo.
+
+Users should be informed about essential status changes that occur on an app level.
```
MikeHillberg, line 4: Would it be easy to add a couple of pictures?

active/InformationBar/InformationBar.md
```diff
@@ -0,0 +1,536 @@
...
+These status changes affect the app as a whole and can be either critical or informational.
+Critical status changes like lost internet connectivity are directly impactful to app functionality while informational status changes like an update has completed and been applied are indirectly impactful to app functionality.
+These notifications and corresponding information should be presented in a consistent, predictable, and relevant way to the user depending on the specific scenario.
+
+Currently, Teaching Tip, Content Dialog, and customizations of other flyouts and dialogs exist as options to show these notifications but these controls were not specifically designed to handle app-wide status change notifications. 
+Due to their visual layouts, inherent intrusiveness, or available features they are not sufficient for displaying notifications at an app-wide level.
+
+# Description
+
+An InfoBar is a persistent, actionable, app-wide notification intended for displaying critical and/or actionable information that impact app perception or user experience.
```
MikeHillberg, line 14: (Minor comment at this stage) This text is the IntelliSense you'll get from typing <InfoBar>. For Xaml controls it always seems to start with "Represents", and tends to focus more on the 'what'.

```suggestion
Represents an informational control that is visible but does not interfere with the user. 
```

active/InformationBar/InformationBar.md
```diff
@@ -0,0 +1,536 @@
...
+
+## Is this the right control?
+Use an InfoBar control when a user needs to be informed of, acknowledge, or take action on a message. By default the notification will remain in the content area until dismissed by the user but will not necessarily break user flow.
```
MikeHillberg, line 17: "User needs to take action" makes this sound modal, like a ContentDialog

active/InformationBar/InformationBar.md
```diff
@@ -0,0 +1,536 @@
...
+
+An InfoBar will take up space in your layout and behave like any other child elements. It will not cover up other content or float on top of it.
+
+Do not use an InfoBar control to confirm or respond directly to a user action, for transient alerts, or for non-essential messages.
```
MikeHillberg, line 21: Why not use this for an essential transient message? Like network connectivity?

active/InformationBar/InformationBar.md
```diff
@@ -0,0 +1,536 @@
...
+
+### Scenarios
```
MikeHillberg, line 23:
```suggestion
### Remarks
```

active/InformationBar/InformationBar.md
```diff
@@ -0,0 +1,536 @@
...
+Use an InfoBar that dismiss via the user or a timer for scenarios that **indirectly** impact app perception or experience ℹ
+
+- A call has begun recording
+- Update applied with link to 'Release Notes'
+- The terms of service have been updated and require acknowledgement
+- An app-wide backup has successfully, asynchronously completed
+- The subscription to the application is close to expiring
+
+
+### When should a different control be used?
```
MikeHillberg, line 41:
```suggestion
#### When should a different control be used?
```

active/InformationBar/InformationBar.md
```diff
@@ -0,0 +1,536 @@
...
+
+There are some scenarios where a Content Dialog, Flyout, or Teaching Tip may be more appropriate to use.
```
MikeHillberg, line 43:
```suggestion
There are some scenarios where a ContentDialog, Flyout, or TeachingTip may be more appropriate to use.
```

active/InformationBar/InformationBar.md
```diff
@@ -0,0 +1,536 @@
...
+
+- For scenarios where a persistent notification is not needed, i.e. displaying information in context of a specific UI element, a [Flyout](https://docs.microsoft.com/en-us/windows/uwp/design/controls-and-patterns/dialogs-and-flyouts/flyouts) is a better option.
```
MikeHillberg, line 45:
```suggestion
- For scenarios where a persistent notification is not needed, e.g. displaying information in context of a specific UI element, a [Flyout](https://docs.microsoft.com/en-us/windows/uwp/design/controls-and-patterns/dialogs-and-flyouts/flyouts) is a better option. 
```

active/InformationBar/InformationBar.md
```diff
@@ -0,0 +1,536 @@
...
+  - Additionally, if a status change to the app is so severe that it needs to block all further ability for the user to interact with the app, use a Content Dialog.
+- For scenarios where a notification is a transient teaching moment, a [Teaching Tip](https://docs.microsoft.com/en-us/windows/uwp/design/controls-and-patterns/dialogs-and-flyouts/teaching-tip) is a better option.
+
+For more info about choosing the right notification control, see the [Dialogs and Flyouts](https://docs.microsoft.com/en-us/windows/uwp/design/controls-and-patterns/dialogs-and-flyouts/) article.
+
+# Examples
+
+An information bar can have several configurations, here are some notable ones.
+
+When an InfoBar is conveying information of a common severity, like displaying an error or warning, it can be set to have one of many severity levels to use consistent Fluent styling for it's identifiers.
```
MikeHillberg, line 56: Link to what "Fluent" means?

active/InformationBar/InformationBar.md
```diff
@@ -0,0 +1,536 @@
...
+C#
+```C#
+public MainPage()
+{
+    this.InitializeComponent();
+
+    if(!SaveSuccessful())
+    {
+        UnsuccessfulSaveNotification.IsOpen = true;
+        SetHaveDisplayedErrorNotification();
```
MikeHillberg, line 94: What is this line for? Is it relevant to the example?

active/InformationBar/InformationBar.md
```diff
@@ -0,0 +1,536 @@
...
+    }
+}
+```
+
+Here is the visual representation of the information bar in the page.
+
+![A mockup of an InfoBar in a Warning state with a close button and a message](images/Warning_NoButton.png)
+
+## Using pre-defined severity levels
+The type of the information bar can be set via the Severity property to automatically set a consistent status color, icon, and assistive technology settings dependent on the criticality of the notification.
```
MikeHillberg, line 104:
```suggestion
The type of the info bar can be set via the Severity property to automatically set a consistent status color, icon, and assistive technology settings dependent on the criticality of the notification.
```

active/InformationBar/InformationBar.md
```diff
@@ -0,0 +1,536 @@
...
+        SetHaveDisplayedErrorNotification();
+    }
+}
+```
+
+Here is the visual representation of the information bar in the page.
+
+![A mockup of an InfoBar in a Warning state with a close button and a message](images/Warning_NoButton.png)
+
+## Using pre-defined severity levels
```
MikeHillberg, line 103: Add a code example?

active/InformationBar/InformationBar.md
```diff
@@ -0,0 +1,536 @@
...
+        IsCloseButtonVisible="False">
+    </controls:InfoBar>
+</StackPanel>
+```
+![Mockup of an InfoBar in a Critical state with no close button](images/Critical_IconNoClose.png)
+
+## Custom styling: status color and icon
+Outside of the pre-defined severity levels, the Background and IconSource properties can be set to customize the icon and background color. The info bar will retain the assistive technology settings of the severity defined, or default if none was defined.
+
+A custom background color can be set via the standard Background property and will override the color set by Severity. Please keep in mind content readability and accessibility when setting your own color.
```
MikeHillberg, line 132: Are the accessibility features explained anywhere? Maybe should be in the remarks

active/InformationBar/InformationBar.md
```diff
@@ -0,0 +1,536 @@
...
+
+A custom icon can be set via the IconSource property left of the Title and Message in the InfoBar. By default, an icon will be visible. This icon can be removed by setting the IsIconVisible property to false. For custom icons, the recommended icon size is 20px.
```
MikeHillberg, line 134: Will it be scaled to 20px if necessary?

MikeHillberg, line 134:
```suggestion
A custom icon can be set via the IconSource property. By default, an icon will be visible. This icon can be removed by setting the IsIconVisible property to false. For custom icons, the recommended icon size is 20px.
```

(Don't need to get into the current visual design details in the docs)

active/InformationBar/InformationBar.md
```diff
@@ -0,0 +1,536 @@
...
+
+XAML
+```xml
+<StackPanel x:Name="ContentArea">
+    <controls:InfoBar x:Name="NewPhotosNotification"
+        Title="New Photos From Cloud"
+        Message="Sync your latest photos to the Cloud"
+        ActionButtonContent="Upload">
+        <controls:InfoBar.Background>
```
MikeHillberg, line 143: Can do both Background and IconSource as attributes rather than tags

active/InformationBar/InformationBar.md
```diff
@@ -0,0 +1,536 @@
...
+
+XAML
+```xml
+<StackPanel x:Name="ContentArea">
+    <controls:InfoBar x:Name="UnsuccessfulSaveNotification"
+        Severity="Warning"
+        Title="Error while saving"
+        Message="Your document was unable to be saved."
+        ActionButtonContent="Try again"
+        ActionButtonCommand="SaveDocuments">
```
MikeHillberg, line 168:
```suggestion
        ActionButtonCommand="{x:Bind SaveDocuments}">
```

active/InformationBar/InformationBar.md
```diff
@@ -0,0 +1,536 @@
...
+
+![A mockup of an InfoBar with a single line message and an action button](images/Warning_Button.png)
+
+XAML
+```xml
+<StackPanel x:Name="ContentArea">
+    <controls:InfoBar x:Name="ConnectionErrorNotification"
+        Title="Error while saving"
+        Message="Lorem ipsum long message">
+        <controls:InfoBar.HyperlinkButtonContent>
```
MikeHillberg, line 181: It's odd that I can provide an actual HyperlinkButton object but not a Button object.

active/InformationBar/InformationBar.md
```diff
@@ -0,0 +1,536 @@
...
+</StackPanel>
+```
+![A mockup of an InfoBar with a single line message and hyperlink](images/Default_MessageHyperlink.png)
+![A mockup of an InfoBar with a message expanding multiple lines and a hyperlink](images/Docked_MultiLineIconTitleHyperlink.png)
+
+## Custom content
+Content can be added to an InfoBar using the Content property. It will appear below the message and above the action or hyperlink buttons if defined. The InfoBar will expand to fit the content defined.
+
+XAML
+```xml
```
MikeHillberg, line 196: Looks like the wrong example code (hyperlink rather than content)

active/InformationBar/InformationBar.md
```diff
@@ -0,0 +1,536 @@
...
+        </controls:InfoBar.HyperlinkButtonContent>
+    </controls:InfoBar>
+</StackPanel>
+```
+
+![A sketch of an InfoBar in its default state with a hyperlink and message](images/Information_CustomContent.jpg)
+Mockup with custom content to be added.
+
+## Content wrapping
+By default, the text set in the Message property will wrap vertically in the control underneath the other visual components.
```
MikeHillberg, line 212: Is there a way to override the default?

active/InformationBar/InformationBar.md
```diff
@@ -0,0 +1,536 @@
...
+
+XAML
+```xml
+<StackPanel x:Name="ContentArea">
```
MikeHillberg, line 216: When the info bar is closed, is there an animation? Or a recommended way for me to add an animation?

active/InformationBar/InformationBar.md
gabbybilka: Do you mean throughout the document or in the Background section specifically? There should be pictures/mockups throughout the spec...

active/InformationBar/InformationBar.md
gabbybilka: Should there be? The alternative would be to allow the content to cut itself off once it is too long. That's an option that Fluent UI's [MessageBar](https://developer.microsoft.com/en-us/fluentui#/controls/web/messagebar) has in their isMultiline property but I struggle to see a scenario where we would want to support that.

Or are you more focused on the wording of "by default"?

active/InformationBar/InformationBar.md
gabbybilka: Currently, no.  I think by default no animations should be shown for these inline InfoBars however if a developer wanted to include their own -- e.g. coming down from the closest orientation (top, bottom, left, right) -- then I think that should be supported. Via storyboards? @teaP?

active/InformationBar/InformationBar.md
gabbybilka: I don't think they should as I don't believe other controls in the past have. @StephenLPeters is this correct for Teaching Tip?

active/InformationBar/InformationBar.md
```diff
@@ -0,0 +1,512 @@
+# Background
+> This spec corresponds to [issue 913](https://github.com/microsoft/microsoft-ui-xaml/issues/913) on the WinUI repo.
+
+Users should be informed about essential status changes that occur on an app level.
+These status changes affect the app as a whole and can be either critical or informational.
+Critical status changes like lost internet connectivity are directly impactful to app functionality while informational status changes like an update has completed and been applied are indirectly impactful to app functionality.
+These notifications and corresponding information should be presented in a consistent, predictable, and relevant way to the user depending on the specific scenario.
+
+Currently, Teaching Tip, Content Dialog, and customizations of other flyouts and dialogs exist as options to show these notifications but these controls were not specifically designed to handle app-wide status change notifications.
```
MikeHillberg, line 9: About "other ... dialogs", what other dialogs beyond ContentDialog?

active/InformationBar/InformationBar.md
```diff
@@ -0,0 +1,512 @@
...
+- Error while saving a document when triggered automatically, not related to specific user action
+- No microphone plugged in when attempting to record
+- Can't connect to your phone
+- The subscription to the application is expired
+
+Use an InfoBar that dismiss via the user or a timer for scenarios that **indirectly** impact app perception or experience ℹ
+
+- A call has begun recording
+- Update applied with link to 'Release Notes'
+- The terms of service have been updated and require acknowledgement
```
MikeHillberg, line 36: If it requires user acknowledgement, wouldn't it be bad for a timer to dismiss it?

active/InformationBar/InformationBar.md
chingucoding: There's also the MessageDialog.

active/InformationBar/InformationBar.md
```diff
@@ -0,0 +1,512 @@
...
+  - Additionally, if a status change to the app is so severe that it needs to block all further ability for the user to interact with the app, use a ContentDialog.
+- For scenarios where a notification is a transient teaching moment, a [TeachingTip](https://docs.microsoft.com/en-us/windows/uwp/design/controls-and-patterns/dialogs-and-flyouts/teaching-tip) is a better option.
+
+For more info about choosing the right notification control, see the [Dialogs and Flyouts](https://docs.microsoft.com/en-us/windows/uwp/design/controls-and-patterns/dialogs-and-flyouts/) article.
+
+# Examples
+
+An info bar can have several configurations, here are some notable ones.
+
+When an InfoBar is conveying information of a common severity, like displaying an error or warning, it can be set to have one of many severity levels to use consistent [Fluent](https://docs.microsoft.com/en-us/windows/apps/fluent-design-system) styling for it's identifiers.
```
MikeHillberg, line 66: I don't understand the connection between the text and the picture. Is this example using a severity level? I would just remove this first section, up to the "Create an InfoBar" example, as it's redundant of the later examples. It provides an introduction/overview of the follow examples, but that's not a style that we use in the docs. It's better to jump right in to the first example.

MikeHillberg, line 66: These examples should include example code

active/InformationBar/InformationBar.md
```diff
@@ -0,0 +1,512 @@
...
+```
+
+Here is the visual representation of the info bar in the page.
+
+![A mockup of an InfoBar in a Warning state with a close button and a message](images/Warning_NoButton.png)
+
+## Using pre-defined severity levels
+The type of the info bar can be set via the Severity property to automatically set a consistent status color, icon, and assistive technology settings dependent on the criticality of the notification.
+
+Mockups in various Severities to be added.
```
MikeHillberg, line 118: We don't usually work with mockups in the documentation. These would be better as examples (including the markup that creates them).

active/InformationBar/InformationBar.md
```diff
@@ -0,0 +1,512 @@
...
+![Mockup of InfoBar with no content in different Severity colors and icons for light mode](images/Docked_ToolkitSeverity.png)
+
+![Mockup of InfoBar with no content in different Severity colors and icons for dark mode](images/Docked_ToolkitSeverityDark.png)
+
+## Programmatic dismiss in info bar
+An info bar can be dismissed by the user via the close button or programmatically. If the notification is required to be in view until the status is resolved and you would like to remove the ability for the user to dismiss the info bar, you can set the IsUserDismissable property to false.
```
MikeHillberg: "Close" is a more standard term for this action than "dismiss". And this looks more like a case for the "Can" pattern. So instead of "IsUserDismissable" I think "CanClose" would be a better name. (For example similar to [TextBox.CanUndo](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.Controls.TextBox.CanUndo).)

active/InformationBar/InformationBar.md
```diff
@@ -0,0 +1,512 @@
...
+                NavigateUri="http://www.microsoft.com"/>
+        <controls:InfoBar.ActionButton/>
+    </controls:InfoBar>
+</StackPanel>
+```
+![A mockup of an InfoBar with a single line message and hyperlink](images/Default_MessageHyperlink.png)
+![A mockup of an InfoBar with a message expanding multiple lines and a hyperlink](images/Docked_MultiLineIconTitleHyperlink.png)
+
+## Custom content
+Content can be added to an InfoBar using the Content property. It will appear below the message and above the action or hyperlink buttons if defined and multiline. The InfoBar will expand to fit the content defined.
```
MikeHillberg, line 207: What does it mean to be multiline?

(Maybe this is answered in the next example and they just need to re-ordered)

active/InformationBar/InformationBar.md
MikeHillberg: Yeah, it says "by default" indicating that there's a non-default case, but not explaining it.

active/InformationBar/InformationBar.md
MikeHillberg: The text in the Canceling example teases the possibility of an animation. Do we need to give an example of how to do one?

active/InformationBar/InformationBar.md
```diff
@@ -0,0 +1,512 @@
...
+{
+    if (args.Reason == InfoBarCloseReason.CloseButton) {
+        // if scenario failed
+        args.Cancel = true;
+    }
+    
+}
+```
+
+# Inputs and Accessibility
```
MikeHillberg, line 266: The public docs don't usually go into this much detail, but it's great detail. This can be moved to an Appendix of this spec.

active/InformationBar/InformationBar.md
```diff
@@ -0,0 +1,512 @@
...
+
+| Severity |  NotificationProcessing | Behavior in Narrator|
+|:--- | :---| :---|
+| Error | NotificationProcessing_ImportantAll| Add new item to the end of the queue. It doesn’t matter the source, what is currently speaking or any other items in the queue queued with a higher priority than normal text. <br><br> This will cause it to get spoken after the current utterance/string finishes. Meaning it won't interrupt the speech but it will queue itself next to be spoken, ahead of anything already queued. <br><br> Focus and keyboard will NOT interrupt or flush, pressing control will silence and flush all |
+| Warning | NotificationProcessing_ImportantAll| See above |
+| Success | Processing_All| Add new item to the end of the queue. <br><br> Will be added to the end of the queue meaning all existing queued text will need to be spoken before this text will be spoken. <br><br>Focus, keyboard and control will silence/flush them all
+| Default | Processing_All | See above
+
+#### Narrator
```
MikeHillberg, line 298: Why are we talking about Narrator? Are we making updates to that app, or does this section apply to any screen reader?

active/InformationBar/InformationBar.md
```diff
@@ -0,0 +1,512 @@
...
+## Message 
+
+Text in your notification will not be a constant length in all languages. For the Title and Message property this may impact whether your notification will expand to a second line. 
+We recommend you avoid positioning based on message length or other UI elements set to a specific language.
+
+The notification will follow standard mirroring behavior when localized to/from languages that are right to left (RTL) or left to right (LTR). The icon will only mirror if there is directionality.
+
+Please view the guidance for [Adjust layout and fonts, and support RTL](https://docs.microsoft.com/en-us/windows/uwp/design/globalizing/adjust-layout-and-fonts--and-support-rtl) for more information about text localization in your notification.
+
+# Remarks
```
MikeHillberg, line 350: How are these remarks different than the earlier remarks? Where are there two versions?

active/InformationBar/InformationBar.md
```diff
@@ -0,0 +1,512 @@
...
+|:-:|:--|
+| IsOpen| Gets or sets a value that determines the visibility of the InfoBar. By default, is set to false. |
+| Severity | Gets or sets a value that indicates the  color and icon to style the InfoBar. It will also define the assistive technology settings. |
+| IsUserDismissable| Gets or sets a boolean that indicates whether the user will be able to dismiss the InfoBar.
+
+
+### Events  
+| Name | Description |
+|:-:|:--|
+| CloseButtonClick | Occurs after the close button has been tapped. |
```
MikeHillberg, line 387: "tapped" => "clicked"  ("Tap" is a touch term, and you can use a mouse or keyboard on a button.)

active/InformationBar/InformationBar.md
MikeHillberg: Good point, thanks.

active/InformationBar/InformationBar.md
gabbybilka: We are not making updates to Narrator however when describing the effect of the [UIA Event](https://docs.microsoft.com/en-us/windows/win32/winauto/uiauto-eventsoverview) to screen readers I've seen Narrator used in other specs as example behavior.

active/InformationBar/InformationBar.md
gabbybilka: I'm open to that, could we leave this as something to be discussed at the API Review?

active/InformationBar/InformationBar.md
gabbybilka: When you mean mockups do you mean the File Explorer sample app screenshots? I can create a simpler sample app to grab screenshots to highlight the functionality if that's valuable at this stage. 
For the content wrapping example, showing just the InfoBar on its own is not as valuable as showing where it is on the page and its relation to other content. For the markup snippet, is it beneficial to include the accompanying layout controls?

active/InformationBar/InformationBar.md
gabbybilka: Removed mention of animation 😊

active/InformationBar/InformationBar.md
```diff
@@ -0,0 +1,503 @@
...
+Use an InfoBar control when a user should be informed of, acknowledge, or take action on a message. By default the notification will remain in the content area until dismissed by the user but will not necessarily break user flow.
+
+An InfoBar will take up space in your layout and behave like any other child elements. It will not cover up other content or float on top of it.
+
+Do not use an InfoBar control to confirm or respond directly to a user action, for time-sensitive alerts, or for non-essential messages.
+
+## Remarks
+Use an InfoBar that is dismissed by the user or when the status is resolved for scenarios that **directly** impact app perception or experience
+
+- Internet connectivity lost
```
ranjeshj, line 34: nit: list seems to be disconnected from the sentence above. Perhaps add a line like. 'Here are some examples'

active/InformationBar/InformationBar.md
```diff
@@ -0,0 +1,503 @@
...
+  - Additionally, if a status change to the app is so severe that it needs to block all further ability for the user to interact with the app, use a ContentDialog.
+- For scenarios where a notification is a transient teaching moment, a [TeachingTip](https://docs.microsoft.com/en-us/windows/uwp/design/controls-and-patterns/dialogs-and-flyouts/teaching-tip) is a better option.
+
+For more info about choosing the right notification control, see the [Dialogs and Flyouts](https://docs.microsoft.com/en-us/windows/uwp/design/controls-and-patterns/dialogs-and-flyouts/) article.
+
+# Examples
+
+## Create an InfoBar
+
+The XAML below describes an inline InfoBar with the default styling for an informational notification. An info bar can be created anywhere in the element tree or code behind. In this example the InfoBar expands to fill the width of the StackPanel layout control it resides in.
```
ranjeshj, line 66: >In this example the InfoBar expands to fill the width of the StackPanel layout control it resides in. [](start = 175, length = 101)

I don't see the StackPanel in the example below. I don't think we need this since it behaves like any other control.

active/InformationBar/InformationBar.md
```diff
@@ -0,0 +1,622 @@
...
+|:-:|:--|
+|InfoBarHyperlinkForeground | Sets the hyperlink button text color. <br> - Note: This is set to keep hyperlinks accessible on the variously colored backgrounds defined by the severity background colors.
+|InfoBarCloseButtonStyle | Sets the close button style of the InfoBar and contains the following properties: <br> - Width <br> - Height <br> - VerticalAlignment <br> - Background <br> - Margin <br> - CornerRadius
+|InfoBarIconFontSize | Sets the icon size of the InfoBar. <br> - Note: Custom icons set via the IconSource property will not resize to 20px unless they are in the standard symbol set.
+
+### All Theme Resources
+| Resource Key | Description |
+|:-:|:--|
+| InfoBarSeverityErrorBackgroundBrush | Error Severity background color 
+| InfoBarSeverityWarningBackgroundBrush | Warning Severity background color
```
MikeHillberg, line 511: As the description implies, this should be ...**WarningSeverity**BackgroundBrush rather than ...SeverityWarning...

active/InformationBar/InformationBar.md
```diff
@@ -0,0 +1,622 @@
...
+    Severity="Warning"
+    Title="Your subscription is expiring in 3 days"
+    Message="Renew your subscription to keep all functionality" />
+```
+![A mockup of an InfoBar in a Warning state with a close button and a message](images/Warning_TitleMessage.png)
+
+## Programmatic dismiss in InfoBar
+An InfoBar can be dismissed by the user via the close button or programmatically. 
+If the notification is required to be in view until the status is resolved and you would like to remove 
+the ability for the user to dismiss the info bar, you can set the IsUserDismissable property to false.
```
MikeHillberg, line 139: "Close" is a more standard term for this action than "dismiss". And this should follow the "Can" pattern for consistency. So "CanClose" rather than "IsUserDismissable". (For example similar to [TextBox.CanUndo](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.Controls.TextBox.CanUndo).)

active/InformationBar/InformationBar.md
```diff
@@ -0,0 +1,622 @@
...
+
+## Remarks
+Use an InfoBar that is dismissed by the user or when the status is resolved for scenarios 
+that **directly** impact app perception or experience
+
+
+Here are some examples:
+- Internet connectivity lost
+- Error while saving a document when triggered automatically, not related to specific user action
+- No microphone plugged in when attempting to record
```
ptorr-msft, line 52: This sounds like it was responding to "user action" (start recording) but that is listed as a "do not use" above.

active/InformationBar/InformationBar.md
```diff
@@ -0,0 +1,622 @@
...
+For more info about choosing the right notification control, see the 
+[Dialogs and Flyouts](https://docs.microsoft.com/en-us/windows/uwp/design/controls-and-patterns/dialogs-and-flyouts/)
+article.
+
+
+# Examples
+
+## Create an InfoBar
+The XAML below describes an inline InfoBar with the default styling for an informational notification. 
+An info bar can be created anywhere in the element tree or code behind and will expand
```
ptorr-msft, line 93: "anywhere" in the tree -- does the parent need to support (multiple) children? E.g. could I put it on a `Border`, even if the `Border` already has content?

active/InformationBar/InformationBar.md
```diff
@@ -0,0 +1,622 @@
...
+
+
+XAML
+```xml
+<controls:InfoBar x:Name="NoInternetNotification"
+    Severity="Error"
+    Title="No Internet"
+    Message="Reconnect to continue working.">
+    <controls:InfoBar.ActionButton>
+        <Button Content="Network Settings"/>
```
ptorr-msft, line 202: Do I need to add a `Click` handler to this `Button`, or is there an `ActionButtonClicked` event on the `InfoBar`?

active/InformationBar/InformationBar.md
```diff
@@ -0,0 +1,622 @@
...
+
+XAML
+```xml
+<controls:InfoBar x:Name="TermsAndConditionsUpdatedNotification"
+    Title="Terms and Conditions Updated"
+    Message="Dismissal of this message implies agreement to the updated Terms and Conditions. Please view the changes on our website.">
+    <controls:InfoBar.ActionButton>
+        <HyperlinkButton
+            Content="Terms and Conditions Sep 2020" 
+            NavigateUri="http://www.microsoft.com"/>
```
ptorr-msft, line 217: Nit: should we use something like `example.com` rather than Microsoft?

active/InformationBar/InformationBar.md
```diff
@@ -0,0 +1,622 @@
...
+    Title="Backup complete: Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua."  
+    Message="All documents were uploaded successfully. Ultrices sagittis orci a scelerisque. Aliquet risus feugiat in ante metus dictum at tempor commodo. Auctor augue mauris augue neque gravida." />
+```
+
+![A mockup of an InfoBar in the Success state multiline title and message InfoBar](images/Success_ContentWrapping.png)
+
+
+## Custom content
+XAML content can be added to an InfoBar using the Content property. 
+It will appear in its own line below the rest of the control content.
```
ptorr-msft, line 242: Below the `ActionButton` as well?

active/InformationBar/InformationBar.md
```diff
@@ -0,0 +1,622 @@
...
+```c++
+enum InfoBarCloseReason
+{
+    CloseButton = 0,
+    Programmatic = 1,
+};
+
+enum InfoBarSeverity
+{
+    Error = 0,
```
ptorr-msft, line 427: Usually the "default" value is set as the `0` value, and so `Default` should be `0`.

active/InformationBar/InformationBar.md
```diff
@@ -0,0 +1,622 @@
...
+runtimeclass InfoBarClosedEventArgs
+{
+    InfoBarCloseReason Reason{ get; };
+};
+
+runtimeclass InfoBarClosingEventArgs
+{
+    InfoBarCloseReason Reason{ get; };
+    Boolean Cancel;
+    Windows.Foundation.Deferral GetDeferral();
```
ptorr-msft, line 442: Is there a good scenario for deferral? Is there a deadline for the deferral? Is it any different than the app specifying `Cancel = true`, disabling input on the control, doing the async work, then programmatically closing the bar?

active/InformationBar/InformationBar.md
```diff
@@ -0,0 +1,622 @@
...
+
+unsealed runtimeclass InfoBar : Windows.UI.Xaml.Controls.ContentControl
+{
+    InfoBar();
+
+    String Title;
+    String Message;
+
+    Boolean IsOpen;
+    Boolean IsUserDismissable;
```
ptorr-msft, line 461: Should this be `IsCloseButtonVisible`?

active/InformationBar/InformationBar.md
ranjeshj: There is nothing special about InfoBar in this sense. It is just another control. Border supports only one child, so this can be the child of the border. If you need multiple children, you need to add some panel (like StackPanel or Grid) and then add InfoBar to that panel.

active/InformationBar/InformationBar.md
ranjeshj: You need to add the Click handler. There is no event for this on the InfoBar. Note that the action button is optional.

active/InformationBar/InformationBar.md
gabbybilka: Yep, see below:
image

active/InformationBar/InformationBar.md
gabbybilka: @mikeh brought this up earlier in the [conversation](https://github.com/microsoft/microsoft-ui-xaml-specs/pull/94#discussion_r487346017):
> "Close" is a more standard term for this action than "dismiss". And this should follow the "Can" pattern for consistency. So "CanClose" rather than "IsUserDismissable". (For example similar to [TextBox.CanUndo](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.Controls.TextBox.CanUndo).)

We avoided IsCloseButtonVisible to remain flexible with devices in the future (like touch-enabled) where a close button may not be visible but the user is still able to close the bar via a swipe or other gesture. What are your thoughts on CanClose?

mikeh: Guessing you mis-tagged and this was meant for @MikeHillberg :-)

active/InformationBar/InformationBar.md
gabbybilka: One example scenario is if the InfoBar is showing a "Terms and Conditions are updated and need to be accepted" message and there's an inline checkbox that a user needs to check before the InfoBar closes. If the user manually closes the InfoBar but the checkbox isn't checked then the close could be cancelled.

@teaP for the deferral deadline and replicating deferral behavior questions?

active/InformationBar/InformationBar.md
ptorr-msft: Suggest adding it to the sample then to make it clear :)

active/InformationBar/InformationBar.md
ptorr-msft: `Cancel` makes perfect sense; it's the deferral that is potentially not useful. Usually it is used if you need to make a network connection or write to a brokered file.

active/InformationBar/InformationBar.md
ptorr-msft: Sure, `CanClose` is good too.

active/InformationBar/InformationBar.md
MikeHillberg: In this case it's to enable a dialog and user interaction before closing the InfoBar, which like files and network is async.

active/InformationBar/InformationBar.md
```diff
@@ -0,0 +1,622 @@
...
+
+    InfoBarSeverity Severity;
+    IconSource IconSource;
+
+    InfoBarTemplateSettings TemplateSettings{ get; };
+    Object Content{ get; set; };
+    Windows.UI.Xaml.DataTemplate ContentTemplate{ get; set; };
+
+    event Windows.Foundation.TypedEventHandler<InfoBar, Object> CloseButtonClick;
+    event Windows.Foundation.TypedEventHandler<InfoBar, InfoBarClosingEventArgs> Closing;
```
oldnewthing, line 478: The fact that you can trigger `Closing` by setting `IsOpen=false` means that inside the `Closing` handler, if you read the `IsOpen` property, you get `false` which is weird because you expect `Closing` to be raised *before* it closes. (As noted later, if you then cancel the Closing event, this sets `IsOpen` back to `true`.)

If the InfoBar is interactively closed, does `IsOpen` read as true or false during `Closing`?

`NavigationView` has the same problem, since you can set `IsPaneOpen = !IsPaneOpen` and the `-ing` events will be raised even though the transition has already occurred as far as the property is concerned. Are we at least consistent with NavigationView's behavior here?

active/InformationBar/InformationBar.md
```diff
@@ -0,0 +1,622 @@
...
+| Action button |  - Will appear directly to the right of the Message in single height notifications, otherwise will be on its own new line. <br> - An action button can be any class that inherits ButtonBase however Button, HyperlinkButton, and Checkbox are our recommended options <br> - Additional action buttons may be added through custom XAML content
+| Close button | - Will appear as 'X' by default <br> - Can be removed by setting IsUserDismissable to false
+| Content | - Can be customizable to include text, hyperlinks, and any other XAML content <br> - Appears in its own line underneath the other InfoBar content
+
+## Behavioral Components
+| Property | Notes |
+|:---:|:---|
+| Opening | - An info bar is shown by setting its IsOpen property to true. |
+| Closing | There are two ways an info bar can close: <br>- The program sets the IsOpen property to false <br> - The user invokes the Close button. <br> Use the InfoBarCloseReason to determine which case has occurred. <br> Closing can be prevented by setting the Cancel property to true. You can use a deferral to respond asynchronously to the event. |
```
oldnewthing, line 571: 3. The user presses the gamepad B button.

active/InformationBar/InformationBar.md
```diff
@@ -0,0 +1,622 @@
...
+    Message="Please close this tip to apply required security updates to this application"
+    Closing="InfoBar_Closing">
+</controls:InfoBar>
+```
+C#
+```C#
+public void InfoBar_Closing(InfoBar sender, InfoBarClosingEventArgs args)
+{
+    if (args.Reason == InfoBarCloseReason.CloseButton) {
+        // if scenario failed
```
oldnewthing, line 593: Use of comment here is confusing. This isn't commenting the code; it's representing code that is missing. Better would be to call a function not implemented here.

```c#
if (args.Reason == InfoBarCloseReason.CloseButton) {
  if (!ApplyUpdates()) {
    // could not apply updates - do not close
    args.Cancel = true;
  }
}
```

active/InformationBar/InformationBar.md
```diff
@@ -0,0 +1,622 @@
...
+```xml
+<Page.Resources>
+    <x:Double x:Key="InfoBarTitleFontSize">22</x:Double>
+</Page.Resources>
+```
+
+# Inputs and Accessibility
+
+## UI Automation Patterns 
+InfoBar will use a Pane control pattern for inline notifications and will implement a custom "information" Landmark.
```
oldnewthing, line 280: Does a closed InfoBar appear in the automation tree at all?

active/InformationBar/InformationBar.md
```diff
@@ -0,0 +1,622 @@
...
+[Dialogs and Flyouts](https://docs.microsoft.com/en-us/windows/uwp/design/controls-and-patterns/dialogs-and-flyouts/)
+article.
+
+
+# Examples
+
+## Create an InfoBar
+The XAML below describes an inline InfoBar with the default styling for an informational notification. 
+An info bar can be created anywhere in the element tree or code behind and will expand 
+to fill the width of the layout control it resides in.
```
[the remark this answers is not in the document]
MikeHillberg: It's a good thing to call out, not sure the best way to word it. The control will use the full layout slot provided by the parent, subject to other standard layout features, like setting the HorizontalAlignment.

active/InformationBar/InformationBar.md
```diff
@@ -0,0 +1,622 @@
...
+            NavigateUri="http://www.microsoft.com"/>
+    <controls:InfoBar.ActionButton/>
+</controls:InfoBar>
+```
+![A mockup of an InfoBar with a message expanding multiple lines and a hyperlink](images/Default_Hyperlink.png)
+
+## Content wrapping
+The text set in the Message property will automatically wrap vertically in the control 
+underneath the other visual components if the control is no longer able to expand horizontally. 
+The action button will also wrap in its own line.
```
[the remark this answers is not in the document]
MikeHillberg: Make sense. But is it really a paragraph or a new line?

active/InformationBar/InformationBar.md
```diff
@@ -0,0 +1,622 @@
...
+# Inputs and Accessibility
+
+## UI Automation Patterns 
+InfoBar will use a Pane control pattern for inline notifications and will implement a custom "information" Landmark.
+
+
+### Keyboard Navigation 
+- No action is needed to invoke the InfoBar
+- After navigating to the control via tabbing, 
+the user is automatically focused into the control once navigating to the first actionable item.
```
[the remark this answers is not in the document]
MikeHillberg: I would just leave out the "in that order" in that suggestion. (Trying to say as little as possible about the current visualization)

active/InformationBar/InformationBar.md
```diff
@@ -0,0 +1,622 @@
...
+- Once focused, tabbing will iterate through all actionable items in the control in order. 
+When tab is pressed on the last element in the InfoBar, focus will continue throughout the page.
+  - A user can also use the left and right arrow keys to navigate between the available buttons
+- To close the InfoBar the action or close button needs to be pressed.
+  - Note: Escape will not close the InfoBar and will instead bubble up the command to the parent components.
+
+### Gamepad
+- The actionable items in the control can be navigated to via spatial navigation.
+- The 'A' button will interact with the item in focus, such as "press" the action or close buttons
+- The InfoBar can be closed via pressing the 'X' close button, pressing the action button, or the 'B' button which will return focus to the element previously in focus.
```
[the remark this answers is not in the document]
MikeHillberg: Is it odd that 'B' doesn't behave like Escape?

active/InformationBar/InformationBar.md
MikeHillberg: Is Default the default?

active/InformationBar/InformationBar.md
MikeHillberg: +1 to consistency, see also [FlyoutBase.IsOpen](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.Controls.Primitives.FlyoutBase.IsOpen)

active/InformationBar/InformationBar.md
```diff
@@ -0,0 +1,622 @@
...
+| InfoBarCloseButtonStyle | Close button style
+
+
+# Appendix
+
+## Design References
+UI Elements for InfoBar
+
+
+![Toolkit page for InfoBar UI elements](images/Docked_Toolkit.png)
```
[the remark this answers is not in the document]
gabbybilka: Need to update this with design.

active/InformationBar/InformationBar.md
```diff
@@ -0,0 +1,622 @@
...
+
+
+### Keyboard Navigation 
+- No action is needed to invoke the InfoBar
+- After navigating to the control via tabbing, 
+the user is automatically focused into the control once navigating to the first actionable item.
+- Once focused, tabbing will iterate through all actionable items in the control in order. 
+When tab is pressed on the last element in the InfoBar, focus will continue throughout the page.
+  - A user can also use the left and right arrow keys to navigate between the available buttons
+- To close the InfoBar the action or close button needs to be pressed.
```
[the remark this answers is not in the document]
gabbybilka: @MikeHillberg, who do you recommend I follow up on this with? Removing action button close functionality from spec for now as implementation doesn't include it.

active/InformationBar/InformationBar.md
```diff
@@ -0,0 +1,622 @@
...
+InfoBar will use a Pane control pattern for inline notifications and will implement a custom "information" Landmark.
+
+
+### Keyboard Navigation 
+- No action is needed to invoke the InfoBar
+- After navigating to the control via tabbing, 
+the user is automatically focused into the control once navigating to the first actionable item.
+- Once focused, tabbing will iterate through all actionable items in the control in order. 
+When tab is pressed on the last element in the InfoBar, focus will continue throughout the page.
+  - A user can also use the left and right arrow keys to navigate between the available buttons
```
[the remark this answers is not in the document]
gabbybilka: I'll follow up on this with accessibility.

active/InformationBar/InformationBar.md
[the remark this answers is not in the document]
gabbybilka: I'll follow up on the close button automation with accessibility.
@StephenLPeters, what happens in Teaching Tip today? When the close button is pressed what is announced? Is the verbage specific to TeachingTip or is it just "close button"?

active/InformationBar/InformationBar.md
gabbybilka: I think the term "responding to user action" may be too ambiguous.
> Do not use an InfoBar control to confirm or react directly to a user action when the state of the app is unchanged, for time-sensitive alerts, or for non-essential messages.

The examples that @oldnewthing included all are related to a changed state of the app in that is is now recording and/or in a meeting. I would argue that the final example, "the meeting has ended" would not be a message shown by an InfoBar as the state of the app is back in its default mode.

Thoughts?

MikeHillberg: Is that a suggestion to change the guideline text?

gabbybilka: Yep.

active/InformationBar/InformationBar.md
```diff
@@ -0,0 +1,512 @@
...
+        IsUserDismissable="False">
+    </controls:InfoBar>
+</StackPanel>
+```
+![Mockup of an InfoBar in a Critical state with no close button](images/Critical_IconNoClose.png)
+
+## Custom styling: status color and icon
+Outside of the pre-defined severity levels, the Background and IconSource properties can be set to customize the icon and background color. The info bar will retain the assistive technology settings of the severity defined, or default if none was defined.
+
+A custom background color can be set via the standard Background property and will override the color set by Severity. Please keep in mind content readability and accessibility when setting your own color.
```
gabbybilka: @MikeHillberg, @ranjeshj, @teaP 

Wanted to highlight an implementation distinction here. Currently, in contrast to the guidance here, the background color of the InfoBar can **not** be custom set via the Background property. Should it? Or, because background color is tied to the Severity level, that we recommend users to overwrite the associated Severity level background brush? @teaP mentioned that if we wanted to allow the Background property to override the background color set by the Severity that we could implement a background panel as an in-between for the Severity brushes and Background property.

Thoughts?

mdtauk: Should the severity set a base value which gets overwritten by the Background and Foreground chosen by the app devs?

Or does Severity override all?

Is there any contextual or functional value to having a set severity, but have the look of the control set differently by the app?

gabbybilka: Conceptually, a Severity is always going to be "set" in that even if a developer doesn't explicitly define a Severity it will follow the "informational" design with a pre-defined background color, icon, and accessibility settings. With that, I don't think Severity should override other custom values and should instead be a baseline.

My opinion is that if a developer wants to overwrite the background color independent of Severity they can. We're allowing them to set a custom icon without having them change the definition of the Severity for their app so why should there be a different pattern for the Background color?

MikeHillberg: It's unfortunate that Background is on Control, since it doesn't necessarily make sense on every control. But in this case I'd expect it to override whatever the control would otherwise do, regardless of the information level.

teaP: No problem, I'll make it work. 👍

robloo:
I have a custom control I've used for my apps which is kind of a 'hint' or 'message' view. I wonder if conceptionally this could be combined with something like the InfoBar which reports info/warnings/errors. 

Simple hint message indicating how the first element should be added.

image

More complete hint message with a title and glyph that can be turned on off.

image

Style-wide all controls are surrounded by a light border and text is the disabled brush and not hit-test visible. It is simply a placeholder message to guide users into an appropriate action.

mdtauk:
I think V2 is planning to have the InfoBar support non docked forms, where they are kind of like a card.

But your custom control @robloo kinda seems more like a Tooltip, just not displayed on hover.  But it could also be a TeachingTip

robloo:
@mdtauk, My example definitely overlaps with the TeachingTip in some areas. However, it is not shown in a popup and is embedded with the content. It is very contextually used in terms of placement as well.

I see the InfoBar as overlapping in some significant areas with TeachingTip as well. This should all probably have been generalized in a better way. I also don't see a good use case for non-popup contextual hints that are not in a title-bar format. Would be great to put all of this together.

mdtauk:
> @mdtauk, My example definitely overlaps with the TeachingTip in some areas. However, it is not shown in a popup and is embedded with the content. It is very contextually used in terms of placement as well.
> 
> I see the InfoBar as overlapping in some significant areas with TeachingTip. This should all probably have been generalized in a better way. I also don't see a good usage of InfoBar for non-popup contextual hints that are not in a title-bar format which is unfortunate because it could support this use case. Would be great to put all of this together.

error banner tweaks
This was the earlier design before the scope adapted to the current InfoBar / Message Bar - but overlayed over content aligned to a corner/side of the container is on the V2 list

robloo:
@mdtauk OK, glad that's on a v2 plan then. That would support my use cases for the most part and I can discuss finer details when the time comes. (I don't have time to go through most things and missed this control until I saw the PR)

mdtauk:
> @mdtauk OK, glad that's on a v2 plan then. That would support my use cases for the most part and I can discuss finer details when the time comes. (I don't have time to go through most things and missed this control until I saw the PR)
> 
> Edit: actually this seems only intended for global notification and action. It isn't really for contextual help but it certainly could be as the designs are very similar.

I just took a look at the most recent version of the specifications, and mentions of the inline card like layout has gone, so maybe those plans have been scrapped? @gabbybilka 

```
- Built-in support for floating notifications, DisplayMode property to switch between "Docked" (current) and "Floating" mode
  - Options to show the InfoBar as a PopUp with simple positioning properties
- Positioning and re-positioning for multiple notifications
```

robloo:
Probably should post this here too: Why is this called InfoBar? Info is only one small class of information it can contain. It also supports errors and warnings plus actions based on them. Bar is also artificially constraining the 'shape' of the control. There are several other use cases for it besides a bar.

NotificationView or MessageView, etc would be better names in my opinion.

gabbybilka:
@mdtauk and @robloo, 
Hopping in here for the design discussion. Good eye on noticing that V2 intended features change 😉  Right now we're not committing to floating mode for InfoBar in a V2 version for a couple of reasons. I'd like to further explore whether floating mode _should_ be in an InfoBar or whether toast-like notifications should be a different control altogether. While investigating we realized that a basic implementation of floating mode would be incredibly similar to Teaching Tip and that to fully explore toast-like notification scenarios functionality like stacking, positioning, and overlay choices (see WCT InAppNotification's [StackMode](https://github.com/windows-toolkit/WindowsCommunityToolkit/blob/master/Microsoft.Toolkit.Uwp.UI.Controls/InAppNotification/StackMode.cs)) are likely needed. This would expand the scope and intention of the InfoBar to be much broader than its initial focus. To note, we haven't fully shut the door on adding floating capabilities to InfoBar but leaning towards no.

@robloo for your custom control it seems that a floating mode version of InfoBar wouldn't meet your use-cases in that your hints and contextual information are inline. I think you _could_ use the InfoBar for your scenarios because it doesn't have to be a wide bar it will adjust based on your layout control or you could manually set its width. The visual design was focused on app-wide scenarios which the hints don't quite apply to as they are in-context with the current page/action. However, if you were satisfied with the visual aspects I don't see why you couldn't apply some custom styling (rounded corners, border color, etc.) to use InfoBar for your usecases.

mdtauk:
> @mdtauk and @robloo,
> Hopping in here for the design discussion. Good eye on noticing that V2 intended features change 😉 Right now we're not committing to floating mode for InfoBar in a V2 version for a couple of reasons. I'd like to further explore whether floating mode _should_ be in an InfoBar or whether toast-like notifications should be a different control altogether. While investigating we realized that a basic implementation of floating mode would be incredibly similar to Teaching Tip and that to fully explore toast-like notification scenarios functionality like stacking, positioning, and overlay choices (see WCT InAppNotification's [StackMode](https://github.com/windows-toolkit/WindowsCommunityToolkit/blob/master/Microsoft.Toolkit.Uwp.UI.Controls/InAppNotification/StackMode.cs)) are likely needed. This would expand the scope and intention of the InfoBar to be much broader than its initial focus. To note, we haven't fully shut the door on adding floating capabilities to InfoBar but leaning towards no.
> 
I can understand the change in behaviour and form would make you turn towards the TeachingTip as an alternative, but I would suggest the lack of severity, status colour, and the design of the teaching tip itself - would not be compatible with the same kind of messages that the InformationBar control is catering to.

The screen width and differences in UI layouts between a Full Screen WinUI app, a narrow width window, and things like Hololens and Xbox - would not lend themselves to a docked Bar form factor.

So with that in mind, perhaps the InformationBar should be wrapped, along with an InformationCard control into an **AppMessage** API - so the same properties of Severity, Colour, Title, Message, Action Buttons, etc can be placed into the XAML, or into the app's code - and the way they are presented on screen changes to suit the form factor/device family.

An unattached Teaching Tip could stand in, but it would not support those customisations that are suited for App Status Messages - which is the original intention behind this control.

gabbybilka:
> The screen width and differences in UI layouts between a Full Screen WinUI app, a narrow width window, and things like Hololens and Xbox - would not lend themselves to a docked Bar form factor.
> 
> So with that in mind, perhaps the InformationBar should be wrapped, along with an InformationCard control into an **AppMessage** API - so the same properties of Severity, Colour, Title, Message, Action Buttons, etc can be placed into the XAML, or into the app's code - and the way they are presented on screen changes to suit the form factor/device family.

These are great ideas. I'm going to bring this conversation over to the issue to get a wider audience and share some thoughts there.

robloo:
Now just getting to this thread. I probably should have stayed here for the entirety of the discussion.

> @robloo for your custom control it seems that a floating mode version of InfoBar wouldn't meet your use-cases in that your hints and contextual information are inline. I think you could use the InfoBar for your scenarios because it doesn't have to be a wide bar it will adjust based on your layout control or you could manually set its width. The visual design was focused on app-wide scenarios which the hints don't quite apply to as they are in-context with the current page/action. However, if you were satisfied with the visual aspects I don't see why you couldn't apply some custom styling (rounded corners, border color, etc.) to use InfoBar for your usecases.

@gabbybilka No, the current proposal even in V2 would not be ideal for inline, contextual hints. I'm not quite satisfied with the visual aspects but was content to discuss them closer to V2.

> So with that in mind, perhaps the InformationBar should be wrapped, along with an InformationCard control into an AppMessage API - so the same properties of Severity, Colour, Title, Message, Action Buttons, etc can be placed into the XAML, or into the app's code - and the way they are presented on screen changes to suit the form factor/device family.

This is exactly what I'm describing here: https://github.com/microsoft/microsoft-ui-xaml/issues/913#issuecomment-700307412 and https://github.com/microsoft/microsoft-ui-xaml/issues/913#issuecomment-700327225 so definitely has my full support!

I'll discuss the rest over on the other thread.

robloo:
@gabbybilka I posted a comparison and analysis of the ideas discussed here: https://github.com/microsoft/microsoft-ui-xaml/issues/913#issuecomment-700799061 since that thread is longer. It is relevant to the spec discussion as well.

I honestly would not go ahead with the InfoBar API as it exists today. There are too many differences to existing concepts and ideas apparent by comparing the TeachingTip and future ideas for controls.

mdtauk:
Is there any value in creating a ThemeResource which takes the Accent colour, and makes a faint Light and Dark version, which could be used for the InfoBar Background colour.

``| InfoBarAccentBackgroundBrush | Adjusted Light and Dark Accent color used as a background``

Neither the Light or Dark shades of the Accent colour provided, are faint enough to match the severity's contrast ratios

gabbybilka:
> Is there any value in creating a ThemeResource which takes the Accent colour, and makes a faint Light and Dark version, which could be used for the InfoBar Background colour.
> 
> `| InfoBarAccentBackgroundBrush | Adjusted Light and Dark Accent color used as a background`
> 
> Neither the Light or Dark shades of the Accent colour provided, are faint enough to match the severity's contrast ratios

@mdtauk Sorry for the late response here and thanks for your comment! Although interesting, I don't see the value in adding this functionality as the behavior doesn't quite match the intention of the InfoBar background colors or the Windows accent color in general. The system accent color is used to convey state information (e.g. item selection) or app branding ([Color in Windows docs](https://docs.microsoft.com/en-us/windows/uwp/design/style/color#accent-color)) while the background color in an InfoBar is to convey app status and follow the recommended associated color [meaning](https://docs.microsoft.com/en-us/windows/uwp/design/style/color#accent-color). As a user with a red system color, I would be confused seeing my accent color popping out on the page associated with non-error info. Is there a specific usage for this you anticipate?

mdtauk:
> > Is there any value in creating a ThemeResource which takes the Accent colour, and makes a faint Light and Dark version, which could be used for the InfoBar Background colour.
> > 
> > `| InfoBarAccentBackgroundBrush | Adjusted Light and Dark Accent color used as a background`
> > 
> > Neither the Light or Dark shades of the Accent colour provided, are faint enough to match the severity's contrast ratios
> 
> @mdtauk Sorry for the late response here and thanks for your comment! Although interesting, I don't see the value in adding this functionality as the behavior doesn't quite match the intention of the InfoBar background colors or the Windows accent color in general. The system accent color is used to convey state information (e.g. item selection) or app branding ([Color in Windows docs](https://docs.microsoft.com/en-us/windows/uwp/design/style/color#accent-color)) while the background color in an InfoBar is to convey app status and follow the recommended associated color [meaning](https://docs.microsoft.com/en-us/windows/uwp/design/style/color#accent-color). As a user with a red system color, I would be confused seeing my accent color popping out on the page associated with non-error info. Is there a specific usage for this you anticipate?
>
I was just pondering if it may be valuable to have. 

Some existing apps which display status banners use the accent colour, rather than a light or dark theme grey colour.  But these can be overridden by the developer, so it's good.

active/InformationBar/InformationBar.md
YuliKl: For NavigationView, we're toggling the IsPaneOpen property as soon as the action begins. So in both PaneClosing and PaneClosed events, we report IsPaneOpen as False; in both PaneOpening and PaneOpen we report IsPaneOpen as True.

YuliKl: Also, unless I'm doing something wrong, cancelling the event using NavigationViewPaneClosingEventArgs does nothing. The behavior of this
```c#
        private void NavView_PaneClosing(Microsoft.UI.Xaml.Controls.NavigationView sender, Microsoft.UI.Xaml.Controls.NavigationViewPaneClosingEventArgs args)
        {
            bool open = sender.IsPaneOpen;
            args.Cancel = true;
            open = sender.IsPaneOpen;
        }
```
is identical to the behavior of this
```c#
        private void NavView_PaneClosing(Microsoft.UI.Xaml.Controls.NavigationView sender, Microsoft.UI.Xaml.Controls.NavigationViewPaneClosingEventArgs args)
        {
            bool open = sender.IsPaneOpen;
        }
```

Felix-Dev: @YuliKI We have the open issue https://github.com/microsoft/microsoft-ui-xaml/issues/3352 which I plan to look into soon and where I could take a look at IsPaneOpen not being updated correctly when cancelling the close.

YuliKl: Thank you, @Felix-Dev, that would be great.

gabbybilka: Wrapping up this comment thread, yes, the behavior for InfoBar is the same as NavView. Once closing or opening starts, the isOpen property is set to false or true respectively.

gabbybilka: Thanks @YuliKl and @Felix-Dev for hopping in here. InfoBar and NavigationView _do_ follow the same pattern here that immediately when the action begins IsOpen is set to the intended value. See RaiseClosingEvent and RaiseClosedEvent in [InfoBar.cpp](https://github.com/microsoft/microsoft-ui-xaml/pull/3325/files?file-filters%5B%5D=.cpp&file-filters%5B%5D=.cs&file-filters%5B%5D=.h&file-filters%5B%5D=.projitems&file-filters%5B%5D=.shproj&file-filters%5B%5D=.targets&file-filters%5B%5D=.vcxproj&file-filters%5B%5D=.xaml#diff-624e3a295bc0485e93ffbc0dc6e381a4240ab7fc621f7600f3ad401ef3942963)

gabbybilka: @teaP ping to ensure I'm correct here 😊

active/InformationBar/InformationBar.md
gabbybilka: My initial comment a month ago wasn't clear, I added more text to clarify that the message should be when the app state changes.

MikeHillberg (approved):

PR #97, opened 2020-08-21
DispatcherQueue updates for WinUI3

MikeHillberg:


winui3/DispatcherQueueUpdates.md
```diff
@@ -0,0 +1,134 @@
...
+
+# Examples
+
+This example shows a thread ensuring it has a DispatcherQueue, setting a Synchronization
+context onto the thread that uses that queue, and then goes into a message pump.
+
+```cs
+static void Run()
+{
+    // Ensure we have a DispatcherQueue on this thread.
```
jeffstall, line 83: Please update comment:
// Note - this pattern is only necessary if you need to create a DispatcherQueue on a thread (ex as part
// of initialization).  For the vast majority of scenarios, a hosting framework has already created one
// on your behalf and so you should simply call DispatcherQueue.GetForCurrentThread() to retrieve it.

MikeHillberg:
@scottflute - you're right. I removed that part of the proposal and will open a separate Issue on it.

winui3/DispatcherQueueUpdates.md
```diff
@@ -0,0 +1,209 @@
...
+> In a WinUI you typically don't need to write this code for the initial UI thread, as it's
+automatically generated.
+
+```cs
+static void Run()
+{
+    // Ensure we have a DispatcherQueue on this thread. 
+    if(DispatcherQueue.GetForCurrentThread() == null)
+    {
+        DispatcherQueueController.CreateOnCurrentThread();
```
oldnewthing, line 125: Does anything interesting happen when the controller is GC'd? Or is it fine to just throw it away - the thread retains a reference to its own controller, but that reference is inaccessilble so you have no way to shut down the dispatcher queue. Your only choice is to rip the rug out from under it at thread termination.

winui3/DispatcherQueueUpdates.md
```diff
@@ -0,0 +1,209 @@
...
+    }
+
+    var syncContext = new DispatcherQueueSyncronizationContext();
```
david-risney, line 128: Typo: DispatcherQueueSync**h**ronizationContext

winui3/DispatcherQueueUpdates.md
MikeHillberg: @scott-moore-ms - look OK?

winui3/DispatcherQueueUpdates.md
MikeHillberg: @scott-moore-ms

scott-moore-ms: It should be fine to throw the controller away.  

By doing that, as you say, you lose the ability to fetch it again, but that is intentional in the design of splitting the controller from the queue - the entity that created the controller is the only one that has the ability to shut down the queue, whereas anyone on a thread should be able to access the queue to schedule their own work.  If after creating the controller, you decide you don't need to explicitly shut it down in the future (which is fairly common), then discarding it is totally fine.

winui3/DispatcherQueueUpdates.md
scott-moore-ms: LGTM

scott-moore-ms (approved):

jeffstall (commented):
1. We may need to change the namespace from Microsoft.System if we move DispatcherQueue.
2. It is still preferable to not have the "if not created, create it now" pattern. It's not good if components just start creating DispatcherQueues. This should almost always be created by the top-level object that created the thread, and it should know if the thread should have a DispatcherQueue.

jeffstall (approved):
(Marking approved to submit... we can continue to refine...)

PR #100, opened 2020-10-29
Expander Control

kat-y:
For [issue 3279](https://github.com/microsoft/microsoft-ui-xaml/issues/3279) 

Some expected changes to be added before this will be merged (not a complete list):

-  [ ] mockups and examples 
-  [ ] clarification on keyboarding and accessibility
-  [ ] details for theme resources

### DIFF
--- a/active/Expander/Expander.md
+++ b/active/Expander/Expander.md
@@ -1,0 +1,79 @@
+> See comments in Markdown for how to use this spec template
+
+<!-- The purpose of this spec is to describe a new feature and
+its APIs that make up a new feature in WinUI. -->
+
+<!-- There are two audiences for the spec. The first are people
+that want to evaluate and give feedback on the API, as part of
+the submission process.  When it's complete
+it will be incorporated into the public documentation at
+docs.microsoft.com (http://docs.microsoft.com/uwp/toolkits/winui/).
+Hopefully we'll be able to copy it mostly verbatim.
+So the second audience is everyone that reads there to learn how
+and why to use this API. -->
+
+# Background
+<!-- Use this section to provide background context for the new API(s) 
+in this spec. -->
+This spec corresponds to [this issue](https://github.com/microsoft/microsoft-ui-xaml/issues/3279) on the WinUI repo. 
+
+<!-- This section and the appendix are the only sections that likely
+do not get copied to docs.microsoft.com; they're just an aid to reading this spec. -->
+
+<!-- If you're modifying an existing API, included a link here to the
+existing page(s) -->
+
+<!-- For example, this section is a place to explain why you're adding this API rather than
+modifying an existing API. -->
+
+<!-- For example, this is a place to provide a brief explanation of some dependent
+area, just explanation enough to understand this new API, rather than telling
+the reader "go read 100 pages of background information posted at ...". -->
+
+
+# Description
+<!-- Use this section to provide a brief description of the feature.
+For an example, see the introduction to the PasswordBox control 
+(http://docs.microsoft.com/windows/uwp/design/controls-and-patterns/password-box). -->
+
+
+# Examples
+<!-- Use this section to explain the features of the API, showing
+example code with each description. The general format is: 
+  feature explanation,
+  example code
+  feature explanation,
+  example code
+  etc.-->
+  
+<!-- Code samples should be in C# and/or C++/WinRT -->
+
+<!-- As an example of this section, see the Examples section for the PasswordBox control 
+(https://docs.microsoft.com/windows/uwp/design/controls-and-patterns/password-box#examples). -->
+
+
+# Remarks
+<!-- Explanation and guidance that doesn't fit into the Examples section. -->
+
+<!-- APIs should only throw exceptions in exceptional conditions; basically,
+only when there's a bug in the caller, such as argument exception.  But if for some
+reason it's necessary for a caller to catch an exception from an API, call that
+out with an explanation either here or in the Examples -->
+
+# API Notes
+<!-- Option 1: Give a one or two line description of each API (type
+and member), or at least the ones that aren't obvious
+from their name.  These descriptions are what show up
+in IntelliSense. For properties, specify the default value of the property if it
+isn't the type's default (for example an int-typed property that doesn't default to zero.) -->
+
+<!-- Option 2: Put these descriptions in the below API Details section,
+with a "///" comment above the member or type. -->
+
+# API Details
+<!-- The exact API, in MIDL3 format (https://docs.microsoft.com/en-us/uwp/midl-3/) -->
+
+# Appendix
+<!-- Anything else that you want to write down for posterity, but 
+that isn't necessary to understand the purpose and usage of the API.
+For example, implementation details. -->
--- a/active/Expander/Expander.md
+++ b/active/Expander/Expander.md
@@ -1,5 +1,3 @@
-> See comments in Markdown for how to use this spec template
-
 <!-- The purpose of this spec is to describe a new feature and
 its APIs that make up a new feature in WinUI. -->
 
@@ -15,7 +13,6 @@ and why to use this API. -->
 # Background
 <!-- Use this section to provide background context for the new API(s) 
 in this spec. -->
-This spec corresponds to [this issue](https://github.com/microsoft/microsoft-ui-xaml/issues/3279) on the WinUI repo. 
 
 <!-- This section and the appendix are the only sections that likely
 do not get copied to docs.microsoft.com; they're just an aid to reading this spec. -->
@@ -29,13 +26,18 @@ modifying an existing API. -->
 <!-- For example, this is a place to provide a brief explanation of some dependent
 area, just explanation enough to understand this new API, rather than telling
 the reader "go read 100 pages of background information posted at ...". -->
+ > This spec corresponds to [this issue](https://github.com/microsoft/microsoft-ui-xaml/issues/3279) on the WinUI repo. 
 
+Throughout Windows, different expander controls are used by various apps and surfaces. There’s currently no consistent way to address this common UX pattern. This control is needed for situations where expanding (instead of overlaying) content is needed.  An Expander control is motivated by its use in many app scenarios and supporting developers in migrating from WPF and WCT.  
 
 # Description
 <!-- Use this section to provide a brief description of the feature.
 For an example, see the introduction to the PasswordBox control 
 (http://docs.microsoft.com/windows/uwp/design/controls-and-patterns/password-box). -->
+An Expander control is a UI component that provides a standard interaction for pushing adjacent content while expanding and collapsing. An Expander is independent of the contents inside it, including controls. 
 
+## Is this the right control?
+An Expander is not intended for flyout-type behaviors where a UI component overlays on top of adjacent content. In other words, Expander will 'push' and not overlay or fly out.
 
 # Examples
 <!-- Use this section to explain the features of the API, showing
@@ -70,10 +72,60 @@ isn't the type's default (for example an int-typed property that doesn't default
 <!-- Option 2: Put these descriptions in the below API Details section,
 with a "///" comment above the member or type. -->
 
+
 # API Details
 <!-- The exact API, in MIDL3 format (https://docs.microsoft.com/en-us/uwp/midl-3/) -->
+~~~~
+Public enum ExpandDirection 
+{ 
+Down = 0 
+Up = 1 
+} 
+ 
+public class Expander : ContentControl 
+{ 
+Expander(); 
+ 
+public object Header {get;set;} 
+public DataTemplate HeaderTemplate { get;set; } 
+public DataTemplate HeaderTemplateSelector {get;set;} 
+ 
+public static readonly DependencyProperty HeaderProperty; 
+public static readonly DependencyProperty HeaderTemplateProperty; 
+public static readonly DependencyProperty HeaderTemplateSelectorProperty {get;} 
+ 
+public bool IsExpanded { get;set} 
+public ExpandDirection ExpandDirection { get;set;} 
+ 
+protected virtual void OnExpanded(); 
+protected virtual void OnCollapsed(); 
+ 
+public event Expanded; 
+public event Collapsed; 
+ 
+public static readonly DependencyProperty IsExpandedProperty; 
+public static readonly DependencyProperty ExpandDirectionProperty; 
+} 
+~~~~
+
 
 # Appendix
 <!-- Anything else that you want to write down for posterity, but 
 that isn't necessary to understand the purpose and usage of the API.
 For example, implementation details. -->
+
+# Inputs and Accessibility
+## UI Automation Patterns
+Expander will use a ExpandCollapsePattern.
+
+## Keyboard Navigation
+TBD
+
+# Open Question
+## General
+* Should this be named "Expander" or "ExpanderView"?
+
+## Input and Accessibility
+* Should the first tab-stop be the entire Expander or just the Header? WCT Expander has the Header as the first tab-stop.
+* Should the second tab-stop be the chevron/expansion icon? WPF has the first tab-stop as the Header, the second as just the chevron. 
+


active/Expander/Expander.md
```diff
@@ -0,0 +1,131 @@
...
+} 
+~~~~
+
+
+# Appendix
+<!-- Anything else that you want to write down for posterity, but 
+that isn't necessary to understand the purpose and usage of the API.
+For example, implementation details. -->
+
+# Inputs and Accessibility
```
chingucoding, line 117: Expanding/Collapsing the expander should raise [RaisePropertyChanged](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.automation.peers.automationpeer.raisepropertychangedevent?view=winrt-19041#Windows_UI_Xaml_Automation_Peers_AutomationPeer_RaisePropertyChangedEvent_Windows_UI_Xaml_Automation_AutomationProperty_System_Object_System_Object_) with the property changed being the [ExpandCollapseProperty](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.automation.expandcollapsepatternidentifiers.expandcollapsestateproperty?view=winrt-19041#Windows_UI_Xaml_Automation_ExpandCollapsePatternIdentifiers_ExpandCollapseStateProperty) property.

active/Expander/Expander.md
```diff
@@ -0,0 +1,131 @@
...
+
+# Remarks
+<!-- Explanation and guidance that doesn't fit into the Examples section. -->
+
+<!-- APIs should only throw exceptions in exceptional conditions; basically,
+only when there's a bug in the caller, such as argument exception.  But if for some
+reason it's necessary for a caller to catch an exception from an API, call that
+out with an explanation either here or in the Examples -->
+
+# API Notes
```
chingucoding, line 65: I think we should determine which properties of [ContentControl](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.controls.contentcontrol?view=winrt-19041) propagate into the control and in which way, e.g. how does foreground or background get propagated down to the content.

active/Expander/Expander.md
```diff
@@ -0,0 +1,131 @@
...
+## Keyboard Navigation
+TBD
+
+# Open Question
+## General
+* Should this be named "Expander" or "ExpanderView"?
+
+## Input and Accessibility
+* Should the first tab-stop be the entire Expander or just the Header? WCT Expander has the Header as the first tab-stop.
+* Should the second tab-stop be the chevron/expansion icon? WPF has the first tab-stop as the Header, the second as just the chevron.
```
chingucoding, line 130: I don't think we should match the WCT expander here, when navigating the expander with narrator, the header content does not get announced. I think having the icon be a separate tab stop is the better idea.

kat-y: Other feedback I've heard is that the if the only action that users can take on the Header is to expand/collapse, the icon shouldn't be a separate tab-stop. This gets messy when there's tab-stop content inside the header, though.

active/Expander/Expander.md
kat-y: Makes sense, thanks for pointing it out! 😄

active/Expander/Expander.md
```diff
@@ -0,0 +1,166 @@
...
+<!-- Anything else that you want to write down for posterity, but 
+that isn't necessary to understand the purpose and usage of the API.
+For example, implementation details. -->
+
+# Inputs and Accessibility
+## UI Automation Patterns
+Expander will use a ExpandCollapsePattern. Expanding/Collapsing the expander will raise [RaisePropertyChanged](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.automation.peers.automationpeer.raisepropertychangedevent?view=winrt-19041) with the property changed being the [ExpandCollapseProperty](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.automation.expandcollapsepatternidentifiers.expandcollapsestateproperty?view=winrt-19041) property.
+
+## Keyboard
+* Only the header takes focus (not the entire Expander, or the chevron)
```
mdtauk, line 153: What about scenarios when the content includes controls like the Button or Hyperlink?

kat-y: Since the header takes focus and not the chevron, after the entire header is focused, content inside the header takes focus (you can see this behavior in the Expander today 😄 https://github.com/microsoft/microsoft-ui-xaml/tree/master/dev/Expander )

active/Expander/Expander.md
```diff
@@ -0,0 +1,166 @@
...
+*	When focus is on the Header - hit space/enter/right key to expand it, and focus does not move
+* Space and enter should expand and collapse 
+* right arrow will expand, left arrow will collapse
+
+## GamePad
+TBD
```
ranjeshj, line 159: @kat-y can you fill this out ?

active/Expander/Expander.md
```diff
@@ -0,0 +1,166 @@
...
+
+# Open Questions
+
+## Input and Accessibility
+* Is there a RTL consideration for right/left keyboarding behavior?
```
ranjeshj, line 164: This isn't navigating, but expanding and collapsing, so I don't think we would want to change behavior for left right based on RTL.

active/Expander/Expander.md
```diff
@@ -0,0 +1,166 @@
...
+# API Details
+<!-- The exact API, in MIDL3 format (https://docs.microsoft.com/en-us/uwp/midl-3/) -->
+~~~~
+Public enum ExpandDirection 
+{ 
+ Down = 0 
+ Up = 1 
+} 
+ 
+public class Expander : ContentControl
```
ranjeshj, line 105: @kat-y The idl is out of date. please see https://github.com/microsoft/microsoft-ui-xaml/blob/master/dev/Expander/Expander.xaml

active/Expander/Expander.md
```diff
@@ -0,0 +1,166 @@
...
+
+<!-- For example, this section is a place to explain why you're adding this API rather than
+modifying an existing API. -->
+
+<!-- For example, this is a place to provide a brief explanation of some dependent
+area, just explanation enough to understand this new API, rather than telling
+the reader "go read 100 pages of background information posted at ...". -->
+ > This spec corresponds to [this issue](https://github.com/microsoft/microsoft-ui-xaml/issues/3279) on the WinUI repo. 
+
+Throughout Windows, different expander controls are used by various apps and surfaces. There’s currently no consistent way to address this common UX pattern. This control is needed for situations where expanding (instead of overlaying) content is needed.  An Expander control is motivated by its use in many app scenarios and supporting developers in migrating from WPF and WCT.
```
MikeHillberg, line 31: Would help to explain what an Expander is (show a sample picture). And add a link to WPF's [Expander](https://docs.microsoft.com/dotnet/api/System.Windows.Controls.Expander).

active/Expander/Expander.md
```diff
@@ -0,0 +1,166 @@
...
+  feature explanation,
+  example code
+  etc.-->
+  
+<!-- Code samples should be in C# and/or C++/WinRT -->
+
+<!-- As an example of this section, see the Examples section for the PasswordBox control 
+(https://docs.microsoft.com/windows/uwp/design/controls-and-patterns/password-box#examples). -->
+
+## Create an Expander
```
MikeHillberg, line 56: Need a picture of what this looks like.

active/Expander/Expander.md
MikeHillberg: Very good point. Specifically to the Header property (the Content property is inherited from ContentControl and not affected by Expander).

kat-y: The background and foreground of ContentControl do not propagate into the Header or Content. Background color set with ExpanderHeaderBackground is overwritten by whatever is inside the Header. How should these properties propagate from ContentControl to Expander's Content?

active/Expander/Expander.md
ranjeshj: You can probably get some pictures from the PR where the styles are being updated - https://github.com/microsoft/microsoft-ui-xaml/pull/4024

active/Expander/Expander.md
```diff
@@ -0,0 +1,173 @@
...
+
+# Description
+<!-- Use this section to provide a brief description of the feature.
+For an example, see the introduction to the PasswordBox control 
+(http://docs.microsoft.com/windows/uwp/design/controls-and-patterns/password-box). -->
+An Expander control is a UI component that provides a standard interaction for showing more content in a container that pushes adjacent content while expanding and collapsing. An Expander cannot be light dismissed and is independent of the contents inside it, including controls. Expander should be used when some content is only relevant some of the time (for example to read more information or access additional options for an item). 
+
+![A collapsed Expander](images/expander_collapsed.PNG)
```
MikeHillberg, line 39: I see no header or content in these examples, is that intentional?

active/Expander/Expander.md
MikeHillberg: The ContentControl.Foreground property does inherit (but not Background). There's other properties too, like FontSize. I think the answer is that properties only inherit to Content, but we should call it out in the description/remarks.

active/Expander/Expander.md
```diff
@@ -0,0 +1,173 @@
...
+~~~~
+<muxc:Expander x:Name="SimpleExpander" />
+~~~~
+## Put controls inside an Expander
+![An expanded Expander with a ToggleButton in the Header and a Button in the Content](images/expander_example_buttons.PNG)
+XAML
+~~~~
+<muxc:Expander x:Name="Expander2" IsExpanded="False" ExpandDirection ="Down">
+    <muxc:Expander.Header>
+        <ToggleButton>This is a ToggleButton in the header</ToggleButton>
```
MikeHillberg, line 68: Do I have to put the ToggleButton in the Header myself? I was expecting I could put any content in there (typically a string), and the toggling behavior would be automatic.

active/Expander/Expander.md
```diff
@@ -0,0 +1,173 @@
...
+
+## Keyboard
+* Tabbing brings focus to the Header, neither the entire Expander nor the chevron is a tab-stop
+* When focus is on the Header, space key expands and collapse and focus does not move
+* Keyboard navigation inside the Expander's Header/Content is based on the content inside them
+
+## GamePad
+The Expander can be expanded and collapsed with A. Spatial navigation will navigate between the content of the Expander. 
+
+## Narrator
```
MikeHillberg, line 167: Just Narrator or any screen reader?

active/Expander/Expander.md
ranjeshj: That is correct, you should not have to put in a toggle button.

active/Expander/Expander.md
kat-y: I'm updating the examples (including putting text in this first one to make it clearer).

active/Expander/Expander.md
kat-y: I'll update this area for any screen reader

active/Expander/Expander.md
kat-y: I'm updating the examples to clarify this - sorry for the confusion! The intention was to show that you can put another control (like a ToggleButton) inside the Header.

active/Expander/Expander.md
```diff
@@ -0,0 +1,155 @@
...
+[lightweight styling guide](https://docs.microsoft.com/en-us/windows/uwp/design/controls-and-patterns/xaml-styles#lightweight-styling).
+
+| Name| Description | 
+| :---------- | :------- | 
+| ExpanderChevronMargin | Chevron margin thickness| 
+| ExpanderChevronGlyph | Chevron glyph| 
+| ExpanderChevronSize | Chevron size|
+| ExpanderPopinVerticalOffset | vertical offset for animation| 
+
+Some theme resources TBD.
```
MikeHillberg, line 68: ?

active/Expander/Expander.md
```diff
@@ -0,0 +1,155 @@
...
+
+| Name | Description | Default |
+| :---------- | :------- | :------- |
+| ExpandDirection | Sets the direction of expansion | Down = 0 |
+| IsExpanded | Whether or not control is expanded | False |
+
+### Events
+| Name | Description | 
+| :---------- | :------- | 
+| Expanding | Occurs when expanded |
```
MikeHillberg, line 84: Should be: occurs when expand- _ing_. Since this is an ing event, it is raised before the content is shown, and so you could use this to delay-load the content?

active/Expander/Expander.md
```diff
@@ -0,0 +1,155 @@
...
+    Content="This is in the content"/>
+~~~~
+
+#### Put controls inside an Expander
+
+![A collapsed Expander. The Header has the text "This is in the header" and the Content has a button with the text "This ToggleButton is in the Content". The Expander is expanded, the button is toggled on and off, and the Expander is collapsed. ](images/Expander_togglebutton.gif)
+
+XAML
+~~~~
+<muxc:Expander x:Name="Expander2" Header="This is in the header">
```
oldnewthing, line 48: Presumably, I could also put content in the Header?

```xml
<muxc:Expander x:Name="Expander2">
    <muxc:Expander.Header>
        <Button>Button in the header</Button>
    </muxc:Expander.Header>
    <Button>Button in the content</Button>
</muxc:Expander>
```

active/Expander/Expander.md
```diff
@@ -0,0 +1,155 @@
...
+You can modify the look of an Expander by specifying Xaml resources in your app.
+For more info, see the
+[lightweight styling guide](https://docs.microsoft.com/en-us/windows/uwp/design/controls-and-patterns/xaml-styles#lightweight-styling).
+
+| Name| Description | 
+| :---------- | :------- | 
+| ExpanderChevronMargin | Chevron margin thickness| 
+| ExpanderChevronGlyph | Chevron glyph| 
+| ExpanderChevronSize | Chevron size|
+| ExpanderPopinVerticalOffset | vertical offset for animation|
```
oldnewthing, line 66: How do I style the background for the header and content separately? Which one does the `Background` property apply to? Similarly for font properties and `Foreground`.

Is this where `HeaderTemplate` enters the picture?

Or is this a case where you have to edit the template, and we expect most people to accept the default background and font?

active/Expander/Expander.md
```diff
@@ -0,0 +1,155 @@
...
+
+
+## Expander members
+
+### Properties
+
+| Name | Description | Default |
+| :---------- | :------- | :------- |
+| ExpandDirection | Sets the direction of expansion | Down = 0 |
+| IsExpanded | Whether or not control is expanded | False |
```
oldnewthing, line 79: Missing Header, HeaderTemplate, HeaderTemplateSelector.

active/Expander/Expander.md
```diff
@@ -0,0 +1,155 @@
...
+## Keyboard
+* Tabbing brings focus to the Header, neither the entire Expander nor the chevron is a tab-stop
+* When focus is on the Header, space key expands and collapse and focus does not move
+* Keyboard navigation inside the Expander's Header/Content is based on the content inside them
+
+## GamePad
+The Expander can be expanded and collapsed with A. Spatial navigation will navigate between the content of the Expander. 
+
+## Screen reader
+When focus is on the Header of an Expander, the screen reader will announce "Expander collapsed" or "Expander expanded". With focus staying on the header, the screen reader will then anounce "expanded" or "collapsed" accordingly. Expander should not have [focus engagement](https://docs.microsoft.com/en-us/windows/uwp/design/input/gamepad-and-remote-interactions#focus-engagement) enabled.
```
oldnewthing, line 147: Does the screen reader also read the header? Is the header read first, and then "Expander collapsed/expanded" comes after?

active/Expander/Expander.md
```diff
@@ -0,0 +1,155 @@
...
+
+# Inputs and Accessibility
+
+## UI Automation Patterns
+Expander will use a ExpandCollapsePattern. Expanding/Collapsing the expander will raise [RaisePropertyChanged](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.automation.peers.automationpeer.raisepropertychangedevent?view=winrt-19041) with the property changed being the [ExpandCollapseProperty](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.automation.expandcollapsepatternidentifiers.expandcollapsestateproperty?view=winrt-19041) property.
+
+## Keyboard
+* Tabbing brings focus to the Header, neither the entire Expander nor the chevron is a tab-stop
+* When focus is on the Header, space key expands and collapse and focus does not move
+* Keyboard navigation inside the Expander's Header/Content is based on the content inside them
```
oldnewthing, line 141: If the Header has focusable content, how does the interaction model work? (Maybe there's a standard XAML answer for this?)

active/Expander/Expander.md
ejserna: Yes

active/Expander/Expander.md
ejserna: I think it's first the header content, then the collapsed/expanded property.

active/Expander/Expander.md
ejserna: I think the HeaderTemplate can be used to style the header, but @alexhinz knows more about this than me.

active/Expander/Expander.md
alexhinz: The part of Expander that holds Header content is the first tab stop, then the Header's actual contents if focusable (e.g. a Button), then the Content's actual contents if focusable.

active/Expander/Expander.md
alexhinz: `Header`, `HeaderTemplate`, and `HeaderTemplateSelector` are functionally analogous to `Content`, `ContentTemplate`, and `ContentTemplateSelector`, and they apply to the area of Expander that's to the left of the chevron. So because they don't span the entire border-to-border (inclusive) area of the header, they're likely not a good fit for that kind of styling. They also would not be aware of any Expander states, like if it's expanded, if a dev would like to change colors based on them.

So devs could retemplate to style the header and background areas separately. Adding in additional lightweight styling properties would help avoid the need to retemplate.

active/Expander/Expander.md
```diff
@@ -0,0 +1,155 @@
...
+| :---------- | :------- | 
+| Expanding | Occurs when expanded |
+| Collapsed| Occurs when collapsed |
+
+
+
+# API Details
+
+~~~~
+enum ExpandDirection
```
jonwis, line 92: `ExpanderExpandingDirection` ? Two-word names are High Value.

active/Expander/Expander.md
```diff
@@ -0,0 +1,155 @@
...
+## UI Automation Patterns
+Expander will use a ExpandCollapsePattern. Expanding/Collapsing the expander will raise [RaisePropertyChanged](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.automation.peers.automationpeer.raisepropertychangedevent?view=winrt-19041) with the property changed being the [ExpandCollapseProperty](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.automation.expandcollapsepatternidentifiers.expandcollapsestateproperty?view=winrt-19041) property.
+
+## Keyboard
+* Tabbing brings focus to the Header, neither the entire Expander nor the chevron is a tab-stop
+* When focus is on the Header, space key expands and collapse and focus does not move
+* Keyboard navigation inside the Expander's Header/Content is based on the content inside them
+
+## GamePad
+The Expander can be expanded and collapsed with A. Spatial navigation will navigate between the content of the Expander.
```
jonwis, line 144: ... "with the A button." ?

active/Expander/Expander.md
```diff
@@ -0,0 +1,155 @@
...
+
+    static Windows.UI.Xaml.DependencyProperty IsExpandedProperty{ get; };
+    static Windows.UI.Xaml.DependencyProperty ExpandDirectionProperty{ get; };
+}
+~~~~
+
+# Inputs and Accessibility
+
+## UI Automation Patterns
+Expander will use a ExpandCollapsePattern. Expanding/Collapsing the expander will raise [RaisePropertyChanged](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.automation.peers.automationpeer.raisepropertychangedevent?view=winrt-19041) with the property changed being the [ExpandCollapseProperty](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.automation.expandcollapsepatternidentifiers.expandcollapsestateproperty?view=winrt-19041) property.
```
jonwis, line 136: (Low-pri) There was an ask in other documentation to remove the `en-us` tag so it lands on the clicker's preferred language.

active/Expander/Expander.md
kat-y: To clarify this, I'll be updating the examples so the second one has a ToggleButton in the header as well!

active/Expander/Expander.md
oldnewthing: I think lightweight styling properties would be popular.

* To convey state. E.g. "Tuesday has slots available, Wednesday is full, and Thursday is oversubscribed."
* To color-code the Expanders. E.g., Errors in red, Warnings in yellow.
* Everybody wants their colors to match their brand.

active/Expander/Expander.md
kat-y: By this I mean putting a ToggleButton inside the Header as part of an example showing controls in the Header and Content.

active/Expander/Expander.md
kat-y: Re @MikeHillberg question: Following the latest Expander PR merge the Background property now does propagates to the Header (while the Foreground, Fontsize, and FontFamily propagate to the Content). As an example, the following code results in an Expander where the entire header area (including 'under' the chevron) is Cyan, with default text styling, but the content area has default background color with text that is pink and Times New Roman font. (If there's a way to add pictures to a PR thread let me know!)

<controls:Expander x:Name="Expander1" Header="This is in the header"  Content="This is in the content" FontSize ="25" FontFamily="Times New Roman" Foreground="Magenta" Background="Cyan"/>

active/Expander/Expander.md
kat-y: I'll add this as an open question, thanks for the note!

active/Expander/Expander.md
kat-y: Thanks for the note! Making the changes

active/Expander/Expander.md
kat-y: I'll add some clarification to this section

active/Expander/Expander.md
MikeHillberg: Some other interactive control would help. A toggle within a toggle creates some confusion.

MikeHillberg: Recommend: Just some interactive content other than a ToggleButton.

active/Expander/Expander.md
MikeHillberg: Discussion: You could set the background of the Header using the Background property, and the background of the Content within the content itself.

Recommend: Maybe show an example?

active/Expander/Expander.md
MikeHillberg: Recommend: Add the rest of the theme resources (brushes etc)

active/Expander/Expander.md
MikeHillberg: Discussion: These names were chosen in order to align with WPF.

active/Expander/Expander.md
MikeHillberg: Recommend: add to description

active/Expander/Expander.md
MikeHillberg: Discussion: The inheritable properties _should_ inherit to both Header and Content, and likely already do in the implementation. Note that frequently controls have a default style though that sets inheritable properties as well.

Recommend: investigate and follow up, but general principle should be that inheritable properties behave the same for the Header and the Content.

active/Expander/Expander.md
chingucoding: Maybe a TextBox or RichEditBox (a scenario might be "Enter your password" with the dropdown showing a "Forgot password button" or something like that)?

oldnewthing: image

chingucoding: Great idea @oldnewthing, that's a very good example; @kat-y I think we should use that.

active/Expander/Expander.md
kat-y: Added this to the UI Automation Patterns section

active/Expander/Expander.md
MikeHillberg: The inheritable properties such as FontSize generally don't inherit to the Header because the Header is in a ToggleButton, and as said above it has a default style that sets most/all of the inheritable properties, and that takes precedence. DataContext does inherit though.

Jay-o-Way:
Is this still an issue?

PR #102, opened 2020-11-23
Breadcrumb Control

kat-y:
For [issue 3326](https://github.com/microsoft/microsoft-ui-xaml/issues/3326) 

Some expected changes to be added before this will be merged


mdtauk:
As to the open questions.  I think from a visual point of view, there should be no difference in sizing between Chevrons that can be tapped, and those that can't.  The sizing should meet the touch target sizing for useability.  However I think there should be PointerOver and Pressed states for the Chevrons and Items.

Would the last BreadcrumbItem in the control also have a chevron?
Should it include Lightweight Styles so it can be styled differently?
Is there a need to have a "current"/selected property for the active breadcrumb item?  
Does it always need to be the final item in the control, or could it support jumping back a level, without losing the full trail?

active/Breadcrumb/Breadcrumb.md
```diff
@@ -0,0 +1,184 @@
...
+* In V2 Breadcrumb, should the chevrons' interactability be indicated through design? Would the node-chevron spacing be the same or different for enabled vs disabled chevrons? 
+
+## Input and Accessibility
+* How should Breadcrumb's spacing/touch targets change for keyboard/mouse versus a touchscreen? 
+
+### Keyboard
+* the enter/space keys to jump to the current node?
+* Should the down arrow (or alt-down) drop down the children of the current node? Potentially also space and enter to drop down the children? the escape key to close the drop down?
+* the up/down arrows to navigate from child to child in the drop down?
+* File Exploreer exhibits wrapping arrowing behavior - pressing right arrow on the rightmost node brings you back to the root. Is this a behavior WinUI Breadcrumb should enable?
```
duke7553, line 179:
```suggestion
* File Explorer exhibits wrapping arrowing behavior - pressing right arrow on the rightmost node brings you back to the root. Is this a behavior WinUI Breadcrumb should enable?
```

mdtauk:
Microsoft Teams has a Breadcrumb control

image

image
They use a drop down to show content that is not able to fit on the bar itself

image
They use the ellipsis as a button to indicate more items

image
The breadcrumb bar includes a space for buttons/actions

----

I am not sure how you have both the more overflow, and the child item lists as flyouts - without possible confusion.  That is something that will need a design approach for.

Material Design has something which may help
image

The separators are not clickable, but the item with a drop down list has a downward chevron.

mdtauk:
image

active/Breadcrumb/Breadcrumb.md
```diff
@@ -0,0 +1,184 @@
...
+
+## V1 API
+~~~~
+    public class Node 
+    { 
+        public string Label; 
+    } 
+App’s ViewModel: 
+
+    public IEnumerable<Node> Nodes { get; }
```
MikeHillberg, line 107: This looks like sample code, we should only have public API here.

active/Breadcrumb/Breadcrumb.md
```diff
@@ -0,0 +1,184 @@
...
+    { 
+        public BreadcrumbItem(); 
+    } 
+
+    // Argument for the ItemClicked event. 
+    public sealed class BreadcrumbItemClickedEventArgs 
+    { 
+    // Returns the clicked item. It is either sender.CurrentItem  
+    // or an item in the Children list accessed for the dropdown of sender.CurrentItem. 
+        public object Item { get; }
```
MikeHillberg, line 138: The item might be ambiguous. For example if you're showing the path `\A\B\A`. We should include an index in here too.

active/Breadcrumb/Breadcrumb.md
```diff
@@ -0,0 +1,150 @@
...
+
+<!-- As an example of this section, see the Examples section for the PasswordBox control 
+(https://docs.microsoft.com/windows/uwp/design/controls-and-patterns/password-box#examples). -->
+## Create a Breadcrumb
+
+![Breadcrumb default with 3 nodes](images/Breadcrumb_default.PNG)
+ ~~~~ 
+     <muxc: Breadcrumb ItemsSource = {"Node", "Node", "Node"} /> 
+~~~~
+More examples to be added
```
MikeHillberg, line 65: ?

active/Breadcrumb/Breadcrumb.md
```diff
@@ -0,0 +1,150 @@
...
+## UI Automation Pattern
+Breadcrumb will use the [InvokePattern](!https://docs.microsoft.com/en-us/dotnet/framework/ui-automation/implementing-the-ui-automation-invoke-control-pattern).
+
+## Keyboard
+* The user can tab into the Breadcrumb and tab out. The first tab in gives focus to the first (highest-level) node. 
+Once the Breadcrumb has focus, the user can use the left/right arrows to navigate from node to node. 
+* If the Breadcrumb is crumbled, down/up will open the flyout from the ellipsis and down/up will navigate between the flyout items
+
+## Gamepad
+* Spatial navigation to focus and select the nodes.
```
MikeHillberg, line 147: What does this mean?

active/Breadcrumb/Breadcrumb.md
```diff
@@ -0,0 +1,150 @@
...
+
+## Narrator
```
MikeHillberg, line 149: Is there code specific to Narrator, or is this discussing any screen reader?

active/Breadcrumb/Breadcrumb.md
```diff
@@ -0,0 +1,150 @@
...
+    BreadcrumbLayout();
+}
+
+[MUX_PROPERTY_CHANGED_CALLBACK(TRUE)]
+[MUX_PROPERTY_CHANGED_CALLBACK_METHODNAME("OnPropertyChanged")]
+unsealed runtimeclass Breadcrumb : Windows.UI.Xaml.Controls.Control
+{
+    Breadcrumb();
+
+    Object DropdownItemTemplate{ get; set; };
```
MikeHillberg, line 119: "Drop down" is two words, so this should capitalize the second 'd': `DropDownItemTemplate`.

active/Breadcrumb/Breadcrumb.md
```diff
@@ -0,0 +1,150 @@
...
+{
+    Object Item { get; };
+}
+
+unsealed runtimeclass BreadcrumbItem : Windows.UI.Xaml.Controls.ContentControl
+{
+    BreadcrumbItem();
+}
+
+unsealed runtimeclass BreadcrumbLayout : NonVirtualizingLayout
```
MikeHillberg, line 108: Why is this here? How does an app use it?

active/Breadcrumb/Breadcrumb.md
MikeHillberg, line 119: What is this for and how is it used?

active/Breadcrumb/Breadcrumb.md
```diff
@@ -0,0 +1,120 @@
...
+This control is similar to an ItemsControl (has an ItemsSource property), but does not derive from it.
+
+## V1/V2 Breadcrumb
+
+The work for Breadcrumb has been scoped to V1 and V2 stages of the API.
+In V1 Breadcrumb no chevrons are interactable to show children.
+V2 Breadcrumb will add functionality to have flyouts from chevrons for every node,
+to view children of that node. This spec focuses on V1 Breadcrumb,
+but keeping in mind that V1 is positioned in a way to have V2 features smoothly added later. 
+
```
gabbybilka, line 18: I'd like to see a couple of sections to provide an overview of the Breadcrumb functionality and usage recommendations:
- Description (with "Is this the right control?")
- Examples: This could merely expand the current "Breadcrumb class" section into separate entries with a summary, code example, and image.

active/Breadcrumb/Breadcrumb.md
```diff
@@ -0,0 +1,120 @@
...
+# API Pages
+
+## Breadcrumb class
+
+Use the Breadcrumb control to private persisting 'trail' that shows the user's navigation path.
```
gabbybilka, line 23: I'm not sure what *private* persisting 'trail' means. Could you clarify?

active/Breadcrumb/Breadcrumb.md
```diff
@@ -0,0 +1,120 @@
...
+
+This is not the 'history' of the user's navigation, but the direct path of pages or folders from
+the root node to their current position. For situations where resizing means the full path can't be shown,
+the Breadcrumb will 'crumble' starting at the root node. In addition,
```
gabbybilka, line 27: Could you define 'crumble' here?
> For situations where resizing means the full path can't be shown, the Breadcrumb control will replace the item text with an ellipsis. This will prioritize the content after the first ellipsis...?
I think my confusion comes with not understanding when exactly the control will crumble or not crumble item text.

active/Breadcrumb/Breadcrumb.md
```diff
@@ -0,0 +1,120 @@
...
+
+A Breadcrumb control with the last node truncated:
+
+![Breadcrumb_crumbled with last node truncated](images/Breadcrumb_truncation.PNG)
+
+The following Xaml markup  creates a simple Breadcrumbe control with three nodes:
+
+![Breadcrumb default with 3 nodes](images/Breadcrumb_default.PNG)
+ ~~~~ 
+     <muxc: Breadcrumb ItemsSource = {"Node", "Node", "Node"} />
```
gabbybilka, line 46: Could you recommend a format for the ItemsSource as it is currently just an Object?

active/Breadcrumb/Breadcrumb.md
```diff
@@ -0,0 +1,120 @@
...
+
+A Breadcrumb "crumbling" with 3 visible nodes:
+
+![Breadcrumb crumbling with 3 visible nodes](images/Breadcrumb_crumbling.PNG)
+
+A Breadcrumb control with the last node truncated:
+
+![Breadcrumb_crumbled with last node truncated](images/Breadcrumb_truncation.PNG)
+
+The following Xaml markup  creates a simple Breadcrumbe control with three nodes:
```
gabbybilka, line 42: nit: typo on Breadcrumbe

active/Breadcrumb/Breadcrumb.md
```diff
@@ -0,0 +1,120 @@
...
+the ellipsis and down/up will navigate between the flyout items
+
+## Gamepad
+
+* Spatial navigation to focus and select the nodes.
+
+## Narrator
+
+When the focus is on a node, Narrator will announce the node
+as well as "n of m" to convey which node in the path it is.
```
almedina-ms, line 118: Would n and m be relative to the number of nodes currently visible or the total number of nodes? for example, if we have a breadcrumb with 5 items:
... > Node 4 > Node 5
where Node 1, Node 2 and Node 3 are 'crumbled', would Node 4 be announced as "4 of 5" or would it be "2 of 3". In the case of the crumbled nodes, would they also be enumerated from 1 to 3?

active/Breadcrumb/Breadcrumb.md
```diff
@@ -0,0 +1,120 @@
...
+* If the Breadcrumb is crumbled, down/up will open the flyout from
+the ellipsis and down/up will navigate between the flyout items
+
+## Gamepad
+
+* Spatial navigation to focus and select the nodes.
+
+## Narrator
+
+When the focus is on a node, Narrator will announce the node
```
almedina-ms, line 117: How would the ellipsis button be read as? Should it state "Collapsed nodes" or some key phrase?

active/Breadcrumb/Breadcrumb.md
MikeHillberg: Great callout. I would expect 4 of 5, but only if keyboarding will scroll through the list. 

Precedence is probably more interesting. But for example in Explorer the breadcrumb isn't actually announced as a list at all.

active/Breadcrumb/Breadcrumb.md
mdtauk: Overflow, "More Breadcrumbs" - the ellipsis icon is used for **More** in the design language

active/Breadcrumb/Breadcrumb.md
ranjeshj: This is a primitive used by the control, the app does not need to know about this layout.

active/Breadcrumb/Breadcrumb.md
ranjeshj: The item would be the data at that index. Providing the index sounds like a good idea to disambiguate.

active/Breadcrumb/Breadcrumb.md
ranjeshj: The ellipses button when clicked shows a flyout with all the crumbled nodes. This is the template to be used for each item in that flyout.

MikeHillberg: I see. We should explain that above. So in the future there might also be a dropdown on each visible button? Do we need to distinguish the names?

active/Breadcrumb/Breadcrumb.md
MikeHillberg: So we could remove this API (doesn't need to be public)?

ranjeshj: We could make it internal and set it from code.

active/Breadcrumb/Breadcrumb.md
ranjeshj: I'm not sure if there is a scenario where we need to distinguish between the two. @kat-y ?

kat-y: Is "Drop down" treated as two words elsewhere? Drop-down is the adjective used here, which is [either hyphenated or one word](https://www.merriam-webster.com/dictionary/drop-down).

kat-y: I don't see a scenario where the flyout of crumbled nodes should be distinct from the (in the future) flyout from chevrons to show children nodes. While the hierarchy structure is different, they are both flyouts from Breadcrumb and consistency here is good.

chingucoding: The DropDownButton control is also written with "Down" capitalized so it would make sense to do that here too. That capitilzation is also used in multiple different APIs and lightweight styling resource names.

almedina-ms: Telerik also uses [DropDown]( https://docs.telerik.com/devtools/wpf/controls/radmulticolumncombobox/dropdown-properties#height-of-the-dropdown) so for consistency I'd say we could go with DropDown, but there are other cases in css where the properties are capitalized as Dropdown in the titles but the property is written as dropdown in code, so that would seem inconclusive

active/Breadcrumb/Breadcrumb.md
almedina-ms: The breadcrumb items will crumble when there is not enough space to render all of them, hence, 'crumbling' the highest-level (closest to the root) ones to provide more space for the other items to be rendered

active/Breadcrumb/Breadcrumb.md
almedina-ms: I believe this description can be a little longer and similar to the pips pager control

The Breadcrumb control can be navigated to via spatial navigation. The first visible breadcrumb item will automatically take focus and the items will be selected when navigated to and focused.

Left and right spatial navigation will focus and select the breadcrumb items. The Breadcrumb control will be navigated out of when the user navigates left or right when the user attempts to navigate past the first visible or last item. 

Down spatial navigation when the ellipsis item is focused will open the flyout containing the 'crumbled' items and the first item in the flyout will gain focus.

active/Breadcrumb/Breadcrumb.md
almedina-ms: This should be discussing any screen reader, the title is incorrect

active/Breadcrumb/Breadcrumb.md
almedina-ms: what do you mean by precedence in this context? are you referring to the highest to lowest level precedence or is it something different?

active/Breadcrumb/Breadcrumb.md
almedina-ms: similar to the ItemsSource in the [ItemsRepeater control](https://docs.microsoft.com/en-us/windows/uwp/design/controls-and-patterns/items-repeater#:~:text=%20ItemsRepeater%20%201%20Examples.%20If%20you%20have,to%20display%20by%20setting%20the%20ItemsSource...%20More), the value should be of "any type that implements IEnumerable"

active/Breadcrumb/Breadcrumb.md
MikeHillberg: Good point about it being a word. The difficulty with language is that it evolves :). The ToolTip API name bugs me, but when it was first created maybe it wasn't a word yet.

"Dropdown" makes sense, but we have a [DropDownButton](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.Controls.DropDownButton) today and [ComboBox.DropDownClosed](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.Controls.ComboBox.DropDownClosed), so I think we should stick with local precedence.

MikeHillberg: We just need to explain what it is then.

active/Breadcrumb/Breadcrumb.md
MikeHillberg: It's easiest to understand Gamepad behavior by describing where it's different than normal keyboard focus. It sounds like the only difference is that you open the dropdown using Gamepad with Down, and using Keyboard with space bar?

active/Breadcrumb/Breadcrumb.md
MikeHillberg: Sorry, precedence as in what is the screen reader experience in other Breadcrumb UXs today.

almedina-ms: Looks like the file explorer works different than breadcrumbs found in websites, in the case of file explorer, the whole list is another representation of the path, that's why when it gets focus the whole path is read, while in most websites, all breadcrumbs are considered elements in a numbered list (to keep hierarchy), also as there's no metadata for breadcrumbs, the only announcement made is that the navigation type is a breadcrumb, the big difference between our control and breadcrumbs in websites (such as the one in the [docs.microsoft.com](https://docs.microsoft.com/en-us/windows/winui/api/microsoft.ui.xaml.controls.itemsrepeater.itemssourceview?view=winui-2.5) website) is that our control performs crumbling while others become multi-leveled, which could explain why no context is lost with the ordered items

image

ranjeshj: @YuliKl any thoughts on this ?

active/Breadcrumb/Breadcrumb.md
```diff
@@ -0,0 +1,120 @@
...
+    static Windows.UI.Xaml.DependencyProperty ItemsSourceProperty{ get; };
+    static Windows.UI.Xaml.DependencyProperty ItemTemplateProperty{ get; };
+}
+```
+
+# Appendix
+
+## UI Automation Pattern
+
+Breadcrumb will use the
```
almedina-ms, line 100: Should we also include the ExpandCollapse pattern for the ellipsis item? 
https://docs.microsoft.com/en-us/dotnet/api/system.windows.automation.expandcollapsepattern?view=net-5.0

chingucoding: I think so, yes. After all you can expand/collapse it.

active/Breadcrumb/Breadcrumb.md
kat-y: If the dropdowns are opened with Down, does that mean IsFocusEngagmentEnabled = True ? https://docs.microsoft.com/en-us/windows/uwp/design/input/gamepad-and-remote-interactions#focus-engagement

active/Breadcrumb/Breadcrumb.md
kat-y: I'll update this section to discuss any screen reader

active/Breadcrumb/Breadcrumb.md
kat-y: I'm adding entries to the API pages section for DropDownItemTemplate and ItemTemplate

active/Breadcrumb/Breadcrumb.md
kat-y: updated this with an "is this the right control", thanks for the catch!

active/Breadcrumb/Breadcrumb.md
kat-y: updating the wording here to clarify this

active/Breadcrumb/Breadcrumb.md
kat-y: Adding this to the UI automation section, thanks for pointing it out!

active/Breadcrumb/Breadcrumb.md
kat-y: decision: read 'n of m' for the visible items. read 'more' for the ellipsis. within a flyout (evoked from the ellipsis or a chevron in future Breadcrumb) read 'n of m'

active/Breadcrumb/Breadcrumb.md
kat-y: discussion: A button should invoke the flyout, not Down

active/Breadcrumb/Breadcrumb.md
kat-y: discussion: expandcollapse pattern does not apply actually (the flyout is invoked, and focus will shift to first item in the flyout)

active/Breadcrumb/Breadcrumb.md
kat-y: have updated the code example here, will add a GIF in the future that shows the ellipsis and flyout behavior

active/Breadcrumb/Breadcrumb.md
```diff
@@ -0,0 +1,143 @@
...
+[InvokePattern](!https://docs.microsoft.com/dotnet/framework/ui-automation/implementing-the-ui-automation-invoke-control-pattern). The flyouts use the [ExpandCollapsePattern](https://docs.microsoft.com/dotnet/api/system.windows.automation.expandcollapsepattern).
+
+## Keyboard
+
+* The user can tab into the Breadcrumb and tab out.
+The first tab in gives focus to the first (highest-level) node.
+Once the Breadcrumb has focus, the user can use the left/right arrows to navigate from node to node.
+* If the Breadcrumb is crumbled, down/up will open the flyout from
```
kat-y, line 131: discussion: only invoking (space or enter) will open the flyout from the ellipsis.

active/Breadcrumb/Breadcrumb.md
kat-y: discussion: using one example of a Breadcrumb, but with a GIF showing the crumbling and flyout behavior

active/Breadcrumb/Breadcrumb.md
kat-y: See Alberto's PR to update the API here https://github.com/microsoft/microsoft-ui-xaml-specs/pull/111/files

active/Breadcrumb/Breadcrumb.md
kat-y: See Alberto's PR including updates to the API here https://github.com/microsoft/microsoft-ui-xaml-specs/pull/111/files

active/Breadcrumb/Breadcrumb.md
kat-y: updated with this behavior, also added the following details that need discussion:
  * The down/up keys will navigate between the flyout items. The ellipsis is not focusable while the flyout is open.
  * From any item in the flyout the user can use right arrow to go to the node right of the ellipsis.

active/Breadcrumb/Breadcrumb.md
```diff
@@ -0,0 +1,146 @@
...
+
+C#
+```cs
+Breadcrumb1.ItemsSource = new string[]{ "Node", "Node" , "Node"}
+```
+
+When there are too many items to display, a Breadcrumb "crumbles" and shows only the 
+last items.
+
+![Breadcrumb crumbling with 3 visible nodes](images/Breadcrumb_crumbling.PNG)
```
MikeHillberg, line 48: This says 3 but looks like 2. Or does crumble not mean what I think it means?

active/Breadcrumb/Breadcrumb.md
```diff
@@ -0,0 +1,146 @@
...
+
+
+If the last item in a Breadcrumb is too long to fit it will be truncated.
```
MikeHillberg, line 51: Is it only the last item that gets truncated? What if a non-last item is really long, will it crumble away?

active/Breadcrumb/Breadcrumb.md
```diff
@@ -0,0 +1,146 @@
...
+
+![Breadcrumb_crumbled with last node truncated](images/Breadcrumb_truncation.PNG)
+ 
+## Other Breadcrumb members
+
+| Name | Description | Default |
+| :---------- | :------- | :------- |
+| DropDownItemTemplate | Gets or sets the template to display the drop-down of nodes |
```
MikeHillberg, line 59: No mention so far of what/where a dropdown is. Is this for the crumble case?

active/Breadcrumb/Breadcrumb.md
```diff
@@ -0,0 +1,146 @@
...
+If the last item in a Breadcrumb is too long to fit it will be truncated.
+
+![Breadcrumb_crumbled with last node truncated](images/Breadcrumb_truncation.PNG)
+ 
+## Other Breadcrumb members
+
+| Name | Description | Default |
+| :---------- | :------- | :------- |
+| DropDownItemTemplate | Gets or sets the template to display the drop-down of nodes | 
+| ItemsSource | Gets or sets the content of the Breadcrumb | null |
```
MikeHillberg, line 60: Do we need an Items property (the common Items/ItemsSource pattern)? That could always be added later though.

active/Breadcrumb/Breadcrumb.md
```diff
@@ -0,0 +1,146 @@
...
+| ItemTemplate | Gets or sets the template to display an item|  | 
+| ItemClicked | Raised when a user interaction causes a jump to the CurrentItem.  | N/A |
```
MikeHillberg, line 62: What is the CurrentItem? Does this update the UI, and the expectation is that the app will respond by updating a view?

active/Breadcrumb/Breadcrumb.md
ranjeshj: Repeater's ItemsSource can be used as Items as well, but we could add that later also if we wanted to.

active/Breadcrumb/Breadcrumb.md
kat-y: I'll update the description here, this was a copy-paste mistake.

active/Breadcrumb/Breadcrumb.md
kat-y: I'll update this behavior explanation to more clearly reflect the hierarchy here. The current node is prioritized most. Crumbling as much as possible, and _then_ the last node truncates if needed. So for the scenario where any non-last node X is really long, if it was **too long for nodes X through the current leaf node to fit,** X would be crumbled.

active/Breadcrumb/Breadcrumb.md
kat-y: the GIF example will include the dropdown from the ellipsis, will that give enough context? 

In this v1 of Breadcrumb the dropdown is for the crumble case. In a v2 Breadcrumb there's a demonstrated need for dropdowns from the chevrons (to show the children nodes), which would be styled the same.

active/Breadcrumb/Breadcrumb.md
kat-y: This is meant to just say 'current item', will be fixed.

active/Breadcrumb/Breadcrumb.md
MikeHillberg: Thanks.

About the truncation, the picture here shows [TextTrimming.CharacterEllipsis](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.TextTrimming#fields), which works if it's text. If it's not text does it get clipped? Is text trimming specifically a special-case where the item is a string?

active/Breadcrumb/Breadcrumb.md
MikeHillberg: Sound perfect, thanks

active/Breadcrumb/Breadcrumb.md
MikeHillberg: Still trying to understand the typical scenario too. I think the 90% case is that you have this next to a content area, listen for this event, and then update the content area when it's raised?

kat-y: Yes, that's correct.

MikeHillberg: A little sample code would be a big help

active/Breadcrumb/Breadcrumb.md
kat-y: @YuliKl could you give any feedback for the additional keyboarding details (the two bullets above)?

active/Breadcrumb/Breadcrumb.md
kat-y: discussion: If the last item in a Breadcrumb is text and is too long to fit it will be truncated. If not text this item will be clipped.

active/Breadcrumb/Breadcrumb.md
```diff
@@ -0,0 +1,181 @@
...
+
+[MUX_PROPERTY_CHANGED_CALLBACK(TRUE)]
+[MUX_PROPERTY_CHANGED_CALLBACK_METHODNAME("OnPropertyChanged")]
+unsealed runtimeclass Breadcrumb : Windows.UI.Xaml.Controls.Control
+{
+    Breadcrumb();
+
+    Object DropDownItemTemplate{ get; set; };
+    Object ItemsSource{ get; set; };
+    Object ItemTemplate{ get; set; };
```
MikeHillberg, line 111: What is this for?

ranjeshj: ItemTemplate ? how each item in the list of crumbs look like..

MikeHillberg: Thanks. I added a diagram.

active/Breadcrumb/Breadcrumb.md
```diff
@@ -0,0 +1,195 @@
...
+
+![Diagram of the two Breadcrumb template properties](images/item-templates.jpg)
+
+The following configures a Breadcrumb so that it's dropdown displays bold text.
+
+```xaml
+<Breadcrumb>
+    <Breadcrumb.DropDownItemTemplate>
+        <DataTemplate x:DataType='BreadcrumbDropDownItem`>
+            <TextBlock Text='{x:Bind Content} FontWeight="Bold" />
```
ranjeshj, line 80: This is part of the item template, so making this bold would make all drop down items bold. Adding FontWeight here is just confusing i think.

MikeHillberg: Yes, just demonstrating why the property exists and how it's used. And yes, FontWeight maybe isn't the best demonstration :). Can you think of another example?

ranjeshj: Perhaps adding an icon ?

active/Breadcrumb/Breadcrumb.md
```diff
@@ -0,0 +1,225 @@
...
+## Breadcrumb class
+
+Represents a control that shows a list of items, for example the navigation trail to the current location.
+
+### Examples
+
+The following example code creates a simple Breadcrumb control with three nodes.
+When there is not enough width to display the whole control,
+the leading nodes are replaced with a node that the user can click on,
+opening the missing nodes in a flyout menu.
```
jonwis, line 38: Is the flyout "..." shorter by design? I can't stop looking at how the "..." and ">" are not vertically aligned with the other items in the trail.

active/Breadcrumb/Breadcrumb.md
```diff
@@ -0,0 +1,225 @@
...
+    public ObservableCollection<Items> Items { get; }
+
+    public ItemClickEventArgs Current Item { get; set;}
+
+    void CrumbClicked(Breadcrumb sender, BreadcrumbItemClickedEventArgs e)
+    {
+        for(int i = Items.Count-1; i > e.Index; i++)
+        {
+            Items.RemoveAt(i);
+        }
```
jonwis, line 85: This removal loop seems strange. Doesn't the RemoveAt(i) with i++ effectively end up skipping items in the collection?

active/Breadcrumb/Breadcrumb.md
```diff
@@ -0,0 +1,225 @@
...
+    <ContentControl Grid.Row="1" Content="{x:Bind CurrentItem, Mode=OneWay}">
+        ...
+    </ContentControl>
+</Grid>
+```
+
+```csharp
+public sealed partial class MainPage : Page
+{
+    public ObservableCollection<Items> Items { get; }
```
jonwis, line 76: Type mismatch with the `NavigateToItems` below - this is a collection of `Items`, the Navigate accepts `Item`

active/Breadcrumb/Breadcrumb.md
```diff
@@ -0,0 +1,225 @@
...
+}
+
+[webhosthidden]
+unsealed runtimeclass BreadcrumbItem : Windows.UI.Xaml.Controls.ContentControl
+{
+    BreadcrumbItem();
+}
+
+[webhosthidden]
+unsealed runtimeclass Breadcrumb : Windows.UI.Xaml.Controls.Control
```
jonwis, line 150: Should the type of control be `BreadcrumbList` ? The things visible to the user are "breadcrumbs"?

active/Breadcrumb/Breadcrumb.md
```diff
@@ -0,0 +1,225 @@
...
+{
+    Breadcrumb();
+
+    Object ItemsSource{ get; set; };
+    Object ItemTemplate{ get; set; };
+
+    event Windows.Foundation.TypedEventHandler<Breadcrumb, BreadcrumbItemClickedEventArgs> ItemClicked;
+
+    static Windows.UI.Xaml.DependencyProperty ItemsSourceProperty{ get; };
```
jonwis, line 159: Should the list of visible crumbs be obtainable, with an event to handle changes in the visible set?

active/Breadcrumb/Breadcrumb.md
MikeHillberg: 🤦‍♂️

active/Breadcrumb/Breadcrumb.md
```diff
@@ -0,0 +1,225 @@
...
+    </ContentControl>
+</Grid>
+```
+
+```csharp
+public sealed partial class MainPage : Page
+{
+    public ObservableCollection<Items> Items { get; }
+
+    public ItemClickEventArgs Current Item { get; set;}
```
oldnewthing, line 78:
```suggestion
    public Item CurrentItem { get; set;}
```

active/Breadcrumb/Breadcrumb.md
```diff
@@ -0,0 +1,225 @@
...
+The Breadcrumb ItemTemplate is used to override the default display of the
+individual items in the Breadcrumb control.
+
+The following uses the ItemTemplate so that the Breadcrumb items show the Name property of the items:
+
+```xaml
+<Breadcrumb ItemsSource='{x:Bind Items}'>
+    <Breadcrumb.ItemTemplate>
+        <DataTemplate x:DataType='BreadcrumbItem`>
+            <TextBlock Text='{x:Bind Name}/>
```
oldnewthing, line 111:
```suggestion
            <TextBlock Text='{x:Bind Name}'/>
```

active/Breadcrumb/Breadcrumb.md
```diff
@@ -0,0 +1,225 @@
...
+
+The Breadcrumb ItemTemplate is used to override the default display of the
+individual items in the Breadcrumb control.
+
+The following uses the ItemTemplate so that the Breadcrumb items show the Name property of the items:
+
+```xaml
+<Breadcrumb ItemsSource='{x:Bind Items}'>
+    <Breadcrumb.ItemTemplate>
+        <DataTemplate x:DataType='BreadcrumbItem`>
```
oldnewthing, line 110:
```suggestion
        <DataTemplate x:DataType='BreadcrumbItem'>
```

active/Breadcrumb/Breadcrumb.md
```diff
@@ -0,0 +1,225 @@
...
+## V1/V2 Breadcrumb
+
+The work for Breadcrumb has been scoped to V1 and V2 stages of the API.
+In V1 the nodes have no option to open a drop-down menu, typically used to show the node's siblings.
+V2 will add this. This spec focuses on V1 Breadcrumb,
+but keeping in mind that V1 is positioned in a way to have V2 features smoothly added later. 
+
+# API Pages
+
+## Breadcrumb class
```
oldnewthing, line 29: To me, each item is a "Breadcrumb" and the overall control is a "breadcrumb bar" / "breadcrumb path" / "breadcrumb trail" / "breadcrumb list".

DevExpress calls it a [BreadcrumbControl](https://docs.devexpress.com/WPF/401002/controls-and-libraries/navigation-controls/breadcrumb-control).

Telerik calls it a [BreadCrumb](https://www.telerik.com/products/winforms/breadcrumb.aspx) (capital C).

Actipro calls it a [Breadcrumb](https://www.actiprosoftware.com/products/controls/wpf/navigation).

Fluent UI calls it a [Breadcrumb](https://fluentsite.z22.web.core.windows.net/0.52.0/components/breadcrumb/definition).

So it seems I am the weirdo.

active/Breadcrumb/Breadcrumb.md
```diff
@@ -0,0 +1,225 @@
...
+## Gamepad
+
+* Spatial navigation to focus and select the nodes. A flyout can be opened using the A button. 
+
+## Screen reader
+
+* When the focus is on a node, the screen reader will announce the node
+as well as "n of m" to convey which node in the path it is. 
+* If the Breadcrumb has been crumbled due to resizing
+  * the ellipsis is read as 'more' and is not counted for m (the total number of nodes)
```
oldnewthing, line 223: Just to see if I understand, if the breadcrumb has crumbled and looks like this:

[...] [A] [B]

then the screen reader will initially read "More. 0 of 2. Collapsed."

If you move to "A" then it will read "A. 1 of 2."

active/Breadcrumb/Breadcrumb.md
```diff
@@ -0,0 +1,225 @@
...
+{
+    Int32 Index { get; };
+    Object Item { get; };
+}
+
+[webhosthidden]
+unsealed runtimeclass BreadcrumbItem : Windows.UI.Xaml.Controls.ContentControl
+{
+    BreadcrumbItem();
+}
```
oldnewthing, line 147: Why do we need the BreadcrumbItem class at all? The ItemsSource is a collection of Objects, which can be anything. If the app doesn't have anything convenient, it can use a String (or if it really wants an element, a TextBlock).

I'm guessing this is for parallel construction with ListBoxItem, ComboBoxItem, etc. But ListBox, ComboBox, etc. are ItemsControls and have an Items collection, so consumers which do not use data binding have something to stuff into the Items collection.

But Breadcrumb is not an ItemsControl. It doesn't have an Items collection. You *must* data-bind. So there's no need for the "lazy" item class.

(Should Breadcrumb be an ItemsControl?)

(Having a BreadcrumbItem control makes the class name BreadcrumbItemCilckedEventArgs ambiguous. Is it Breadcrumb+ItemClicked+EventArgs? Or is it BreadcrumbItem+Clicked+EventArgs?)

active/Breadcrumb/Breadcrumb.md
ranjeshj: The pictures don't show the final visual design.  @kat-y can you update the pictures ?

active/Breadcrumb/Breadcrumb.md
ranjeshj: BreadcrumbItem is what provides the default visuals for pointer pressed/over etc, so you need some container that can provide those visuals. That is similar to ListViewItem etc as you mentioned. You can manually add items to the collection that are of type BreadcrumbItem and we will use that as opposed to us automatically creating them if you provide data objects.

BreadCrumbItem+Clicked+EventArgs . I agree, that its not obvious which one we are talking about from the name.

active/Breadcrumb/Breadcrumb.md
ranjeshj: Could you provide a case where this might be useful to have ? We could add it in the future if there is a need for that.

active/Breadcrumb/Breadcrumb.md
ranjeshj: @RBrid/Alberto to confirm. UIA counts are '1' based, so I think it would read 1/3, 2/3, 3/3. If the flyout opens, that will be a separate list and have 1/x, 2/x etc..

active/Breadcrumb/Breadcrumb.md
MikeHillberg: I looked at the casing of the "C" too, and breadcrumb is indeed a word.

We also have a naming guideline against one-word names, and so we could call this BreadcrumbControl or BreadcrumbBar (aligns with ToolBar). But the proposal here is that this is a standard enough term, and won't conflict with other APIs.

active/Breadcrumb/Breadcrumb.md
MikeHillberg: Recommend: update to latest

active/Breadcrumb/Breadcrumb.md
MikeHillberg: FooItem for the container is an existing pattern, such as ListView/ListViewItem and ComboBox/ComboBoxItem.

We could still name this BreadcrumbList and then have BreadcrumbListItem though. The proposal here is that the one-word name Breadcrumb is OK.

active/Breadcrumb/Breadcrumb.md
MikeHillberg: Recommend: consider as a new feature.

active/Breadcrumb/Breadcrumb.md
almedina-ms: Yes, as @ranjeshj  mentioned "A" and "B" are 1 indexed out of the visible items, while the items in the drop-down are indexed in their context. The ellipsis should not be enumerated

MikeHillberg: Recommend: Update description

active/Breadcrumb/Breadcrumb.md
```diff
@@ -0,0 +1,225 @@
...
+Notes:
+* You can use a BreadcrumbItem as the root of the DataTemplate
+* If no ItemTemplate is supplied, the Breadcrumb control will convert to a string.
+
+## Other Breadcrumb members
+
+| Name | Description | Default |
+| :---------- | :------- | :------- |
+| ItemsSource | Gets or sets the content of the Breadcrumb |
+| ItemClicked | Raised when a user interaction causes a jump to the CurrentItem.  |
```
mrlacey, line 126: As the content of this control is set via DataBinding, it seems at odds that there is also no `ICommand` built in to handle the "Click" event.  To be more MVVM-friendly, it would be great if there wasn't a need to use an EventToCommand conversion.

Add a `Command` member that can also be invoked when an item is "clicked" so as to match the behavior of other controls and provide a consistent API experience.

mrlacey:
The current templating information is insufficient as it does not allow for the following:
- Adjust the size, shape, or position of the chevron. 
    - What if I want a different glyph?
    - What If I want the glyph on the left (before) the text and to always be displayed (even for the first item)?
- Determine the position of the item in the control.
    - E.G. If I want to style the first and last items in the control differently from ones in the middle
- Adjust the clipping/or truncation of content.
    - And how does this apply with a custom item template, where it's not just a case of setting a property on a TextBlock?
    - What if I wanted to use a glyph or different text property when space is limited? Knowing the available space or being able to set templates dependent upon the size of the available space would be very helpful.
- Style the contents of the drop down differently.
    - If this is even possible. (The current spec suggest it won't be.) 
    - Is the intention to use the same template for items in the drop down as well as directly in the control?




Some of the above may be related to adjusting the item or content template of `BreadcrumbItem` but there is no spec to cover this type (as far as I can find) beyond saying it will be a `ContenteControl`. More detail is needed for `BreadcrumbItem`.

active/Breadcrumb/Breadcrumb.md
```diff
@@ -0,0 +1,225 @@
...
+        <DataTemplate x:DataType='BreadcrumbItem`>
+            <TextBlock Text='{x:Bind Name}/>
+        </DataTemplate>
+    </Breadcrumb.ItemTemplate>
+</Breadcrumb>
+```
+
+Notes:
+* You can use a BreadcrumbItem as the root of the DataTemplate
+* If no ItemTemplate is supplied, the Breadcrumb control will convert to a string.
```
mrlacey, line 119: > If no ItemTemplate is supplied, the Breadcrumb control will convert to a string.

This is vague. I assume it means that if no custom ItemTemplate is supplied, the control will display a TextBlock with the text set as the value of calling `.ToString()` on the item, but I can't be sure from what's written.

active/Breadcrumb/Breadcrumb.md
```diff
@@ -0,0 +1,225 @@
...
+    <Breadcrumb.ItemTemplate>
+        <DataTemplate x:DataType='BreadcrumbItem`>
+            <TextBlock Text='{x:Bind Name}/>
+        </DataTemplate>
+    </Breadcrumb.ItemTemplate>
+</Breadcrumb>
+```
+
+Notes:
+* You can use a BreadcrumbItem as the root of the DataTemplate
```
mrlacey, line 118: > * You can use a BreadcrumbItem as the root of the DataTemplate

How will this be different to the example given above?
If not using a BreadcrumbItem as the root of the data template is what's provided set as the `Content` of a BreadcrumbItem automatically?
What if I tried to nest Breadcrumb controls by setting a `Breadcrumb` in the ItemTemplate of another Breadcrumb? I'm assuming it won't be a supported scenario but could it have implications that could or should be detected and reported as a warning or error?

active/Breadcrumb/Breadcrumb.md
oldnewthing: I interpreted this to be "converted to a string the same way ItemsControl converts its items to strings." The conversion appears to be language-independent. For example, C++/CX just converts everything to its class name. C# uses `Object.ToString()`. Not sure what C++/WinRT uses. Is this just `IStringable`?

chingucoding: Yes, C++/WinRT uses `IStringable` for this.

mdtauk:
> The current templating information is insufficient as it does not allow for the following:
> 
> * Adjust the size, shape, or position of the chevron.
>   
>   * What if I want a different glyph?
>   * What If I want the glyph on the left (before) the text and to always be displayed (even for the first item)?
> * Determine the position of the item in the control.
>   
You could imagine using a Slash instead of the chevron to represent a folder path.  When chevrons are added to display child nodes for each node, this kind of change would help distinguish a childless Node with a parent Node.


> * Style the contents of the drop down differently.
>   
>   * If this is even possible. (The current spec suggest it won't be.)
>   * Is the intention to use the same template for items in the drop down as well as directly in the control?
> 
I think the intent is to merge the Children Nodes menu and the overflow More menu - It would be good if these had separate lightweight styles, even if both are assigned the same for the default templates.

MikeHillberg:
Moved this to PR #122

PR #103, opened 2020-11-25
Add InfoBarPanel to InfoBar control

MikeHillberg:
Corresponds to https://github.com/microsoft/microsoft-ui-xaml/pull/3704.


active/InformationBar/InformationBar.md
```diff
@@ -398,6 +398,62 @@ Please view the guidance for
...
+The `PaddingInVerticalOrientation` property does likewise when the panel is layout out
+items vertically.
+
+[API note: these padding properties are analogous to ex the 
+[StackPanel.Padding](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.Controls.StackPanel.Padding)
+property.]
+
+## InfoBar Spacing attached properties
+
+The `SpacingInHorizontalOrientation` attached property gets/sets the distance between i tems
```
ptorr-msft, line 420: This is ambiguous; for example if I have two items like this:

```
<Button IB.SIHO="5,0"/>
<Button IB.SIHO="10,0"/>
```
Then is the distance between the two 5, 10, or 15 pixels? I assume it is 15.

active/InformationBar/InformationBar.md
```diff
@@ -398,6 +398,62 @@ Please view the guidance for
...
+when the panel is laying out items in the horizontal orientation. Similarly 
+`SpacingInVerticalOrientation` for vertical layout. This property is set on the child items,
+so the spacing can be different between different item pairs.
+
+[API note: these spacing properties are analogous to ex the
```
ptorr-msft, line 425: `Spacing` is a `double` and a simple concept, but this is a `Thickness` and as noted above is more complex.

active/InformationBar/InformationBar.md
```diff
@@ -398,6 +398,62 @@ Please view the guidance for
...
+[StackLayout.Spacing](https://docs.microsoft.com/uwp/api/Microsoft.UI.Xaml.Controls.StackLayout.Spacing)
+property, except they're attached properties so that they can be set on the panel children.]
+
+Note that if adjacent children both have non-zero spacing set, the values will be added
+together. That is if a left child has right spacing, and a right child has left spacing,
+the total spacing will be the sum of the two values.
```
ptorr-msft, line 431: OK here is the answer :). It really doesn't sound like `Spacing` then. `Spacing` is a `double` and a simple concept, but this is a `Thickness` and is more complex.

active/InformationBar/InformationBar.md
```diff
@@ -398,6 +398,62 @@ Please view the guidance for
...
+
+## InfoBarPanel example
+
+This example shows an InfoBarPanel with 10px of padding above and below
```
ptorr-msft, line 435: Two images would be nice - showing the InfoBar in both orientations and highlighting the the spacing differences.

active/InformationBar/InformationBar.md
```diff
@@ -398,6 +398,62 @@ Please view the guidance for
...
+## InfoBar Spacing attached properties
+
+The `SpacingInHorizontalOrientation` attached property gets/sets the distance between i tems
+when the panel is laying out items in the horizontal orientation. Similarly 
+`SpacingInVerticalOrientation` for vertical layout. This property is set on the child items,
+so the spacing can be different between different item pairs.
+
+[API note: these spacing properties are analogous to ex the
+[StackLayout.Spacing](https://docs.microsoft.com/uwp/api/Microsoft.UI.Xaml.Controls.StackLayout.Spacing)
+property, except they're attached properties so that they can be set on the panel children.]
```
oldnewthing, line 427: Not really analogous.

StackLayout.Spacing is a scalar and is not additive. That spacing is the distance between adjacent elements,

```xml
<!-- Child1 and Child2 have 2 units of space between them -->
<StackLayout Spacing="2"><Child1/><Child2/></StackLayout>
```

In InfoBarPanel, the spacing is a Thickness and is additive, as noted in the next paragraph. If adjacent elements both have non-zero spacing along the shared edge, this value adds them together.

```xml
<!-- Child1 and Child2 have 2+2=4 units of space between them -->
<InfoBarPanel><Child1 IBP.SIHO="2"/><Child2 IBP.SIHO="2" /></StackLayout>
```

Thickness+Additive behavior appears to be what XAML calls "Margin".

Should these be named **Margin**InHorizontalOrientation / **Margin**InVerticalOrientation?

If we rename, then we also should decide which margin wins if the children have a Margin directly applied.

active/InformationBar/InformationBar.md
```diff
@@ -398,6 +398,62 @@ Please view the guidance for
...
+[API note: these padding properties are analogous to ex the 
+[StackPanel.Padding](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.Controls.StackPanel.Padding)
+property.]
+
+## InfoBar Spacing attached properties
+
+The `SpacingInHorizontalOrientation` attached property gets/sets the distance between i tems
+when the panel is laying out items in the horizontal orientation. Similarly 
+`SpacingInVerticalOrientation` for vertical layout. This property is set on the child items,
+so the spacing can be different between different item pairs.
```
oldnewthing, line 423: Is this spacing added to any existing Margins on the children?

active/InformationBar/InformationBar.md
MikeHillberg: The distance between the two is 10, because first button has 0 right spacing, and the second button has 10 left spacing.

active/InformationBar/InformationBar.md
MikeHillberg: This also doesn't describe how the other values of the Thickness are used (for example how the height is used when it's  in a horizontal layout).

active/InformationBar/InformationBar.md
ptorr-msft: "5,0" means 5 left and right, 0 top and bottom...[?](https://docs.microsoft.com/en-us/uwp/api/Windows.UI.Xaml.Thickness?view=winrt-19041)

active/InformationBar/InformationBar.md
```diff
@@ -497,6 +555,34 @@ unsealed runtimeclass InfoBar : Windows.UI.Xaml.Controls.ContentControl
...
+## Microsoft.UI.Xaml.Controls.Primitives
+
+```cs
+[webhosthidden]
+unsealed runtimeclass InfoBarPanel : Windows.UI.Xaml.Controls.Panel
+{
+    InfoBarPanel();
+
+    // PaddingInHorizontalOrientation property
+    Windows.UI.Xaml.Thickness PaddingInHorizontalOrientation;
```
MikeHillberg, line 567: [Suggestion](https://github.com/microsoft/microsoft-ui-xaml/issues/3701#issuecomment-735314353) from @robloo: 

PaddingInHorizontalOrientation => HorizontalOrientationPadding

active/InformationBar/InformationBar.md
ranjeshj: Looking at the code, this is my understanding of how these are used. 

image

mdtauk: So the InfoBarPanel contains InfoBar items?

ranjeshj: Its an implementation detail in the template of InfoBar, it contains the icon, message text etc.. Its just a panel to ensure that all children get the correct layout positions.

ranjeshj: Couple of things to note
We cannot use SIVO[0].Top instead of PIVO.Top - because there are cases where item0 might be collapsed/hidden.
We cannot replace SIVO[x].Left with PIVO.Left - the values can be different for each item to get the correct visual alignment.

active/InformationBar/InformationBar.md
ranjeshj: Agree. Added above :)

active/InformationBar/InformationBar.md
chingucoding: What if we rename this to something more appropriate, maybe OrientationWrapPanel if that's what the purpose of this template part? Maybe someone needs something similiar to VariableSizedWrapGrid but instead for the entire collection, not individual items. If it's named InfoBarPanel I am not sure how likely it is that someone will know about the capabilities of the control.

mdtauk: > What if we rename this to something more appropriate, maybe OrientationWrapPanel if that's what the purpose of this template part? Maybe someone needs something similiar to VariableSizedWrapGrid but instead for the entire collection, not individual items. If it's named InfoBarPanel I am not sure how likely it is that someone will know about the capabilities of the control.

Could it be a control primitive that would make sense for others to use?

InfoBarPanel does make it seem like it is a Panel that contains InfoBar controls

chingucoding: From what I understand it's just a StackPanel that is oriented horizontally if there is enough space, and is oriented vertically if there is not enough horizontal space.

MikeHillberg: I think we could simplify this by having:
* Padding applies to all 4 sides
* Spacing similarly applies to all 4 sides, except it's ignored on the leading and trailing edges

The spacing property was originally named "margin", which I found confusing because the items already have a [Margin](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.FrameworkElement.Margin) property, and because this property is defined by the parent panel, so has parallels to the other panel Spacing properties like [StackPanel.Spacing](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.Controls.StackPanel.Spacing). But what this review has brought out is that that's pretty confusing, because it's a Thickness struct like a margin, and it has additive behavior like a margin.

So it looks to me like the padding can just be like every other padding, and the margins like the other Margin property, just two special things to learn:
* This margin property is additive with the existing Margin property
* The leading/trailing margins are ignored. For example, in vertical layout, the top margin of the first visible item and the bottom margin of the last visible item are ignored.

ptorr-msft: Another option -- just throwing it out there -- is to use the `Spacing` concept but allow multiple values. So for example:

```
<InfoBar HorizontalSpacing="5">...
```
In the above snippet with a single attribute value, there is 5 pixels between every item in the control (plus whatever margin the items have on themselves).

```
<InfoBar HorizontalSpacing="5,10,15">...
```
In the above snippet with multiple comma-separated values, the first gap (between item 1 and 2) is 5 pixels, the second gap (betweeen items 2 and 3) is 10 pixels, etc. again, added to whatever margin the individual items have. If there are more items than spacing values, either set to zero or just repeat.

Same for `VerticalSpacing`.

Each item can still use its own `Margin` property for precise alignment, but this makes it trivial to get consistent spacing and easy to get different spacings and not worry about special rules like "left value is ignored on the first item if it's horizontal" and such.

MikeHillberg: I think what I missed originally in this spec is that the "spacing" values are used for more than spacing between the items; in horizontal layout, the spacing.top is also used (intentionally). That's why I'm liking "margin" now, but simplify it by saying that we use all four values of the Thickness, whether horizontal or vertical.

MikeHillberg: Recommend:
* Keep the Padding property, and always use all 4 members of the struct (Top/Bottom/Right/Left)
* Rename the Spacing (attached) properties to ...Margin
* The only special case to the margin properties is to ignore the first/last. 

By "ignore first/last margin", for example in vertical layout, ignore the top margin of the first _visible_ (not collapsed) item and the bottom margin of the last visible item.

Combine that with robloo's suggestion of removing the "In" naming pattern, we have:
* HorizontalOrientationPadding / VerticalOrientationPadding (set on the parent panel)
* HorizontalOrientationMargin / VerticalOrientationMargin (set on the child elements)

active/InformationBar/InformationBar.md
MikeHillberg: Discussion: Agreed, this is now a margin (see earlier comment).

active/InformationBar/InformationBar.md
MikeHillberg: Discussion: See earlier comment, agreed on making this "margin". This margin is additive with the existing Margin property.

active/InformationBar/InformationBar.md
robloo: And I did mean that for all cases. I think there are around 4 properties total doing that.

MikeHillberg: Recommend:
     HorizontalOrientationPadding 
     VerticalOrientationPadding 
     InfoBarPanel.HorizontalOrientationMargin (attached)
     InfoBarPanel.VerticalOrientationMargin (attached)

gabbybilka (approved):

PR #104, opened 2020-11-30
PipsPager control

gabbybilka:
Add PipsPager control noted by issue [#3464](https://github.com/microsoft/microsoft-ui-xaml/issues/3464)

Open questions:

- [x] What additional theme resources need to be added?
- [ ] TODO: Update PipsPager visuals


active/PipsPager/PipsPager.md
```diff
@@ -0,0 +1,199 @@
...
+    static Windows.UI.Xaml.DependencyProperty PreviousButtonStyleProperty{ get; };
+    static Windows.UI.Xaml.DependencyProperty NextButtonStyleProperty{ get; };
+    static Windows.UI.Xaml.DependencyProperty SelectedIndicatorButtonStyleProperty{ get; };
+    static Windows.UI.Xaml.DependencyProperty DefaultIndicatorButtonStyleProperty{ get; };
+}
+```
+
+![](images/pager-control-vertical-pips-scroll-ex.jpg)
+
+## Theme Resources
```
MikeHillberg, line 153: Do we have any precedent on docs.microsoft.com where we've done a good job listing theme resources, to explain what they're about? This is a useful list but doesn't make much sense to most people.

active/PipsPager/PipsPager.md
```diff
@@ -0,0 +1,199 @@
...
+| PipsPagerSelected | Sets the selected visual state for the pip display mode | Brush
+| PipsPagerMargin | Sets the margin, in pixels, between each pip glyph | Thickness
+| PipsPagerIndicatorWidth | Sets the bounding box width for each pip glyph | Double
+| PipsPagerIndicatorHeight | Sets the bounding box height for each pip glyph | Double
+
+
+# Input & Accessibility
+
+## Keyboard
+- The PipsPager is a tab stop which will navigate to the first selectable item -- first button, previous button, or first pip. Any visible directional buttons can be navigated to via tab.
```
MikeHillberg, line 173: When the preview/next buttons are VisibleOnHover, you can still move keyboard focus to an invisible control?

active/PipsPager/PipsPager.md
```diff
@@ -0,0 +1,209 @@
...
+
+A common UI pattern not yet supported by the existing PagerControl visualizations appears in media viewer scenarios like photo carousels or app lists. In these scenarios, the pages are commonly indicated with 'dot' glyphs instead of numbers. Conceptually, this is similar to an unordered list represented by bullet points instead of numbers. We would like to introduce a PipsPager to be used when the page content is not explicitly ordered and a symbol or glyph can instead represent the page number. 
+
+# Description
+The PipsPager is a UI component that provides a standard interaction for pagination with a layout view (ListView, GridView, ItemsRepeater, DataGrid, etc.) when the page numbers do not need to be visually known. The PipsPager, like the other pager controls, is independent of the layout views and does not control any of the data being displayed in the layout view.
```
MikeHillberg, line 7: Can we get side-by-side pictures of PipsPager vs PagerControl here? It would do wonders to explain.

active/PipsPager/PipsPager.md
beervoley: @MikeHillberg yes, you can if that control is enabled (previous or last button will be disabled when first or last pip is selected, respectively).

active/PipsPager/PipsPager.md
```diff
@@ -0,0 +1,209 @@
...
+## PipsPagerButtonVisibility enum
+
+Defines constants that specify how a the buttons of a PipsPager are to be displayed. Defaults to Collapsed.
+
+| Value | Description |
+|:---:|:---|
+| Visible | The button is visible and selectable. An exception to this is that if the selected page is on the associated "edge" of the button then the button will be hidden |
+| VisibleOnHover | The behavior is the same as Visible *except* that the button is only Visible when the user is hovering over the paging UI with their cursor. |
+| COllapsed | The button is not visible to the user and does **not** take up layout space. |
+
```
beervoley, line 93: nit: upper case (C**o**llapsed)

active/PipsPager/PipsPager.md
```diff
@@ -0,0 +1,209 @@
...
+XAML
+```XAML 
+    <muxc:PipsPager NumberOfPages="5"/>
+```
+
+## PipsPager with visible buttons
+The visibility of the previous and next buttons can be set via the PreviousButtonVisibility and NextButtonVisibility properties. These buttons allow the user to select a page
+- Visible: The button is visible and selectable. An exception to this is that if the selected page is on the associated "edge" of the button then the button will be hidden.
+  - The edge for the previous button is the first page and the edge for the next button is the last page.
```
MikeHillberg, line 30: So if you click on the center of the button it goes to the next/prev page, but if you click on the _edge_ of the button it jumps to the last/first page? How close to the edge do you have to click?

active/PipsPager/PipsPager.md
```diff
@@ -0,0 +1,209 @@
...
+
+XAML
+```XAML 
+    <muxc:PipsPager NumberOfPages="5"
+      PreviousButtonVisibility=Visible
+      NextButtonVisibility=Visible />
+```
+
+## Vertical PipsPager with VisibleOnHover button visibility
+The PipsPager can also be oriented vertically which does not change its behavior or interaction. The various button visibility modes can apply to either orientation. The below example highlights the VisibleOnHover button visibility.
```
MikeHillberg, line 45: It's difficult to show this behavior in a still image. Maybe later we could get an AniGif.

active/PipsPager/PipsPager.md
MikeHillberg: Does it become visible when it gets keyboard focus?

active/PipsPager/PipsPager.md
beervoley: @MikeHillberg edge in this context doesn't literally mean "edge of the button" but rather "edge" of the pager. It means that if we're on the first page - the previous button will be hidden, alternatively if we are on the last page - the last page will be hidden. This is due to the reason, that we cannot go any further back or forward :)

gabbybilka: I was going to comment here that I think we need new terminology for this concept as its commonly confused, what makes more sense to you @MikeHillberg?

active/PipsPager/PipsPager.md
beervoley: @MikeHillberg no, it does not. We probably missed this part. Should it become visible? @gabbybilka

active/PipsPager/PipsPager.md
MikeHillberg: I see. In scrolling this is roughly what we call the "extent"?

Something like:

> Visible: The button is visible and enabled, except it is hidden at the extents. For example, when the current page is the first page, the previous button is hidden.

(Also changed 'selectable' to 'enabled', because I don't think this is a selection concept.)

active/PipsPager/PipsPager.md
```diff
@@ -0,0 +1,225 @@
...
+![A sketch of five horizontal dots with the first selected and the previous and next buttons are visible](images/PipsPager_HorizontalVisible.jpg)
+
+XAML
+```XAML 
+    <muxc:PipsPager NumberOfPages="5"
+      PreviousButtonVisibility=Visible
+      NextButtonVisibility=Visible />
+```
+
+## Vertical PipsPager with VisibleOnHover button visibility
```
MikeHillberg, line 60: I don't think we're perfectly consistent about it, but `Hover` means that the pointer is over something for a while, for example the tooltip case. (The amount of time you need to delay is specified in [UISettings.MouseHoverTime](https://docs.microsoft.com/uwp/api/Windows.UI.ViewManagement.UISettings.MouseHoverTime).)  And `Over` means the pointer is over something, no delay.

In Xaml, setting a button's mode to [ClickMode.Hover](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.Controls.ClickMode#fields) means the button will click if you put the pointer over it and wait a moment. But [ButtonBase.IsPointerOver](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.Controls.Primitives.ButtonBase.IsPointerOver) is true the whole time.

active/PipsPager/PipsPager.md
gabbybilka: See [Button](https://github.com/microsoft/microsoft-ui-xaml-specs/pull/104) API example.

active/PipsPager/PipsPager.md
gabbybilka: I believe this is resolved and yes, they will be visible. Will add in next update, thanks!

active/PipsPager/PipsPager.md
gabbybilka: Updating to use that phrasing, thanks!

active/PipsPager/PipsPager.md
```diff
@@ -0,0 +1,324 @@
...
+<StackPanel>
+  <FlipView x:Name="Gallery" MaxWidth="400" Height="270" ItemsSource="{x:Bind Pictures}">
+      <FlipView.ItemTemplate>
+          <DataTemplate x:DataType="x:String">
+              <Image Source="{x:Bind Mode=OneWay}"/>
+          </DataTemplate>
+      </FlipView.ItemTemplate>
+  </FlipView>
+
+  <muxc:PipsPager x:Name="FlipViewPipsPager"
```
MikeHillberg, line 143: Is PipsPager hard-coded to understand FlipView somehow? Or is code required here to tie these together?

active/PipsPager/PipsPager.md
```diff
@@ -0,0 +1,324 @@
...
+    [MUX_DEFAULT_VALUE("Collapsed")]
+    PagerControlButtonVisibility PreviousButtonVisibility;    
+    [MUX_DEFAULT_VALUE("Collapsed")]
+    PagerControlButtonVisibility NextButtonVisibility;
+
+    Windows.UI.Xaml.Style PreviousButtonStyle;
+    Windows.UI.Xaml.Style NextButtonStyle; 
+
+    Windows.UI.Xaml.Style SelectedPipStyle;
+    Windows.UI.Xaml.Style DefaultPipStyle;
```
MikeHillberg, line 269: Suggest (minor) calling this NormalPipStyle. I can't find API precedent, but what this represents is what Visual State Groups usually call the Normal state.

active/PipsPager/PipsPager.md
```diff
@@ -0,0 +1,324 @@
...
+By default, the previous and next buttons are collapsed, the pips are oriented horizontally, and the total number of pages is infinite.
+
+![A prototype of five horizontal dots with the first selected.](images/PipsPager_Default.png)
+
+XAML
+```XAML 
+<muxc:PipsPager x:Name="DefaultPipsPager" />
+```
+
+#### PipsPager with visible
```
MikeHillberg, line 62: Looks like an incomplete

active/PipsPager/PipsPager.md
```diff
@@ -0,0 +1,324 @@
...
+Pips represent a unit of numerical value, typically represented as dots, however can be customized to other glyphs like dashes or squares.
+In the PipsPager control by default, a solid dot will represent each page of the layout. 
+They can be selected by the user to navigate to a specific page.
+
+### Examples
+
+#### Create a PipsPager
+
+A PipsPager with five visible pips that can be selected by the user. 
+By default, the previous and next buttons are collapsed, the pips are oriented horizontally, and the total number of pages is infinite.
```
MikeHillberg, line 53: Haven't really introduced the concept of previous and next buttons. Suggest adding to the above "PipsPager Class" section a description of the parts of the control. (Or maybe generalize the "what is a pip?" section)

active/PipsPager/PipsPager.md
ranjeshj: No it is not tied together. The two way binding on SelectedIndex is what makes that work.

active/PipsPager/PipsPager.md
```diff
@@ -0,0 +1,324 @@
...
+
+runtimeclass PipsPagerSelectedIndexChangedEventArgs
+{
+    Int32 NewPageIndex{get; };
+    Int32 PreviousPageIndex{get; };
+};
+
+unsealed runtimeclass PipsPagerTemplateSettings : Windows.UI.Xaml.DependencyObject
+{
+    Windows.Foundation.Collections.IVector<Object> PipsPagerItems { get; };
```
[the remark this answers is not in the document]
ranjeshj: @StephenLPeters @beervoley Could this be int ? Is the content always ints?

active/PipsPager/PipsPager.md
MikeHillberg: Ah, sorry, I missed that. Thanks

active/PipsPager/PipsPager.md
gabbybilka: I'll leave this unresolved until extended review, thanks!

active/PipsPager/PipsPager.md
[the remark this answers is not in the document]
beervoley: @ranjeshj it's only ints. This array is just to keep the number of pips for the Repeater to keep.

active/PipsPager/PipsPager.md
gabbybilka: Hoping I helped address this in the latest push!

active/PipsPager/PipsPager.md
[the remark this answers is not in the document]
gabbybilka: I'll leave this open for now and we can discuss in review. The value of the item in PipsPagerItems isn't required or used -- only the fact that an item is there to use to calculate the size of the items collection.

mdtauk:
It's it wise to be specifically calling out the MDL2 font, and not providing a FontFamily resource here?

Fluent UI System Icons will make their way to Windows soon right?

active/PipsPager/PipsPager.md
```diff
@@ -0,0 +1,335 @@
...
+
+## PipsPager Theme Resources
+
+You can modify the look of a PipsPager by specifying Xaml resources in your app.
+For more info, see the
+[lightweight styling guide](https://docs.microsoft.com/en-us/windows/uwp/design/controls-and-patterns/xaml-styles#lightweight-styling).
+
+| Resource key | Description | Type
+|:-:|:--| :-- |
+| PipsPagerButtonWidth | Sets the bounding box width for each pip | Double |
```
gabbybilka, line 202: Wanted to highlight this issue [#4071](https://github.com/microsoft/microsoft-ui-xaml/issues/4071) that likely impacts the button width and height theme resources. Having separate theme resources for both vertical and horizontal layout orientation is a pattern we introduced previously with InfoBar. 

Would you recommend we follow that pattern here or keep the current theme resources and make changes in the code behind to support orientation switching?

active/PipsPager/PipsPager.md
```diff
@@ -0,0 +1,335 @@
...
+
+![](images/pager-control-vertical-pips-scroll-ex.jpg)
+
+
+# Input & Accessibility
+
+## Keyboard
+
+- Tab navigation into the PipsPager proceeds through the first actionable item. This would be the previous button (if not Collapsed) or the first pip. Any directional buttons can be navigated to via tab if they are not Collapsed and will be visible on focus.
+- The arrow keys can be used to navigate between the pips and directional buttons. Independent of the orientation set, the left and up arrow keys will change focus to the previous pip or to the previous button and the right and down arrow keys will change focus to the next pip or the next button.
```
oldnewthing, line 308: Presumably, navigating to a pip does not select it. (Otherwise entering the PipsPager would move you back 2 pages.) Does the space bar select the current navigated-to item?

active/PipsPager/PipsPager.md
```diff
@@ -0,0 +1,335 @@
...
+## Touch
+
+For touch-optimized experiences, we recommend that developers integrate the PipsPager with a view control (i.e. FlipView, see guidance [above](#integrate-pipsPager-with-a-collection-control)) to take advantage of on-content pagination with touch.
+
+The user can also touch to select the individual pip if able.
+
+## Screen readers
+
+- When the focus is on the control, the screen reader will announce "pager".
+- When the focus is on the previous or next button the screen reader will announce "first page", "last page", "next page", or "previous page".
```
oldnewthing, line 325: When would it announce "first page" or "last page"? Is it when focus is on the second item / second-to-last item?

active/PipsPager/PipsPager.md
```diff
@@ -0,0 +1,335 @@
...
+For touch-optimized experiences, we recommend that developers integrate the PipsPager with a view control (i.e. FlipView, see guidance [above](#integrate-pipsPager-with-a-collection-control)) to take advantage of on-content pagination with touch.
+
+The user can also touch to select the individual pip if able.
+
+## Screen readers
+
+- When the focus is on the control, the screen reader will announce "pager".
+- When the focus is on the previous or next button the screen reader will announce "first page", "last page", "next page", or "previous page". 
+  - If the buttons have text properties set by the developer, the screen reader will announce that text instead of the default announcement.
+- When the focus is on a pip, the screen reader will announce "page x of y".
```
oldnewthing, line 327: What if the number of pages is infinite?

active/PipsPager/PipsPager.md
```diff
@@ -0,0 +1,335 @@
...
+
+Commonly, a PipsPager is used in conjunction with a Collection control.
+The below example shows how to bind a PipsPager with a FlipView to provide
+an extra way to navigate through content and indicate its page.
+
+Note: To use the PipsPager as a page indicator **only**,
+set IsEnabled to false in the control to disable user interaction.
+
+![A prototype of five dots, the PipsPager, laid out horizontally underneath a FlipView showing a picture of a bicycle. The third dot in the PipsPager is selected indicated the 3rd page of content.](images/PipsPager_FlipViewExample.png)
+
```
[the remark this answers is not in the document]
ranjeshj: @gabbybilka The design has been updated. you should get screenshots from latest.
https://github.com/microsoft/microsoft-ui-xaml/pull/4060

active/PipsPager/PipsPager.md
```diff
@@ -0,0 +1,335 @@
...
+<muxc:PipsPager x:Name="CustomNavButtonPipsPager"
+  PreviousButtonStyle="{StaticResource PreviousButtonStyle}"
+  NextButtonStyle="{StaticResource NextButtonStyle}" />
+```
+## PipsPager member notes
+
+| Name | Description| Default | 
+|:---:|:---|:---|
+| NumberOfPages | Sets the max number of pages the index control will iterate through. The default will represent an infinite page range. | -1
+| Previous and Next Style | Give the developer the option to customize the style by changing the text or glyph for the edge buttons. | N/A
```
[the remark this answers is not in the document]
ranjeshj: They have a default style.

active/PipsPager/PipsPager.md
```diff
@@ -0,0 +1,335 @@
...
+      HorizontalAlignment="Center"
+      Margin="0, 10, 0, 0"
+      NumberOfPages="{x:Bind Pictures.Count}"
+      SelectedPageIndex="{x:Bind Path=Gallery.SelectedIndex, Mode=TwoWay}" />
+</StackPanel>
+```
+#### Pip and Button customization
+
+The navigational buttons and pips can be customized via the PreviousButtonStyle, NextButtonStyle, SelectedPipStyle, and DefaultPipStyle.
+As a note, setting the PreviousButtonStyle and NextButtonStyle can overwrite what is set in PreviousButtonVisibility and NextButtonVisibility.
```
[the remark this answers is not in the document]
beervoley: @oldnewthing you're correct, it's more clear this way.
It will not take precedence only in 1 case: if the **.....ButtonVisibility** is set to **Collapsed** then no matter what was in the .....ButtonStyle passed into the Pager, the Button will be collapsed at all times.

active/PipsPager/PipsPager.md
```diff
@@ -0,0 +1,335 @@
...
+| PipsPagerDefaultGlyph | Sets the default pip glyph in MDL2 icon set | String |
+| PipsPagerNavigationButtonHeight | Sets the height for each navigational button | Double |
+| PipsPagerNavigationWidth | Sets the width for each navigational button | Double |
+| PipsPagerNavigationButtonFontSize | Sets the font size for each navigational button | Double |
+| PipsPagerSelectedGlyphFontSize | Sets the size in pixels of the pip glyph when selected | Double |
+| PipsPagerDefaultGlyphFontSize | Sets the size in pixels of the pip glyph at rest| Double |
+
+## PipsPagerButtonVisibility enum
+
+Defines constants that specify how a the buttons of a PipsPager are to be displayed. Defaults to Collapsed.
```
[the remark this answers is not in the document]
beervoley: That's correct.
image

active/PipsPager/PipsPager.md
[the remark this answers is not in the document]
beervoley: @oldnewthing all the style properties have default values defined in PipsPager_themeresources.xaml. It's not necessary for the user to provide custom styles.

active/PipsPager/PipsPager.md
```diff
@@ -0,0 +1,335 @@
...
+They can be selected by the user to navigate to a specific page.
+
+### Examples
+
+#### Create a PipsPager
+
+A PipsPager with five visible pips that can be selected by the user to jump to that specific page.
+Additionally, the user can use navigational buttons; previous and next, to change the selected page incrementally.
+By default, the navigational buttons are collapsed, the pips are oriented horizontally,
+and the total number of pages is infinite.
```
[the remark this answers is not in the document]
beervoley: Yes, that's correct. We do not wrap around (in any case) beyond the first or last item.

active/PipsPager/PipsPager.md
beervoley: @oldnewthing navigating to a pip does not select it. Space bar will select the current navigated-to item.

beervoley: @gabbybilka: Tab navigation into the PipsPager proceeds through the first actionable item. This would be the previous button (if not Collapsed) or **CURRENTLY SELECTED PIP**. Any directional buttons can be navigated to via tab if they are not Collapsed and will be visible on focus.

active/PipsPager/PipsPager.md
beervoley: @oldnewthing @gabbybilka it will never announce "first" or "last" page. It will say "Page x of z".

active/PipsPager/PipsPager.md
beervoley: @oldnewthing @gabbybilka the current (incorrect) behaviour is that the Narrator will announce "
There is an [issue](https://github.com/microsoft/microsoft-ui-xaml/issues/4133) regarding this problem.

active/PipsPager/PipsPager.md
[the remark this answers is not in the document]
gabbybilka: Thanks, I addressed this in the latest update. I included the Visibility, Opacity, and IsEnabled properties as properties that will take precedence in the button Styles that are set in the visual state groups. @beervoley We can follow up on which exact properties take precedence offline.

active/PipsPager/PipsPager.md
```diff
@@ -0,0 +1,335 @@
...
+## PipsPager member notes
+
+| Name | Description| Default | 
+|:---:|:---|:---|
+| NumberOfPages | Sets the max number of pages the index control will iterate through. The default will represent an infinite page range. | -1
+| Previous and Next Style | Give the developer the option to customize the style by changing the text or glyph for the edge buttons. | N/A
+| SelectedPageIndex | The 0 based index that is currently selected. It will default to the first index. | 0
+| MaxVisiblePips | The maximum number of pips that will appear in the control at once. If the NumberOfPages is greater than this value, the UI will scroll so that the selected pip is centered. | 5
+| Orientation | The orientation of the control. Can be either Vertical or Horizontal. | Horizontal
+| PipsPagerSelectedIndexChangedEvent | Event that is fired after the user selects a pip or the directional buttons. This event will return the index number of the page that the user selected. | N/A
```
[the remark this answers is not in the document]
gabbybilka: Thanks, added these in the latest update. Note the description for the selected and default pip style

> | Name | Description| Default | 
> |:---:|:---|:---|
> | DefaultPipStyle and SelectedPipStyle | Gives the developer the option to customize the style by changing the appearance of the pips. | Based on default and selected pip theme resources and the PipsPagerButtonBaseStyle defined in theme resources.

active/PipsPager/PipsPager.md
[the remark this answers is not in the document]
gabbybilka: Thanks, added in latest update:

> | Name | Description| Default | 
> |:---:|:---|:---|
> | PreviousButtonStyle and NextButtonStyle | Gives the developer the option to customize the style by changing the text or glyph for the edge buttons. | The Previous/Next page button glyph, "Content", and PipsPagerNavigationButtonBaseStyle defined in theme resources.

active/PipsPager/PipsPager.md
gabbybilka: Added both space to select and currently selected pip note in latest update, thanks.

active/PipsPager/PipsPager.md
gabbybilka: This was an artifact of when PipsPager might have had first and last navigational buttons and were tied to those. Removed in latest update, thanks.

active/PipsPager/PipsPager.md
gabbybilka: I added this note in the latest update, thanks. I'll leave this thread open as an FYI to the open bug.

active/PipsPager/PipsPager.md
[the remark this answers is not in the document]
gabbybilka: Updated once more to clarify after discussion offline with @beervoley:

> If the PreviousButtonStyle or NextButtonStyle sets the Visibility property on the navigational button, it will take precedence over PreviousButtonVisibility or NextButtonVisibility respectively unless they are set to the PipsPagerButtonVisibility value, "Collapsed". 
> If the PreviousButtonVisibility or NextButtonVisibility is set to "Collapsed", the styles set will not overwrite the Visibility of the button itself.

active/PipsPager/PipsPager.md
```diff
@@ -0,0 +1,349 @@
...
+
+Represents a control that provides interaction and navigation of a paginated layout view 
+(ListView, GridView, ItemsRepeater, DataGrid, etc.) when the page numbers do not need to be visually known.
+
+The PipsPager, like the other pager controls, is independent of the layout views and
+does not control any of the data being displayed in the layout view.
+
+### Is this the right control?
+
+Use a PipsPager when the content in the layout is not explicitly ordered by relevancy or
```
MikeHillberg, line 37: It's still totally ordered. I think the notion of unordered here is because it's like a bulleted list, and in HTML a bulleted list is called an "unordered list". But I think this just makes the description confusing.

active/PipsPager/PipsPager.md
```diff
@@ -0,0 +1,349 @@
...
+
+#### Number of pages displayed less than total
+
+If the number of pages of content is large and do not need to be navigated to at once,
+the MaxVisiblePips property can be set to limit the number of visible, interactable, pips.
+If the NumberOfPages is greater than the MaxVisiblePips then the pips will scroll so that the selected page is
+centered in the control. If the NumberOfPages is equal to or less than the MaxVisiblePips,
+no scrolling will occur and the number of pips shown will be the same as the NumberOfPages.
+
+By default, the number of visible pips is 5.
```
MikeHillberg, line 113: Is there a special value that means no max?

active/PipsPager/PipsPager.md
```diff
@@ -0,0 +1,349 @@
...
+  NumberOfPages="5"
+  Orientation="Vertical" 
+  PreviousButtonVisibility="VisibleOnPointerOver" 
+  NextButtonVisibility="VisibleOnPointerOver" />
+```
+
+#### Number of pages displayed less than total
+
+If the number of pages of content is large and do not need to be navigated to at once,
+the MaxVisiblePips property can be set to limit the number of visible, interactable, pips.
```
MikeHillberg, line 108: Checking: the only reason I'd get fewer pips than the max is if there are fewer pages? Or is this like a scroll bar where I might get less if there's less layout space available?

active/PipsPager/PipsPager.md
```diff
@@ -0,0 +1,349 @@
...
+      NumberOfPages="{x:Bind Pictures.Count}"
+      SelectedPageIndex="{x:Bind Path=Gallery.SelectedIndex, Mode=TwoWay}" />
+</StackPanel>
+```
+
+#### Pip and Button customization
+
+The navigational buttons and pips are able to be customized via the PreviousButtonStyle, NextButtonStyle, SelectedPipStyle, and DefaultPipStyle.
+
+If the PreviousButtonStyle or NextButtonStyle sets the Visibility property on the navigational button, it will take precedence over PreviousButtonVisibility or NextButtonVisibility respectively unless they are set to the PipsPagerButtonVisibility value, "Collapsed".
```
MikeHillberg, line 159: This is a great detail to call out. This should be in the member notes (it should be on the API pages for those properties on docs.microsoft.com).

active/PipsPager/PipsPager.md
```diff
@@ -0,0 +1,349 @@
...
+      SelectedPageIndex="{x:Bind Path=Gallery.SelectedIndex, Mode=TwoWay}" />
+</StackPanel>
+```
+
+#### Pip and Button customization
+
+The navigational buttons and pips are able to be customized via the PreviousButtonStyle, NextButtonStyle, SelectedPipStyle, and DefaultPipStyle.
+
+If the PreviousButtonStyle or NextButtonStyle sets the Visibility property on the navigational button, it will take precedence over PreviousButtonVisibility or NextButtonVisibility respectively unless they are set to the PipsPagerButtonVisibility value, "Collapsed". 
+If the PreviousButtonVisibility or NextButtonVisibility is set to "Collapsed", the styles set will not overwrite the Visibility of the button itself.
```
MikeHillberg, line 160: The priority should be reversed; an local property value should override a style value.

And the priority should be consistent, rather picking different winners based on the values.

active/PipsPager/PipsPager.md
```diff
@@ -0,0 +1,349 @@
...
+
+Similarly, the SelectedPipStyle and DefaultPipStyle will take precedence over the values in the theme resources.
```
MikeHillberg, line 162: I think we can strike this line. I assume the styles reference the theme resources? Which theme resources are we talking about?

active/PipsPager/PipsPager.md
```diff
@@ -0,0 +1,349 @@
...
+  PreviousButtonVisibility="VisibleOnPointerOver" 
+  NextButtonVisibility="VisibleOnPointerOver" />
+```
+## PipsPager member notes
+
+| Name | Description| Default | 
+|:---:|:---|:---|
+| NumberOfPages | Sets the max number of pages the index control will iterate through. The default will represent an infinite page range. | -1
+| PreviousButtonStyle and NextButtonStyle | Gives the developer the option to customize the style by changing the text or glyph for the edge buttons. | The Previous/Next page button glyph, "Content", and PipsPagerNavigationButtonBaseStyle defined in theme resources.
+| DefaultPipStyle and SelectedPipStyle | Gives the developer the option to customize the style by changing the appearance of the pips. | Based on default and selected pip theme resources and the PipsPagerButtonBaseStyle defined in theme resources.
```
MikeHillberg, line 196:
```suggestion
| DefaultPipStyle and SelectedPipStyle | Gives the developer the option to customize the appearance of the pips. | Based on default and selected pip theme resources and the PipsPagerButtonBaseStyle defined in theme resources.
```

active/PipsPager/PipsPager.md
```diff
@@ -0,0 +1,349 @@
...
+  PreviousButtonStyle="{StaticResource PreviousButtonStyle}"
+  NextButtonStyle="{StaticResource NextButtonStyle}"
+  PreviousButtonVisibility="VisibleOnPointerOver" 
+  NextButtonVisibility="VisibleOnPointerOver" />
+```
+## PipsPager member notes
+
+| Name | Description| Default | 
+|:---:|:---|:---|
+| NumberOfPages | Sets the max number of pages the index control will iterate through. The default will represent an infinite page range. | -1
```
MikeHillberg, line 194: What does infinite mean? Is there a special UX?

active/PipsPager/PipsPager.md
beervoley: @MikeHillberg imagine having 100 pages, we wouldn’t want to display them all (100 pips), so the user can limit that number of displayed pips by setting the Max visible pips number and the Pager will be able to scroll to every consecutive page:)

active/PipsPager/PipsPager.md
beervoley: @MikeHillberg you’re right, we need to make sure that local value will override a style set value. @gabbybilka can you please open an issue (I’ll do it if I don’t forget) about making sure that button visibility is always matching the local PipsPager value?

active/PipsPager/PipsPager.md
beervoley: @MikeHillberg  Infinite mean that the number of pages is literally infinite and the pager will scroll through the displayed pips like a scroll viewer. We’ll attach the gifs to display behaviour

active/PipsPager/PipsPager.md
oldnewthing: Perhaps "not explicitly *numbered*, or you desire a glyph-based representation".

active/PipsPager/PipsPager.md
oldnewthing: Mike was asking about the case where there are (say) 100 pages, and MaxVisiblePips = 10, but there is only enough room to show 5 pips. Do we show 5 pips? Or do we try to show 10 pips and it's the app's job to provide a scrolling affordance?

active/PipsPager/PipsPager.md
oldnewthing, line 113: Update if true:

```suggestion
By default, the maximum number of visible pips is 5.
```

active/PipsPager/PipsPager.md
```diff
@@ -0,0 +1,349 @@
...
+| PreviousButtonStyle and NextButtonStyle | Gives the developer the option to customize the style by changing the text or glyph for the edge buttons. | The Previous/Next page button glyph, "Content", and PipsPagerNavigationButtonBaseStyle defined in theme resources.
+| DefaultPipStyle and SelectedPipStyle | Gives the developer the option to customize the style by changing the appearance of the pips. | Based on default and selected pip theme resources and the PipsPagerButtonBaseStyle defined in theme resources.
+| NextButtonVisibility | Sets the visibility of the button that navigates to the next item. | Collapsed |
+| PreviousButtonVisibility | Sets the visibility of the button that navigates to the previous item. | Collapsed |
+| SelectedPageIndex | The 0 based index that is currently selected. It will default to the first index. | 0
+| MaxVisiblePips | The maximum number of pips that will appear in the control at once. If the NumberOfPages is greater than this value, the UI will scroll so that the selected pip is centered. | 5
```
oldnewthing, line 200: Design comps disagree. Design comps do not put the selected pip in the middle if there aren't enough pips to fill to the edge. E.g., if you are on the first page, the selected pip is at the far left, rather than putting the selected pip in the center and leaving the space blank on the left.

```suggestion
| MaxVisiblePips | The maximum number of pips that will appear in the control at once. If possible, the UI will scroll the selected pip to the center. | 5
```

active/PipsPager/PipsPager.md
```diff
@@ -0,0 +1,349 @@
...
+  NextButtonVisibility="VisibleOnPointerOver" />
+```
+
+#### Number of pages displayed less than total
+
+If the number of pages of content is large and do not need to be navigated to at once,
+the MaxVisiblePips property can be set to limit the number of visible, interactable, pips.
+If the NumberOfPages is greater than the MaxVisiblePips then the pips will scroll so that the selected page is
+centered in the control. If the NumberOfPages is equal to or less than the MaxVisiblePips,
+no scrolling will occur and the number of pips shown will be the same as the NumberOfPages.
```
oldnewthing, line 111: If NumberOfPages is less than MaxVisiblePips, is space reserved for the unused pips? Or does it collapse out?

If space is reserved, does that extra space go on the right? Or do the pips center?

E.g. NumberOfPages = 3, MaxVisiblePips = 5

Option: Pips are centered in reserved space, arrows stay put.
| ◀ &nbsp; • • • &nbsp; ▶ |

Option: Pips are left-aligned in reserved space, arrows stay put.
| ◀ • • • &nbsp; &nbsp; ▶ | 

Option: Space collapses out.
| ◀ • • • ▶ |

Option: Pips are spread evenly in reserved space.
| ◀ • &nbsp; • &nbsp; • ▶ |

Option: Pips and arrows center in reserved space.
| &nbsp; ◀ • • • ▶ &nbsp; |

Option: Pips and arrows left-aligned in reserved space.
| ◀ • • • ▶ &nbsp; &nbsp; |

active/PipsPager/PipsPager.md
```diff
@@ -0,0 +1,349 @@
...
+| PipsPagerDefaultGlyphFontSize | Sets the size in pixels of the pip glyph at rest| Double |
+
+## PipsPagerButtonVisibility enum
+
+Defines constants that specify how a the buttons of a PipsPager are to be displayed.
+
+| Value | Description |
+|:---:|:---|
+| Visible | The button is visible and enabled, except it is hidden at the extents. For example, when the current page is the first page, the previous button is hidden. |
+| VisibleOnPointerOver | The behavior is the same as Visible *except* that the button is only Visible when the user is hovering over the paging UI with their cursor. |
```
oldnewthing, line 231: Is there a way to get them to be visible on touch systems? Or is the idea that on touch, you just touch the pip you want, so the arrows aren't necessary? (Except as a visual cue that there are more items.)

active/PipsPager/PipsPager.md
```diff
@@ -0,0 +1,349 @@
...
+| Collapsed | The button is not visible to the user and does **not** take up layout space. (Default for ) |
```
oldnewthing, line 232: Can I swipe the PipsPager with touch to change the selection? Or do we expect people to two-way-bind with a FlipView, and you just swipe the FlipView?

active/PipsPager/PipsPager.md
ranjeshj: If design calls for different values across horizontal and vertical, we should expose them as separate theme resources.

active/PipsPager/PipsPager.md
ranjeshj: Swiping pips pager is not supported currently. The collection itself usually provides a better swipe experience and the pips pager will respond to that based on the two way binding. Swiping the pips is something we could consider doing in the future.

active/PipsPager/PipsPager.md
gabbybilka: The idea is that on touch, you would touch the pip you want or use the touch interaction on the collection to navigate between the pages. 

I'll take a note to edit this description to reflect that the navigational button is also visible on keyboard focus.

active/PipsPager/PipsPager.md
gabbybilka: @MikeHillberg would you recommend dedicating an example in the docs to infinite pips?

active/PipsPager/PipsPager.md
gabbybilka: Yes, the styles reference the theme resources. It was to point out that they take precedence over the theme resources however if that is implied, typical, behavior I will remove, thanks!

active/PipsPager/PipsPager.md
```diff
@@ -0,0 +1,349 @@
...
+  Collapsed, 
+}
+
+runtimeclass PipsPagerSelectedIndexChangedEventArgs
+{
+    Int32 NewPageIndex{get; };
+    Int32 PreviousPageIndex{get; };
+};
+
+unsealed runtimeclass PipsPagerTemplateSettings : Windows.UI.Xaml.DependencyObject
```
gabbybilka, line 259: I noticed in the idl that the PipsPagerTemplateSettings constructor is still public. Is this intentional? I believed we had discussed previously this could be private.

active/PipsPager/PipsPager.md
gabbybilka: If there is less space available than MaxVisiblePips (e.g. setting MaxWidth to 50px with MaxVisiblePips of 5 and NumberOfPages > 5) the pips will be cut off and not scroll. I'm not sure if the app would be able to provide a scrolling affordance to navigate to the 4th pip and beyond.
image

active/PipsPager/PipsPager.md
MikeHillberg: Recommend

active/PipsPager/PipsPager.md
```diff
@@ -0,0 +1,335 @@
...
+
+The visibility of the previous and next buttons can be set via the
+`PreviousButtonVisibility` and `NextButtonVisibility` properties. 
+These buttons allow the user to incrementally select a page. The possible values are:
+- Visible: The button is visible and enabled, except it is hidden at the extents.
+  - For example, when the current page is the first page, the previous button is hidden.
+  - Note: Hidden means that the button is invisible but does take up layout space.
+- VisibleOnPointerOver: The behavior is the same as Visible *except* that the button is only Visible when
+the user is hovering over the paging UI with their cursor.
+- Collapsed: The button is not visible to the user and does **not** take up layout space.
```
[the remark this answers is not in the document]
MikeHillberg: Recommend

active/PipsPager/PipsPager.md
MikeHillberg: Discussion: If there's not enough layout space for the number of pips it will just clip, not show fewer pips.

Recommend: Call this out in the description.

active/PipsPager/PipsPager.md
MikeHillberg: Discussion: The layout space is the minimum of Number and Max.

Recommend: add a comment to the description.

active/PipsPager/PipsPager.md
MikeHillberg: Discussion: No, you must set a max (default is 5). But if you set it too big, the control will be clipped. The expectation is that it's not a common scenario to want to show more than a few (thus the 5 default).

Consider: feature to calculate the effective number of pips based on available layout size.

Recommend: Doc on Max property what the valid range is (0+). (What does zero do?)

MikeHillberg: Recommend

active/PipsPager/PipsPager.md
MikeHillberg: Recommend

active/PipsPager/PipsPager.md
MikeHillberg: Recommend: Visibility properties should always override the style.

active/PipsPager/PipsPager.md
MikeHillberg: Recommend: remove line

active/PipsPager/PipsPager.md
MikeHillberg: Recommend: just a line of text would be great.

active/PipsPager/PipsPager.md
MikeHillberg: Recommend

active/PipsPager/PipsPager.md
MikeHillberg: Recommend: update wording

active/PipsPager/PipsPager.md
MikeHillberg: Recommend: Add theme resources. FooHeightWhenVertical, FooHeightWhenHorizontal? (Follow InfoBar precedence).

active/PipsPager/PipsPager.md
MikeHillberg: Recommend: yes please on the note

active/PipsPager/PipsPager.md
MikeHillberg: Recommend: shouldn't have a public constructor.

active/PipsPager/PipsPager.md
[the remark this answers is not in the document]
MikeHillberg: Discussion: it can be anything, currently it's an int. Making this object-typed means that those ints are getting boxed though, which is unnecessary overhead.

Recommend: IVector<Int32>

active/PipsPager/PipsPager.md
```diff
@@ -0,0 +1,349 @@
...
+enum PipsPagerButtonVisibility
+{
+  Visible,
+  VisibleOnPointerOver,
+  Collapsed, 
+}
+
+runtimeclass PipsPagerSelectedIndexChangedEventArgs
+{
+    Int32 NewPageIndex{get; };
```
MikeHillberg, line 255: Discussion: precedence is Old/New, not Previous/New. Really we don't need New, because it's already on the sender.

Do we need this at all? Is there a scenario for it? App could always remember previous on its own.

Recommend: Rename "Previous" if we keep it, but consider removing both of these.

active/PipsPager/PipsPager.md
MikeHillberg: Recopmmend: NormalPipStyle

active/PipsPager/PipsPager.md
gabbybilka: Updated spec to reflect the changes proposed in issue above. @beervoley can you link your PR here when opened? Thanks!

active/PipsPager/PipsPager.md
gabbybilka: @beervoley, do we have an issue open for this?

active/PipsPager/PipsPager.md
gabbybilka: @beervoley Did you end up opening an issue for this?

MikeHillberg (approved):
Completing this old PR (API has shipped)

PR #110, opened 2021-01-12
AnimatedIcon

MarissaMatt:
Initial Commit of AnimatedIcon spec


MarissaMatt:
@mdtauk yes I plan on providing more examples that describe what you are looking for with custom states in he visual state manager. An example of how we plan to use this within WinUI is with the CheckBox control. The Checkbox control has states that include PressedOn and PressedOff and PointerOver On/Off states and we update the template for checkbox to look for the markers with these state transitions so AnimatedIcon will play the correct animation segment that translates to the visual state transition. @StephenLPeters Martin's comment about a custom state (like BluetoothOff) is an interesting one. If a developer puts a custom state string in to the template would AnimatedIcon still play the correct segment? 

To answer your theming question, you can update the Lottie file json to specify a themable color but we do not have a way to animate the custom color to a different color. In AnimatedIcon we are exposing a few color names (foreground and accent) through the interface but the developer could set any color in the lottie file via the c++/c# type rather than using the interface. Depending on the amount of customization you want to do with the colors it might make more sense to use AnimatedVisualPlayer since that has more customizations than this control. 

Do you mind expanding on your progress ring example? Progress animations are interesting because we explored creating a custom progress control that will give the developer a way to add their own animation but have all of the same properties and accessibility behaviors as the ProgressRing/ProgressBar.

mdtauk:
> @mdtauk yes I plan on providing more examples that describe what you are looking for with custom states in he visual state manager. An example of how we plan to use this within WinUI is with the CheckBox control. The Checkbox control has states that include PressedOn and PressedOff and PointerOver On/Off states and we update the template for checkbox to look for the markers with these state transitions so AnimatedIcon will play the correct animation segment that translates to the visual state transition. @StephenLPeters Martin's comment about a custom state (like BluetoothOff) is an interesting one. If a developer puts a custom state string in to the template would AnimatedIcon still play the correct segment?
> 

As an illustration here is the Bluetooth Icon, and the off state has a line through it, and an area cut off.
image
Plus there is a Searching and Connected state - so if this were an Animated Icon, it would need to transition through these.

> To answer your theming question, you can update the Lottie file json to specify a themable color but we do not have a way to animate the custom color to a different color. In AnimatedIcon we are exposing a few color names (foreground and accent) through the interface but the developer could set any color in the lottie file via the c++/c# type rather than using the interface. Depending on the amount of customization you want to do with the colors it might make more sense to use AnimatedVisualPlayer since that has more customizations than this control.
> 

When I saw the new control transitions for Sliders, and ToggleSwitches, I guessed the Checkbox would be ideal for this control.

https://user-images.githubusercontent.com/7389110/104635682-2cb5db00-569a-11eb-8dfc-ce0d136295de.mp4
_Example from Material Design_


> Do you mind expanding on your progress ring example? Progress animations are interesting because we explored creating a custom progress control that will give the developer a way to add their own animation but have all of the same properties and accessibility behaviors as the ProgressRing/ProgressBar.
>
The ProgressRing does use a Lottie player now.  But something like this
image
_example found on pintrest_

mdtauk:
Will you consider adding support for Looping animations within a visual state.  So for example, the Settings gear could continuously rotate while it is in the PointerOver state, as you hover over it.

MarissaMatt:
> Will you consider adding support for Looping animations within a visual state. So for example, the Settings gear could continuously rotate while it is in the PointerOver state, as you hover over it.

That is something I haven't thought about for the first version of this but I could definitely see the use case for it! Or maybe having a property that you can control the number of times the segment is played? So the settings gear could rotate 4 times then stop?

mdtauk:
> > Will you consider adding support for Looping animations within a visual state. So for example, the Settings gear could continuously rotate while it is in the PointerOver state, as you hover over it.
> 
> That is something I haven't thought about for the first version of this but I could definitely see the use case for it! Or maybe having a property that you can control the number of times the segment is played? So the settings gear could rotate 4 times then stop?

Yea, so maybe a
```Xaml
<AnimatedIcon>
    <AnimatedIcon.StateProperty="PointerOver" Loop="4" />
</AnimatedIcon>
```

``AnimatedIcon.StateProperty.Loop``
- Null = no loop
- Int = number of loops
- "Forever" = keep looping whilst in this state

----

Also there could be a ``Loop.Behavior``
- Smooth = waits for an animation loop to finish before changing to the next state
- Immediate = switches state immediately which may exhibit frame jumps

active/AnimatedIcon/AnimatedIconSpec.md
```diff
@@ -0,0 +1,282 @@
...
+
+1. Download the Lottie file for the icon you are looking to add to your application.
+2. Run the Lottie file through LottieGen. The LottieGen will output the file that you will add to your project that will include the markers needed for animated icon to map the correct state transitions to the correct animation segments. It will also contain the themable color properties you can set in code. 
+3. Take the output file from LottieGen and place it in your project.
+
+## Design Requirements  
+
+Designers will need to add markers named like the following to all Lottie files that will be used with AnimatedIcon.
+The marker names should reflect the visual state the animation is going from to the visual state the animation is
+going to. For example, the bare minimum of the markers that are needed for icons being used in a Button are:
```
MikeHillberg, line 44: How do you know for each control what the list of special marker names are?

active/AnimatedIcon/AnimatedIconSpec.md
```diff
@@ -0,0 +1,282 @@
...
+
+{"tm":100,"cm":"PressedToPointerOver_Start","dr":0},{"tm":101,"cm":"PressedToPointerOver_End","dr":0}] 
+
+
+
+# API Pages
+
+## AnimatedIcon class
+
+An AnimatedIcon is a UI component which plays animated images in response to user interaction and visual state changes.
```
MikeHillberg, line 79: AnimatedIcon doesn't response to interaction and visual state changes, does it? It has APIs that let you manipulate the animation, and some Xaml controls will call these APIs in response to visual states, and some visual states response to user interaction?

active/AnimatedIcon/AnimatedIconSpec.md
```diff
@@ -0,0 +1,282 @@
...
+need one, like 
+[NavigationViewItem.Icon](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.Controls.NavigationViewItem.Icon)
+or in a button.
+
+This spec introduces an AnimatedIcon IconElement which reduces, or ideally eliminates, the amount of code-behind needed to
+implement most scenarios, and it can be used in WinUI controls that support an IconElement.
+
+# How to design animations for AnimatedIcon
+
+## Creating animated content with Lottie
```
MikeHillberg, line 34: Is this different than how you create Lottie content for AVP? If not, can we just point to what's there?

active/AnimatedIcon/AnimatedIconSpec.md
MarissaMatt: That's a great question. The plan is to update the icon documentation to talk about Animated icons and to also create guidance specifically for animated icon to talk about the formatting and when/where to use animated icons.

active/AnimatedIcon/AnimatedIconSpec.md
MarissaMatt: Yes that is true and a great clarification. I will update this section to make it clearer.

active/AnimatedIcon/AnimatedIconSpec.md
MarissaMatt: No this is not a different process so I can update this to point to the information for AVP.

active/AnimatedIcon/AnimatedIconSpec.md
MikeHillberg: Shouldn't the documentation be on each control? Different controls will have different states.

active/AnimatedIcon/AnimatedIconSpec.md
```diff
@@ -0,0 +1,282 @@
...
+[NavigationViewItem.Icon](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.Controls.NavigationViewItem.Icon).
+AnimatedVisualPlayer is a more  general animation player that can be used anywhere in an application where AnimatedIcon is intended to only be used where icons are used in WinUI. 
+
+AVP allows developers to control animations using the following methods: Pause, PlayAsync, Resume, SetProgress, and Stop. When using AVP, developers may need to write a lot of code when getting their generated animation to properly respond to user interaction and state changes. This is the main distinction between the workflows for AVP and AnimatedIcon. An AnimatedIcon will be used when animations need to be driven by state changes and user interaction in a markup friendly way. It would reduce, or ideally eliminate the amount of code-behind needed to implement such scenarios.
+
+# Examples
+
+## Swap out a static icon with an animated icon in a Navigation View Item using the standard markers
+
+An icon animation will be activated by changes of state in XAML input controls (e.g. Button, and ToggleSwitch).
```
MikeHillberg, line 116: The examples here show the case of an Icon property (like NavigationViewItem), how does it work with Button, which doesn't have an Icon property?

StephenLPeters: If the default style for button gets updated (not currently planned) we could support this just working (with just the basic interaction states):
```
<Button>
    <AnimatedIcon>
        <MyAnimatedVisual/>
    </AnimatedIcon>
</Button>
```

Without an update to the default button style, you would need to either 1) retemplate button to set the animatedIcon.state attached properties on the content presenter in button. or 2) set the state property from code behind.

active/AnimatedIcon/AnimatedIconSpec.md
StephenLPeters: I think that the (small) majority of animated icon use cases will only respond to interaction events (PointerOver/pressed/normal) states so I think this basic use case can be documented with animated Icon.  Controls which have more specific needs (ie checkbox) should document the state names in their control documentation.

active/AnimatedIcon/AnimatedIconSpec.md
StephenLPeters: `will include the markers needed for animated icon to map the correct state transitions to the correct animation segments. It will also contain the themable color properties you can set in code.`

This part is different, although the steps you take as a developer (assuming the json file is formatted with animated icon in mind ie the markers were added) are the same.

active/AnimatedIcon/AnimatedIconSpec.md
StephenLPeters: You could say "An AnimatedIcon is a UI component which **enables the playing of** <s>plays</s> animated images in response to user interaction and visual state changes."

active/AnimatedIcon/AnimatedIconSpec.md
MarissaMatt: Right the designer has different steps for AnimatedIcon compared to AVP but the steps the developer needs to do to incorporate the Lottie are the same. That's a good distinction that I will make clearer.

active/AnimatedIcon/AnimatedIconSpec.md
MarissaMatt: Yeah since we are not planning on updating button for V1 I will remove that from the spec so it's not confusing.

MikeHillberg: Anything you want to keep is fine to put into the Appendix so that we can find it later.

active/AnimatedIcon/AnimatedIconSpec.md
MikeHillberg: I still can't figure out how AnimatedIcon is responding to user interaction and visual state changes. Is it getting input events somehow? I thought it had APIs on it that the control would call, but I'm not sure which ones. Does the control code just call SetState to the VSM state?

active/AnimatedIcon/AnimatedIconSpec.md
```diff
@@ -0,0 +1,282 @@
...
+## AnimatedIcon class
+
+An AnimatedIcon is a UI component which plays animated images in response to user interaction and visual state changes.
+
+On Windows, the VisualStateManager is used to manage the logic for transitioning between states for controls.
+Each control has a specified VisualStateGroup, which contains mutually exclusive VisualState objects.
+For example, a Button may have a VisualStateGroup called “CommonStates”, which specifies a VisualState object for
+“Normal” “Pressed”, “PointerOver”, and “Disabled”.
+
+A control will use the VisualStateManager to determine what state the animated icon is transitioning to and
```
MikeHillberg, line 86: What AnimatedIcon API does the control call to play the animation segment?

active/AnimatedIcon/AnimatedIconSpec.md
StephenLPeters: Agreed, and more specicially we should list which control's default template support animated icons with what markers. Its not the control which does the work to support AnimatedIcon its the template.

active/AnimatedIcon/AnimatedIconSpec.md
StephenLPeters: The only thing AnimatedIcon listens to is its state property. So in a precise sense it does not listen to user interaction.  However the scenario we were targetting when designing the control is enabling people to have a lottie icon which responds to user interaction easily.  We accomplish that through VSM setting the state property on AnimatedIcon.

active/AnimatedIcon/AnimatedIconSpec.md
StephenLPeters: You set the state property on AnimatedIcon and the property changed handler figures out what segement to play and plays it immediately (technically, it waits one layout cycle).

MikeHillberg: Can we add some sample code? (Especially useful for the AnimatedIconSource case.)

active/AnimatedIcon/AnimatedIconSpec.md
MikeHillberg: We just need more of that precision in the public docs, which is what this is going to become.

active/AnimatedIcon/AnimatedIconSpec.md
MarissaMatt: Updated.

active/AnimatedIcon/AnimatedIconSpec.md
MarissaMatt: updated.

active/AnimatedIcon/AnimatedIconSpec.md
MarissaMatt: updated.

active/AnimatedIcon/AnimatedIconSpec.md
MarissaMatt: updated and removed the button scenario.

active/AnimatedIcon/AnimatedIconSpec.md
MarissaMatt: I included an example that is for a control that already has the default template updated and an example of adding an AnimatedIcon to a control that you would need to update the template to.

active/AnimatedIcon/AnimatedIconSpec.md
```diff
@@ -280,6 +280,77 @@ XAML
 </ muxc:NavigationView.MenuItems >
 ```
 
+# Add an animated icon to a button
```
MarissaMatt, line 283: @MikeHillberg added example.

active/AnimatedIcon/AnimatedIconSpec.md
```diff
@@ -0,0 +1,417 @@
...
+
+XAML
+
+```xml
+< muxc:NavigationView.MenuItems >
+    < muxc:NavigationViewItem Content = "Menu Item1" Tag="SamplePage" x:Name = "SamplePageItem" >
+        < muxc:NavigationViewItem.Icon >
+            < controls:AnimatedIcon x:Name = "PlayAnimatedIcon" >
+                < controls:AnimatedIcon.Source >
+                    < local:PlayIcon />
```
MikeHillberg: Where did this come from? Did this come out of the LottieGen tool?

active/AnimatedIcon/AnimatedIconSpec.md
```diff
@@ -0,0 +1,417 @@
...
+Here is the XAML for the changes you would need to make to add an AnimatedIcon to a button. This includes the changes you will need to make to the state transitions for Normal, PointerOver, Pressed, and Disabled. This will update the button template to add the state transitons needed for AnimatedIcon to map the animation segment in the PlayIcon file to the correct visual state.
+
+This example also shows how to add the fallback static icon that will be used if the animated icon cannot be played. In this example we are using a SymbolIcon glyph. 
+
+XAML
+```xml
+<Button Click="Button_Click"
+        Background="LightGray"
+        Height="100" Widht="80">
+        <AnimatedIcon Source="PlayIcon" FallBackSource="{SymbolIconSource Symbol="Play"}"/>
```
MikeHillberg, line 315: Does {SymbolIconSource} work? I've been wanting to add support for that syntax, but Xaml doesn't have it today.

active/AnimatedIcon/AnimatedIconSpec.md
MikeHillberg: Does this example provide value after the previous one? It's a lot of code and I think the only new thing it demonstrates is the FallBackSource feature. I think previously I'd asked for a button example, but that was when the spec said/implied that there would be built-in support for Button content.

active/AnimatedIcon/AnimatedIconSpec.md
```diff
@@ -0,0 +1,363 @@
...
+                                    BorderBrush="{​​​​​​​​​TemplateBinding BorderBrush}​​​​​​​​​"
+                                    BorderThickness="{​​​​​​​​​TemplateBinding BorderThickness}​​​​​​​​​"
+                                    Content="{​​​​​​​​​TemplateBinding Content}​​​​​​​​​"
+                                    ContentTemplate="{​​​​​​​​​TemplateBinding ContentTemplate}​​​​​​​​​"
+                                    ContentTransitions="{​​​​​​​​​TemplateBinding ContentTransitions}​​​​​​​​​"
+                                    Padding="{​​​​​​​​​TemplateBinding Padding}​​​​​​​​​"
+                                    HorizontalContentAlignment="{​​​​​​​​​TemplateBinding HorizontalContentAlignment}​​​​​​​​​"
+                                    VerticalContentAlignment="{​​​​​​​​​TemplateBinding VerticalContentAlignment}​​​​​​​​​"
+                                    AutomationProperties.AccessibilityView="Raw" />
+                                <Grid x:Name="Icon" HorizontalAlignment="Right" Content="{​​​​​​​​​TemplateBinding Icon}​​​​​​​​​/>
```
MikeHillberg, line 292: Grid doesn't have a Content property. Is this supposed to be a ContentControl or ContentPresenter?

active/AnimatedIcon/AnimatedIconSpec.md
```diff
@@ -0,0 +1,418 @@
...
+            <StyleTargetType="Button">
+                <SetterProperty="Template">
+                    <Setter.Value>
+                        <ControlTemplateTargetType="Button">
+                            <Grid>
+                                <VisualStateManager.VisualStateGroups>
+                                    <VisualStateGroupx:Name="CommonStates">
+                                        <VisualStatex:Name="Normal">
+                                            <VisualState.Setters>
+                                                <Setter Target="Icon.(AnimatedIcon.StateProperty)"Value="Normal"/>
```
MikeHillberg, line 264: These need to be just "State", not "StateProperty". And need a space before the "Value=...".

active/AnimatedIcon/AnimatedIconSpec.md
```diff
@@ -0,0 +1,418 @@
...
+                                            </VisualSTate.Setters>
+                                        </VisualState>
+                                        <VisualState x:Name="PointerOver">
+                                            <VisualState.Setters>
+                                                <Setter Target="Icon.(AnimatedIcon.StateProperty)"Value="PointerOver"/>
+                                            </VisualSTate.Setters>
```
MikeHillberg, line 270: What are the rules on the State attached property? Where can it be set?

active/AnimatedIcon/AnimatedIconSpec.md
```diff
@@ -0,0 +1,744 @@
...
+
+XAML
+
+```xml
+<muxc:NavigationView.MenuItems >
+    <muxc:NavigationViewItem Content = "Menu Item1" Tag="SamplePage" x:Name = "SamplePageItem" >
+        <muxc:NavigationViewItem.Icon >
+            <muxc:AnimatedIcon x:Name = "PlayAnimatedIcon" >
+                <muxc:AnimatedIcon.Source >
+                    < animatedvisuals:PlayIcon />
```
MikeHillberg, line 222: What is this? Did this come from LottieGen?

active/AnimatedIcon/AnimatedIconSpec.md
```diff
@@ -0,0 +1,894 @@
...
+</NavigationView.MenuItems>
+```
+
+The NavigationViewItem will automatically set the following states on the AnimatedIcon:
+* Normal
+* PointerOver
+* Pressed
+* Selected
+* PressedSelected
+* PointerOverSelected
```
MikeHillberg, line 320: There should be a public API with static strings? LIke [MapStyleSheetEntryStates](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.Controls.Maps.MapStyleSheetEntryStates)

active/AnimatedIcon/AnimatedIconSpec.md
```diff
@@ -0,0 +1,894 @@
...
+
+```xml
+<AnimatedVisualPlayer>
+    <local:PlayAnimation />
+</AnimatedVisualPlayer>
+```
+
+The difference is that AnimatedVisualPlayer has methods such as Play and Pause.
+AnimatedIcon controls the animation with a State property, and can color the animation using the Foreground property.
+AnimatedIcon can also be used where specfically an Icon-typed value is required.
```
MikeHillberg, line 57: Another differences between the new AnimatedIcon and existing AnimatedVisualPlayer is that they consume different IAnimatedVisualSource interfaces.

active/AnimatedIcon/AnimatedIconSpec.md
```diff
@@ -0,0 +1,894 @@
...
+    AnimatedIcon.SetState(this.StackPaAnimatedIcon1nel1, "Normal");
+}
+```
+
+## AnimatedIcon.State attached property
+
+```c#
+static public void SetState(UIElement element, string value);
+static public string GetState(UIElement element);
+```
```
MikeHillberg, line 207: What's the default State?

active/AnimatedIcon/AnimatedIconSpec.md
```diff
@@ -0,0 +1,894 @@
...
+    void SetColorProperty(String propertyName, Windows.UI.Color value);
+};
+
+[webhosthidden]
+[contentproperty("Source")]
+unsealed runtimeclass AnimatedIcon : Windows.UI.Xaml.Controls.IconElement
+{
+    AnimatedIcon();
+
+    IAnimatedVisualSource2 Source{ get; set; };
```
jonwis, line 490: Super duper nit - sometimes there's a space between the property name and the brace, sometimes there isn't.

active/AnimatedIcon/AnimatedIconSpec.md
```diff
@@ -0,0 +1,894 @@
...
+This class supports markers that can be driven by an AnimatedIcon.State:
+* TBD
+
+
+
+# API Details
+
+## Microsoft.UI.Xaml.Controls
+
+```c++
```
jonwis, line 474: IIRC, `c#` as the markup might render a little better for IDL.  At some point we need to get MIDL3 to have a real grammar and stylesheet and get GitHub to support it.

active/AnimatedIcon/AnimatedIconSpec.md
```diff
@@ -0,0 +1,894 @@
...
+* NormalOffToNormalOn
+* NormalOffToPointerOverOff
+* NormalOffToPressedOff
+* PointerOverOnToPointerOverOff
+* PointerOverOnToNormalOn
+* PointerOverOnToPressedOn
+* PointerOverOffToPointerOverOn
+* PointerOverOffToNormalOff
+* PointerOverOffToPressedOff
+* PressedOnToPressedOff
```
oldnewthing, line 408: How is this transition possible? Similarly PressedOffToPressedOn. Would these be "control is programmatically checked/unchecked while the user is pressing it"?

active/AnimatedIcon/AnimatedIconSpec.md
```diff
@@ -0,0 +1,894 @@
...
+| :-| :-|
+| Markers | Provides a mapping from marker names to positions in the animation. The names are defined by the animation. |
+| SetColorProperty | Sets a color for the animated visual. The property names are defined by the animation. |
+
+## AnimatedAcceptVisualSource class
+
+This class can be used to show an animation for an Accept operation.
+This can be used as the Source of an AnimatedIcon and is used by the XAML CheckBox control.
+
+This class supports markers that can be driven by an AnimatedIcon.State:
```
oldnewthing, line 396: Shouldn't we be documentation the *states*, not the *transitions*? The transitions are autogenerated from the states, and the system looks for a marker according to the usual algorithm. As documented here, it means that if you create a marker called "NormalOn", it will never be used. But really, it will, to establish the initial state or if there is no applicable transition from the previous state.

States:
* NormalOff
* PointerOverOff
* PressedOff
* NormalOn
* PointerOverOn
* PressedOn

active/AnimatedIcon/AnimatedIconSpec.md
```diff
@@ -0,0 +1,894 @@
...
+* PressedOffToPressedOn
+* PressedOffToPointerOver
+* PressedOffToNormalOn
+
+## AnimatedGlobalNavigationButtonVisualSource class
+
+This class can be used to show an animation for a Navigation operation.
+This can be used as the Source of an AnimatedIcon and is used by the XAML NavigationView control.
+
+This class supports markers that can be driven by an AnimatedIcon.State:
```
oldnewthing, line 420: Again, we should be documenting states, not markers. The algorithm for generating markers from states is already described elsewhere, and it avoids having to provide a list of the combinatorial explosion of markers. Saves us a bunch of typing, and saves the developer a bunch of reading (and then trying to reverse-engineer the states from the list of markers, like I had to do with the checkbox).

active/AnimatedIcon/AnimatedIconSpec.md
```diff
@@ -0,0 +1,894 @@
...
+
+# AnimatedChevronUpDownSmallVisualSource
+
+This class can be used to show an animation for button that opens a flyout on another button.
+This can be used as the Source of an AnimatedIcon and is used by the XAML NavigationView control.
+
+This class supports markers that can be driven by an AnimatedIcon.State:
+* TBD
+
+# AnimatedChevronRightDownSmallVisualSource class
```
oldnewthing, line 460:
```suggestion
## AnimatedChevronRightDownSmallVisualSource class
```

active/AnimatedIcon/AnimatedIconSpec.md
```diff
@@ -0,0 +1,894 @@
...
+
+## AnimatedChevronDownSmallVisualSource class
+
+This class can be used to show an animation for button that opens a flyout on another button.
+This can be used as the Source of an AnimatedIcon and is used by the XAML DropDownButton and Expander controls.
+
+This class supports markers that can be driven by an AnimatedIcon.State:
+* TBD
+
+# AnimatedChevronUpDownSmallVisualSource
```
oldnewthing, line 452:
```suggestion
## AnimatedChevronUpDownSmallVisualSource
```

active/AnimatedIcon/AnimatedIconSpec.md
```diff
@@ -0,0 +1,894 @@
...
+[webhosthidden]
+[contentproperty("Source")]
+unsealed runtimeclass AnimatedIcon : Windows.UI.Xaml.Controls.IconElement
+{
+    AnimatedIcon();
+
+    IAnimatedVisualSource2 Source{ get; set; };
+    IconSource FallbackSource {get; set; };
+
+    static Windows.UI.Xaml.DependencyProperty State{ get; };
```
oldnewthing, line 493:
```suggestion
    static Windows.UI.Xaml.DependencyProperty StateProperty{ get; };
```

and move to the end next to SourceProperty.

active/AnimatedIcon/AnimatedIconSpec.md
```diff
@@ -0,0 +1,894 @@
...
+    static void SetState(Windows.UI.Xaml.DependencyObject object, String value); 
+    static String GetState(Windows.UI.Xaml.DependencyObject object); 
+
+    static Windows.UI.Xaml.DependencyProperty SourceProperty{ get; };
```
oldnewthing, line 497: What about FallbackSourceProperty? Wouldn't the XAML compiler need that in order to set FallbackSource from markup?

active/AnimatedIcon/AnimatedIconSpec.md
```diff
@@ -0,0 +1,894 @@
...
+
+```
+
+# Appendix
+
+## Accessibility
+
+This control is going to be part of the Icon family of controls so the accessibility requirements will reflect what the other icon controls have. AnimatedIcon will need to ensure that the high contrast brushes propagate correctly if the user changes their settings to a high contrast theme.  
+
+If the user turns off animations in theri system settings, the AnimatedIcons will not animate.
```
oldnewthing, line 596:
```suggestion
If the user turns off animations in their system settings, the AnimatedIcons will not animate. 
```

oldnewthing, line 596: Presumably "will not animate" means "will hard-cut to the final position."

active/AnimatedIcon/AnimatedIconSpec.md
```diff
@@ -0,0 +1,894 @@
...
+
+### Updates to IconSource to support AnimatedIconSource
+
+Since some controls take an IconElement and others take an IconSource, we are going to add two methods (public CreateIconElement and 
+protected CreateIconElementCore) to IconSource. This will help cover the two sccenarios below: 
+
+1.	I’m a developer creating a new control and I want to expose an IconSource property where consumers of my control can set ‘any’ iconsource. This control developer should not have to worry about each type of IconSource like FontIconSource, BitmapIconSource, AnimatedIconSource etc. In this case the developer can just use IconSource and call IconSource.CreateIconElement to get the corresponding IconElement without knowing what specific IconSource/Element it is. 
+2.	I’m a developer creating a new type of IconSource, say Animated2IconSource. I would have to override Animated2IconSource.CreateElementCore and return an Animated2IconElement so that my new IconSource will work anywhere an IconSource can be used.
```
oldnewthing, line 608: What about

3. I'm a developer who already created a new type of IconSource, say, AppxPackageIconSource and shipped it for use with mux:v2. When the app updates to the new version of mux, my AppxPackageIconSource.CreateIconElement method will not work, because I did not implement CreateIconElementCore.

active/AnimatedIcon/AnimatedIconSpec.md
```diff
@@ -0,0 +1,894 @@
...
+Composition [Visual](https://docs.microsoft.com/uwp/api/Windows.UI.Composition.Visual),
+which can be added to a XAML element tree using the methods of
+[ElementCompositionPreview](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.Hosting.ElementCompositionPreview).
+
+Use the `SetColorProperty` method to set a color of the animated visual,
+and the `Markers` property to find named positions in the animation timeline.
+
+_Spec note: `IAnimatedVisualSource2` is like `IAnimatedVisualSource` in purpose and use,
+but there is no derivation because the '2' interface does not derive from
+[IDynamicAnimatedVisualSource](https://docs.microsoft.com/uwp/api/Microsoft.UI.Xaml.Controls.IDynamicAnimatedVisualSource)._
```
oldnewthing, line 352: How did I**Dynamic**AnimatedVisualSource enter the picture? The 2 interface would presumably derive from IAnimatedVisualSource.

Or are you saying that IDynamicAnimatedVisualSource does not derive from IAnimatedVisualSource2? Would we expect it to?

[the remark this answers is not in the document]
oldnewthing: Why not derive from 1? Implementations can just ignore the `diagnostics` parameter.

As for names: How about IAnimatedVisualSourceWithMarkers?

active/AnimatedIcon/AnimatedIconSpec.md
```diff
@@ -0,0 +1,894 @@
...
+
+```xml
+<Button PointerEntered="HandlePointerEntered" PointerExited="HandlePointerExited">
+    <StackPanel Orientation='Horizontal' AnimatedIcon.State='Normal'>
+
+        <AnimatedIcon x:Name='AnimatedIcon1'>
+            <local:MyAcceptAnimation/>
+            <AnimatedIcon.FallbackIconSource>
+                <SymbolIconSource Symbol='Accept'/>
+            </AnimatedIcon.FallbackIconSource>
```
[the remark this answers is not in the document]
MarissaMatt: Are you referring to if the user turns all animations off in system settings? In that case we would just display the first frame of the animation.

active/AnimatedIcon/AnimatedIconSpec.md
```diff
@@ -0,0 +1,894 @@
...
+* Selected
+* PressedSelected
+* PointerOverSelected
+
+If the PlayIcon has a marker for "NormalToPointerOver", it will be animated
+to when the mouse moves to over the NavigationViewItem.
+See [link] for more information on combining marker names.
+
+To create this custom PlayIcon class, you run a Lottie JSON file (with markers) through the Windows
+[CodeGen](https://docs.microsoft.com/en-us/windows/communitytoolkit/animations/lottie-scenarios/getting_started_codegen)
```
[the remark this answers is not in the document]
MarissaMatt: Yes it is the same tool and that should actually say LottieGen instead of CodeGen. LottieGen is the tool the developer uses and CodeGen is the output generated from the tool.

active/AnimatedIcon/AnimatedIconSpec.md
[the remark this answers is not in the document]
oldnewthing: Shouldn't it be the final frame? Because the first frame is probably going to look like a fade-in or some other entrance animation, not steady state.

active/AnimatedIcon/AnimatedIconSpec.md
```diff
@@ -0,0 +1,894 @@
...
+[NavigationViewItem.Icon](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.Controls.NavigationViewItem.Icon).
+`AnimatedVisualPlayer` is a more  general animation player that can be used anywhere in an application where AnimatedIcon is intended to
+only be used where icons are used in WinUI.
+
+### Example
+
+The following example is a button that the user clicks to Accept a prompt.
+The MyAcceptAnimation is a class created using the LottieGen tool.
+The FallbackIconSource will be used rather than the animation when animations can't be played,
+such as on older versions of Windows that don't support Lottie animations.
```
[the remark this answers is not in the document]
MarissaMatt: Another reason the FallbackIconSource would be used is if the animation file has something in it that is only supported on a newer version of windows and the OS cannot play it. An example would be a certain component that was added to the composition layer that is only supported by newer version of the OS.

active/AnimatedIcon/AnimatedIconSpec.md
MarissaMatt: I believe it's "Normal" but I will let @StephenLPeters confirm.

active/AnimatedIcon/AnimatedIconSpec.md
chingucoding: Agree with @oldnewthing, we shouldn't list the markers but rather the states to save us and developers time.

active/AnimatedIcon/AnimatedIconSpec.md
```diff
@@ -0,0 +1,894 @@
...
+Call this method to get an IconElement from an IconSource.
+
+An IconElement can be used by controls that have an IconElement property,
+or since it's an element it can be used anywhere in the XAML element tree.
+
+This example creates a BitmapIcon from a BitmapIconSource:
+
+```cs
+IconSource iconSource = new BitmapIconSource();
+BitmapIcon = iconSource.CreateIconElement as BitmapIcon;
```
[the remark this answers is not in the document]
MikeHillberg: Recommend

active/AnimatedIcon/AnimatedIconSpec.md
```diff
@@ -0,0 +1,894 @@
...
+
+            <AnimatedIcon >
+                <local:PlayIcon />
+            </AnimatedIcon>
+
+    </NavigationViewItem.Icon>
+</NavigationView.MenuItems>
+```
+
+The NavigationViewItem will automatically set the following states on the AnimatedIcon:
```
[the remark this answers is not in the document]
MikeHillberg: Don't use the markers it's looking for?

active/AnimatedIcon/AnimatedIconSpec.md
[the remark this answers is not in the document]
MikeHillberg: Recommend: pick one name and stick with it

active/AnimatedIcon/AnimatedIconSpec.md
```diff
@@ -0,0 +1,894 @@
...
+
+If the PlayIcon has a marker for "NormalToPointerOver", it will be animated
+to when the mouse moves to over the NavigationViewItem.
+See [link] for more information on combining marker names.
+
+To create this custom PlayIcon class, you run a Lottie JSON file (with markers) through the Windows
+[CodeGen](https://docs.microsoft.com/en-us/windows/communitytoolkit/animations/lottie-scenarios/getting_started_codegen)
+tool to generate the animation code as a C# class.
+
+Once you run your Lottifile through codegen you can add the output class to your project.
```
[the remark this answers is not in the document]
MikeHillberg: Just checking: is it OK that we use "Lottie" everywhere? That's not a nickname?

active/AnimatedIcon/AnimatedIconSpec.md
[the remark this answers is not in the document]
MikeHillberg: IAnimatedVisualSourceWithMarkersAndColors

MikeHillberg: Discussion: I got confused and reversed the inheritance order.

Recommend: delete that

active/AnimatedIcon/AnimatedIconSpec.md
```diff
@@ -0,0 +1,894 @@
...
+
+```cpp
+void AddVisualAndShowStartAnimation(
+    const winrt::Border& element,
+    const winrt::IAnimatedVisualSource2 source,
+    const winrt::hstring& initialState,
+    const winrt::hstring& steadyState,
+    const winrt::Color& themeColor )
+{
+    winrt::IAnimatedVisual animatedVisual = source.TryCreateAnimatedIconVisual();
```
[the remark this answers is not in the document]
MikeHillberg: Recommend: Get compiling code

active/AnimatedIcon/AnimatedIconSpec.md
```diff
@@ -0,0 +1,894 @@
...
+
+    // Find the positions in the animation of the two states
+    auto const markers = source.Markers())
```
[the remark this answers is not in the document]
MikeHillberg: Recommend: Get compiling code

active/AnimatedIcon/AnimatedIconSpec.md
```diff
@@ -0,0 +1,894 @@
...
+    auto const fromProgress = static_cast<float>(markers.Lookup(initialState));
+    auto const toProgress = static_cast<float>(markers.Lookup(steadyState));
+
+    // Set the theme color as the animated visual's foreground.
+    source.SetColorProperty("Foreground", themeColor);
+
+    // Helper that uses TryCreateAnimatedVisual to add the source to the XAML element tree.
+    // See IAnimatedVisualSource.TryCreateAnimatedVisual for more information.
```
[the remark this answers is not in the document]
MikeHillberg: Comment is wrong; TryCreateAnimatedVisual was already called earlier.

Is it OK to be modifying the source after the visual has been created?

active/AnimatedIcon/AnimatedIconSpec.md
MikeHillberg: This is an L2, the L1 is "API Pages"

active/AnimatedIcon/AnimatedIconSpec.md
MikeHillberg: This is an L2, the L1 is "API Pages"

active/AnimatedIcon/AnimatedIconSpec.md
MikeHillberg: Recommend: "csharp" (I think the DMC lint plugin prefers this)

active/AnimatedIcon/AnimatedIconSpec.md
MikeHillberg: Agreed there should be a DP. Note that markup doesn't depend on it though; it can set any WinRT or C# property.

active/AnimatedIcon/AnimatedIconSpec.md
MikeHillberg: I swear I spell-checked.

Recommend

active/AnimatedIcon/AnimatedIconSpec.md
MikeHillberg: I think this is equivalent to: I wrote an IconSource subclass but didn't do the override (we don't have the ability to make this an abstract method). I think we should just have a documented error.

active/AnimatedIcon/AnimatedIconSpec.md
```diff
@@ -0,0 +1,894 @@
...
+3) If "[PreviousState]To[NewState]_End" is not found but "[PreviousState]To[NewState]_Start" is found,
+then we will hard cut to the "[PreviousState]To[NewState]_Start" position.
+4) Check if the provided IAnimatedVisualSource2's markers for the marker "[PreviousState]To[NewState]".
+If it is found then we will hard cut to the "[PreviousState]To[NewState]" position.
+5) Check if the provided IAnimatedVisualSource2's markers for the marker "[NewState]".
+If it is found, then we will hard cut to the "[NewState]" position.
+6) Check if the provided IAnimatedVisualSource2's markers has any marker which ends with "To[NewState]_End".
+If any marker is found which has that ending, we will hard cut to the first marker found with the appropriate ending's position.
+7) Check if "[NewState]" parses to a float. If it does, animated from the current position to the parsed float. **
+Hard cut to position 0.0.
```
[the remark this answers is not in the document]
MikeHillberg: Recommend

active/AnimatedIcon/AnimatedIconSpec.md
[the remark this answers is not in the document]
MikeHillberg: Sounds good, thanks

active/AnimatedIcon/AnimatedIconSpec.md
```diff
@@ -0,0 +1,894 @@
...
+<Button>
+    <StackPanel Orientation='Horizontal' x:Name='StackPanel1'>
+        <AnimatedIcon ... />
+        <TextBlock Text='Click' />
+    </StackPanel>
+</Button>
+```
+
+Some controls will automatically set the State on an AnimatedIcon, for example
+see NavigationViewItem.Icon.
```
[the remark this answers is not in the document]
MikeHillberg: Sorry, yes, that text isn't accurate until we put this onto DMC.

Recommend: update

active/AnimatedIcon/AnimatedIconSpec.md
```diff
@@ -0,0 +1,894 @@
...
+_Spec note: `IAnimatedVisualSource2` is like `IAnimatedVisualSource` in purpose and use,
+but there is no derivation because the '2' interface does not derive from
+[IDynamicAnimatedVisualSource](https://docs.microsoft.com/uwp/api/Microsoft.UI.Xaml.Controls.IDynamicAnimatedVisualSource)._
+
+This example displays and animates an object implementing IAnimatedVisualSource2:
+
+```cpp
+void AddVisualAndShowStartAnimation(
+    const winrt::Border& element,
+    const winrt::IAnimatedVisualSource2 source,
```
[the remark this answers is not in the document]
MikeHillberg: Recommend

active/AnimatedIcon/AnimatedIconSpec.md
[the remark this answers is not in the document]
MarissaMatt: Added it to the example here but I think accessibility should be it's own section in the final documentation.

MikeHillberg:
Completing this PR that got dropped. Moved most of the open issues into "issue:" text in the doc.

MikeHillberg (approved):

PR #112, opened 2021-02-11
ThemeVersion API

MikeHillberg:


active/Resources/StyleVersion2.md
```diff
@@ -0,0 +1,89 @@
...
+</Application.Resources>
+```
+
+# API Details
+
+**Namespace Microsoft.UI.Xaml.Controls**
+
+```cs
+[webhosthidden]
+enum ThemeVersion
```
MikeHillberg, line 54: Too generic? Name it `XamlThemeVersion`?

active/Resources/StyleVersion2.md
```diff
@@ -0,0 +1,89 @@
...
+
+The default value of this property will be the latest version, but an app can set another value to force to a
+previous version.
+
+# API Pages
+
+## XamlControlsResources.ThemeVersion
+
+Set this property to specify which version of resources to use for the look and styling
+of XAML controls. Default value is Version2.
```
MikeHillberg, line 26: How does this affect breaking change rules?

active/Resources/StyleVersion2.md
```diff
@@ -0,0 +1,89 @@
...
+
+XAML (WinUI) has a resources feature where you can put objects in a "resource dictionary", then
+use the resources by referencing the key name. All of the styling for XAML controls
+is defined in a resource dictionary and referenced by the controls.
+But an application can override these by defining a resource dictionary with one or more matching keys.
+
+XAML ships such an overrides dictionary named
+[XamlControlsResources](http://msdn.microsoft.com/library/Microsoft.UI.Xaml.Controls.XamlControlsResources).
+This dictionary defines all the resources necessary for all of the built-in controls.
+Since the resources in this dictionary can change over time, a new `ThemeVersion` property is being added.
```
oldnewthing, line 14: The word "theme" seems to be used for a bunch of related (but different) concepts.

1. [XamlUIPresenter.ThemeKey](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.Hosting.XamlUIPresenter.ThemeKey?view=winrt-19041) / [XamlUIPresenter.ThemeResourcesXaml](https://docs.microsoft.com/uwp/api/windows.ui.xaml.hosting.xamluipresenter.themeresourcesxaml?view=winrt-19041) / [ResourceDictionary.ThemeDictionaries](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.ResourceDictionary.ThemeDictionaries?view=winrt-19041)

This sense of "theme" appears to be "choosing a portion of the combined resource dictionary to apply to the app."

2. [Application.RequestedTheme](https://docs.microsoft.com/en-us/uwp/api/Windows.UI.Xaml.Application.RequestedTheme?view=winrt-19041)

This sense of "theme" appears to be "choosing between light or dark default styling."

3. XamlControlsResources.ThemeVersion (introduced here)

This sense of theme appears to be "choosing a historical styling set."

The new sense is similar to number 2, in that they both mean "Completely swapping out a set of resources."

So you could pick a ThemeVersion = Version2, and a RequestedTheme = Light, and you get "Version 2 with light parts."

Maybe this is really a "StyleSet"?

active/Resources/StyleVersion2.md
```diff
@@ -0,0 +1,89 @@
...
+
+# Background
+
+XAML (WinUI) has a resources feature where you can put objects in a "resource dictionary", then
+use the resources by referencing the key name. All of the styling for XAML controls
+is defined in a resource dictionary and referenced by the controls.
+But an application can override these by defining a resource dictionary with one or more matching keys.
+
+XAML ships such an overrides dictionary named
+[XamlControlsResources](http://msdn.microsoft.com/library/Microsoft.UI.Xaml.Controls.XamlControlsResources).
```
oldnewthing, line 12: Broken link. This one seems to work: [XamlControlsResources](https://docs.microsoft.com/en-us/windows/winui/api/microsoft.ui.xaml.controls.xamlcontrolsresources?view=winui-2.5).

(Same broken link appears later in the document as well.)

active/Resources/StyleVersion2.md
```diff
@@ -0,0 +1,89 @@
...
+
+Version 1 aligns to WinUI 2.5, Version 2 aligns with WinUI 2.6.
+
+> This property must be set during app startup and cannot be changed after that point.  
+(When, exactly?)
+
+> Spec note: the alternative is to have an 'Auto' value in the enum,
+and then the default value of the property is always that.
+That would mean, though, that we also need an ActualThemeVersion property for
+controls to query
```
oldnewthing, line 36: Or we do it this way: Each time the theme is refreshed, we introduce a new enum for it, which becomes the new default.

So you always get what the property says, and you can peek at the default value to see what the system recommends.

active/Resources/StyleVersion2.md
ShawnSteele: How does this differ from MRT?  Getting different resources for different scales or languages?

active/Resources/StyleVersion2.md
```diff
@@ -0,0 +1,89 @@
...
+Some explanation of this feature and how it relates to "System XAML" ...
+
+System XAML ships in Windows and defines a resource dictionary with all of the resources
+for the XAML controls. WinUI2 is a control library that ships in NuGet and defines the
+[XamlControlsResources](http://msdn.microsoft.com/library/Microsoft.UI.Xaml.Controls.XamlControlsResources)
+type, which has overrides for all of the built-in resources and more.
+
+The resources in System XAML can change from release to release, and there is no override
+API to request an older version. The version of the resources that your app gets is determined
+by the version of Windows your app requests in the app manifest (MaxTestedVersion). The new API
```
jonwis, line 88: Ok so there's now four(?) sources of this version:

1. What your packaged app version says about Windows tested version
2. What your XAML `Application.Resources` says
3. What your app specified by setting `XamlControlsResources.ThemeVersion`
4. What your XAML version (System XAML, WinUI3) supports

phew.

active/Resources/StyleVersion2.md
MikeHillberg: The plan is that the default value will change from version to version.

active/Resources/StyleVersion2.md
MikeHillberg: Unfortunately Xaml and the [ResourceManager](https://docs.microsoft.com/uwp/api/Windows.ApplicationModel.Resources.Core.ResourceManager) have independent features both named "resources".  Adapting to resolution scales and languages are generally done in RESW, Xaml resources are used for sharing objects such as sharing a Brush between controls.

Regarding "StyleSet", the original name we had for this was "StylesVersion", but my confusion with that name is that the resource dictionary has more than Style objects in it, it has a lot of brushes and colors and numerics as well.

But in addition to that list of "theme" terminology is "{ThemeResource}".

Brainstorming:
ThemeVersion
StylesVersion
StyleSet
ResourcesVersion
ControlsResourcesVersion
?

active/Resources/StyleVersion2.md
MikeHillberg: That's the plan in this spec. The spec note here is calling out the alternative that was rejected.

active/Resources/StyleVersion2.md
oldnewthing: DefaultResourcesVersion? ("What version of the default resources do you want?" "Oh, I want the version that came with WinUI 2.1.")

active/Resources/StyleVersion2.md
jonwis: Maybe that's a prose problem in the spec, then. Something like:

> This value defaults to the Framework's preferred version, which at this time is `Version2`. Setting this property overrides the framework's preference, allowing your app to specify which theme version should be applied.

active/Resources/StyleVersion2.md
jonwis: I know _why_ we're having this conversation, but the notion of a "version" is that it's nominally linear and you can compare them to know when X is better (greater) than Y.  Maybe a compromise is `ThemeStyle` or `StyleTheme` ? Something that we won't regret when we want to say that "apples are not better than oranges, they're simply different."

MikeHillberg: In this case they are linear, but I don't think there are any comparison scenarios. You typically don't set it, but maybe you have a custom set of resources that's designed to align with Version3, so you want to set this property to Version3.

By ThemeStyle, you mean e.g. ThemeStyle.Version2?

MikeHillberg: Recommend:
Since this is the version of the XamlControlsResources, name this enum ControlsResourcesVersion.

active/Resources/StyleVersion2.md
MikeHillberg: Recommend: update to that, that makes sense.

active/Resources/StyleVersion2.md
MikeHillberg: Discussion: see earlier comment (rename)

active/Resources/StyleVersion2.md
```diff
@@ -0,0 +1,89 @@
...
+# API Pages
+
+## XamlControlsResources.ThemeVersion
+
+Set this property to specify which version of resources to use for the look and styling
+of XAML controls. Default value is Version2.
+
+Version 1 aligns to WinUI 2.5, Version 2 aligns with WinUI 2.6.
+
+> This property must be set during app startup and cannot be changed after that point.
```
chingucoding, line 30: Users still can do ´<XamlControlsResources />` though since there will be a default value right?

MikeHillberg: Still do what?

